PR #7, opened 2020-12-09
fix(dropdown): add interfaces with default props, testids

joasmi723:
Adding interfaces for all dropdown components that extend default html elements, so styles (and other props) can be added.  I also added testIds and refs, and made sure all the interfaces were exported from react-magma-dom.


packages/react-magma-dom/src/components/Dropdown/DropdownDivider.tsx
```diff
@@ -2,14 +2,22 @@ import * as React from 'react';
...
-export const DropdownDivider: React.FunctionComponent = () => {
-  const theme = React.useContext(ThemeContext);
+export const DropdownDivider: React.FunctionComponent<DropdownDividerProps> = (
+  props: DropdownDividerProps
+) => {
+  const { testId, ...other } = props;
 
-  return <StyledHr theme={theme} />;
+  const theme = React.useContext(ThemeContext);
+  return <StyledHr {...other} data-testid={props.testId} theme={theme} />;
```
rcgillet, line 22: `testId` was pulled off the `props` but you still used `props.testId` here.

rcgillet, line 22:
```suggestion
  return <StyledHr {...other} data-testid={testId} theme={theme} />;
```

rcgillet (changes requested):

rcgillet (approved):

PR #19, opened 2020-12-10
Feat/custom select item

rcgillet:


packages/react-magma-dom/src/components/Combobox/Combobox.tsx
```diff
@@ -44,6 +44,7 @@ export function InternalCombobox<T>(props: ComboboxInterface<T>) {
     onIsOpenChange,
     onItemCreated,
     placeholder,
+    renderItem,
```
avanslaars, line 47: We have a `components` object for swapping out default implementations of components. Is there a reason to have this in a separate prop? Seems like `components` would cover this and all future custom component needs.

avanslaars (changes requested):
I'll look again after we sort out the current design vs making this part of the `components` prop

packages/react-magma-docs/src/pages/api/select.mdx
```diff
@@ -795,6 +796,70 @@ interface CustomSelectItem {
 }
 ```
 
+## Custom Item Render
+
```
joasmi723, line 800: I feel like this could use a paragraph describing what this does (and how it's different from the other custom examples)

packages/react-magma-docs/src/pages/api/combobox.mdx
```diff
@@ -913,6 +914,71 @@ interface CustomComboboxItem {
 }
 ```
 
+## Custom Item Render
+
```
joasmi723, line 918: I feel like this could use a paragraph describing what this does (and how it's different from the other custom examples)

joasmi723: Looks good!

packages/react-magma-docs/src/data/properties.json
```diff
@@ -6213,6 +6240,14 @@
         },
         "deprecated": false
       },
+      "renderItem": {
```
avanslaars, line 6243: Looks like the old prop name is being generated as an option here... is this correct, or should there be an update to the `components` prop documentation?

packages/react-magma-docs/src/pages/api/select.mdx
```diff
@@ -800,6 +801,82 @@ interface CustomSelectItem {
 Out of the box, the `Select` component uses React Magma-styled iconography for the button used to clear the selection and the caret for the dropdown.
 However, these can be overridden by providing custom components to the `components` prop, in an object containing a `ClearIndicator` and a `DropdownIndicator`.
 
+### Custom Item Component
+
+By default each item will be rendered with the `itemToString` value in a styled `li`. If you would like to add more information or
+would just like to be able to fully customize the look of your items you can pass in an `Item` component that uses the `props` that
```
avanslaars, line 807: I want to make sure the props that are passed get documented. I looked at the generated `properties.json` and it didn't look right. I might not be understanding that output, so if it's working, great. I just want to make sure we're not leaving folks guessing.

packages/react-magma-docs/src/data/properties.json
rcgillet: should be taken out.

packages/react-magma-docs/src/pages/api/select.mdx
rcgillet: Right now select and combobox do not have generated props tables. Those will be added in a different PR. I will check the props tables built here to make sure it reflects what is there now.

joasmi723 (approved):

packages/react-magma-docs/src/data/properties.json
rcgillet: I'm sorry, I was wrong. This is being auto-generated but not yet being used. I don't believe there is a way to stop generating some of the files that haven't had the typedoc info put in yet so you will see this, but the docs page does not use the generated table.

avanslaars (approved):

PR #27, opened 2020-12-14
feat(select-combobox): add a max-height for select and combobox items…

rcgillet:
… list menu

Add a prop named itemsListMaxHeight as well as a theme value for the max heights to be overwritten


packages/react-magma-docs/src/pages/api/select.mdx
```diff
@@ -1,6 +1,7 @@
----
+m---
```
joasmi723, line 1:
```suggestion
---
```

joasmi723: Typo here was causing the page not to show up in the docs nav

packages/react-magma-docs/src/pages/api/combobox.mdx
```diff
@@ -1284,6 +1284,13 @@ corresponds to a `stateChangeTypes` property. The list of all the possible types
       description:
         'Selectable options.  Can be an array of strings or objects.',
     },
+    itemsListMaxHeight: {
```
joasmi723, line 1287: Could you also add working examples of this to the docs?

packages/react-magma-docs/src/pages/api/select.mdx
```diff
@@ -1075,6 +1076,13 @@ corresponds to a `stateChangeTypes` property. The list of all the possible types
       required: false,
       description: 'Selectable options.  Can be an array of strings or objects',
     },
+    itemsListMaxHeight: {
```
joasmi723, line 1079: Could you also add working examples of this to the docs?

joasmi723 (changes requested):

joasmi723 (approved):

PR #29, opened 2020-12-15
ci(netlify): preview

aaronmaturen:


packages/react-magma-dom/README.md
```diff
@@ -0,0 +1,316 @@
+# React magma
+
+## Known Issues
```
avanslaars, line 3: Parts of the readme that were removed in #25 are back

aaronmaturen: Just noticed this is the react-magma-dom readme... it gets clobbered by the top level one [on publish](https://github.com/cengage/react-magma/blob/d33dc7ec958bae0e8aca248ce2cf176c1bf11a51/package.json#L26)

We want the readme at the top level for GitHub ... but inside the published package for npm 🤔

aaronmaturen:
creating a new branch from `dev`

PR #38, opened 2020-12-17
docs(simplepropstable): remove SimplePropsTable import in docs pages

rcgillet:
…where it is no longer used


aaronmaturen (approved):
https://github.com/cengage/react-magma/blob/dev/packages/react-magma-docs/src/components/layout.js#L108-L116

Technically none of them need the SimplePropTable since it gets passed in by default

PR #46, opened 2020-12-23
Docs/css

joasmi723:
This PR is mostly changes to the docs, moving more styles out of css files and into react components. 

It does include a few changes to react-magma-dom, including where table font styles are set and adding more styles to the GlobalStyles component.

Some of the layout of the docs will be slightly different (most notably the left nav) to align with the new spacing and font scales in the magma theme.


packages/react-magma-docs/gatsby-ssr.js
```diff
@@ -23,6 +23,17 @@ const StyledSkipLink = styled(SkipLink)`
   }
 `;
 
+const MainContainer = styled.div`
```
aaronmaturen, line 26: Since `MainContainer` is used in multiple spots should it be pulled out into it's own file and then we can import it in both places?

packages/react-magma-docs/src/pages/design/alerts.mdx
```diff
@@ -26,7 +27,7 @@ Use alerts to inform users of updates, changes to system status, or as feedback
   <img src="../../images/alerts/alert-intro-image.png" alt=" " />
 </figure>
 <span class="horizontal-spacer"></span>
```
aaronmaturen, line 29: if there isn't a divider without this `span.horizontal-spacer` before it... couldn't we beef up the top margin on the `hr` and call it good ¯\_(ツ)_/¯

packages/react-magma-docs/src/pages/design/selection-controls.mdx
```diff
@@ -26,7 +27,7 @@ Selection controls allow users to complete tasks that involve making choices suc
     <p>Selection controls</p>
   </figcaption>
 </figure>
-<hr class="article-divider" />
+<Divider />
```
aaronmaturen, line 30: oh man! a divider without the `span.horizontal-spacer` 😂

packages/react-magma-docs/src/pages/design/selection-controls.mdx
```diff
@@ -159,7 +160,7 @@ Sometimes you have to put those checkboxes on a dark background, and that's when
     </div>
   </div>
 </div>
-<hr class="article-divider" />
+<Divider />
```
aaronmaturen, line 163: and another one!

packages/react-magma-docs/src/pages/design/typography.mdx
```diff
@@ -775,7 +776,7 @@ Another advantage to using the presets above is that Responsive behavior is alre
   <img src="../../images/typography/responsive-example.png" alt=" " />
 </figure>
 <span class="horizontal-spacer"></span>
-<hr class="article-divider" />
+<Divider />
```
aaronmaturen, line 779: I was going to say it looked like all the pages in design didn't have `span.horizontal-spacer` and then this one did 🤔

aaronmaturen (approved):
I don't have strong opinions on any of my comments... they're just the thoughts I had :D

aaronmaturen:
react-magma-docs: error Building static HTML failed for path "/api/announce/"
react-magma-docs:   289 |   });
react-magma-docs:   290 |   // eslint-disable-next-line no-new-func
react-magma-docs: > 291 |   var res = new (Function.prototype.bind.apply(Function, [null].concat(['_poly', 'React'], scopeKeys, [code])))();
react-magma-docs:       | ^
react-magma-docs:   292 |   return res.apply(undefined, [_poly, React].concat(scopeValues));
react-magma-docs:   293 | };
react-magma-docs:   294 | 
react-magma-docs: 
react-magma-docs:   WebpackError: ReferenceError: Divider is not defined





New
14:01
Ryan Gillett:magma: welp
14:01
you know what she could do
14:01
put divider in the layout
14:01
unless you already told her to do that
14:01
Aaron Maturen if you add divider here: https://github.com/cengage/react-magma/blob/dev/packages/react-magma-docs/src/components/layout.js#L102-L120
14:02
you get it magiccally and you don't need to import it everywhere
14:02
I don't know how to comment on a file in a PR that hasn't changed
14:04
in fact...
14:04
if you did hr: Divider mdx would map -- to <Divider/> for you :smile:
14:05
Ryan Gillett:magma: yea
14:05
Aaron Maturen then you can be all markdown
14:05
Ryan Gillett:magma: that would be even better

packages/react-magma-docs/gatsby-ssr.js
joasmi723: Yes, yes it definitely should be :). I will make that change

packages/react-magma-docs/src/pages/design/alerts.mdx
joasmi723: Good call!  I'll look into that --  I wasn't sure where all horizontal-spacer was used.

packages/react-magma-docs/src/pages/design/selection-controls.mdx
joasmi723: Womp womp.  It _should_ probably be consistent though -- not sure if that was intentional.

packages/react-magma-docs/src/pages/design/typography.mdx
joasmi723: Yeah, again, that might not be intentional.  I like the idea of adding the space to the divider.

joasmi723:
Aaron Maturen  1:08 PM
you can either pass new Components in and use them by name, or if its something that md maps to, you can use the thing they map to
New
1:09
like how, SimplePropsTable is there... you don't have to import that everywhere to use it :smile: you just use it
1:09
the random function after it is how the magic props just show up

PR #60, opened 2021-01-04
fix(button, hyperlink): remove css prop

joasmi723:


packages/react-magma-dom/src/components/Hyperlink/index.tsx
```diff
@@ -73,16 +78,20 @@ export const Hyperlink = React.forwardRef<HTMLAnchorElement, HyperlinkProps>(
         </ClassNames>
       );
     } else {
+      const LinkStyledAsButton = BaseStyledButton.withComponent('a');
+      const HyperlinkComponent =
+        styledAs === 'Button' ? LinkStyledAsButton : StyledLink;
```
avanslaars, line 83: In a future PR, we should make this an enum to be consistent with all the other constrained props.

packages/react-magma-dom/src/components/StyledButton/index.tsx
```diff
@@ -19,17 +19,15 @@ import { ThemeContext } from '../../theme/ThemeContext';
...
   iconOnly?: boolean;
-  to?: string;
 }
 
 export const buttonStyles = props => css`
   align-items: center;
   background: ${buildButtonBackground(props)};
   border: ${props.variant === 'outline' ||
-  (props.variant === 'solid' && props.color === 'secondary' && !props.isInverse)
+  (props.variant !== 'link' && props.color === 'secondary' && !props.isInverse)
```
avanslaars, line 30: These should be updated in a future PR to use the enum values in the comparisons.

avanslaars (approved):
LGTM

PR #70, opened 2021-01-07
Docs/codesandbox examples

aaronmaturen:


aaronmaturen:
poked around a little bit.. seems https://docs-preview-70--upbeat-sinoussi-f675aa.netlify.app/api/select/ makes chrome really angry... my guess is too many examples?

aaronmaturen:
That works better for selects now...

PR #82, opened 2021-01-13
fix(select-combo-types): properly allow the usage of single or multi select and combobox

rcgillet:


avanslaars (approved):
This all makes sense to me. Assuming it clears up the issue that brought about the change, I'm good with this.

PR #84, opened 2021-01-14
fix(dropdown): fix divider color

joasmi723:


orion-cengage (approved):
Everything is looking good to me now.

PR #85, opened 2021-01-15
feat(form): create a generic Form component

aaronmaturen:


packages/react-magma-dom/src/components/Form/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { AlertVariant } from '../AlertBase';
+import { Heading } from '../Heading';
+import { Paragraph } from '../Paragraph';
+
+/**
+ * @children required
+ */
+export interface FormProps extends React.FormHTMLAttributes<HTMLFormElement> {
+  /**
+   * Form action buttons are supposed to be placed here
```
joasmi723, line 12:
```suggestion
   * React Node containing the form action buttons; will appear under the form fields
```

PR #103, opened 2021-01-20
docs(docs): style leadparagraph as first paragraph sibling of the Doc…

aaronmaturen:
…sHeading


aaronmaturen:
I know styling siblings to a react-element _might_ not be reacty... but it's also in markdown and I'd rather the markdown be sane :D

joasmi723 (approved):

PR #104, opened 2021-01-20
Fix/tooltip arrow

joasmi723:


rcgillet (changes requested):
`TooltipPosition` needs to be exported from `index.ts` so that people can import it to use it. But also, keep the `EnumTooltipPosition` for now.

rcgillet (approved):

PR #106, opened 2021-01-22
Docs/contributing and readme

joasmi723:


README.md
```diff
@@ -118,47 +121,105 @@ your editor right next to your code.
...
 
-Be sure to add in the navigation details at the top of the page in the front matter.
+Add in the navigation details at the top of the page in the front matter.
 
 ```yaml
 ---
+pageTitle: '{ComponentName} API'
 title: '{ComponentName}'
-order: 2
+props:
```
aaronmaturen, line 139: This might be sufficient, but I'll nitpick that it's only `{ComponentName}Props` by convention and some components have more than one Interface/type that we export... something like card ¯\_(ツ)_/¯

aaronmaturen: I think this would be good enough to just show it, but we're missing a blurb to say add `<{ComponentName}Props/>` at the end of the page

README.md
```diff
@@ -118,47 +121,105 @@ your editor right next to your code.
...
-and import a link to the design guidelines with the correct route
+Import the `DocsHeading` and `ScopeableLayout` components. You may also need other things, such as components from `react-magma-dom` to use outside of examples (such as an `Alert`), or the`Link` from `gatsby` to link to other pages.
 
 ```js
 import { DocsHeading } from '../../components/DocsHeading';
+import { ScopeableLayout } from '../../components/layout';
+```
 
-<DocsHeading to="/design/{component}/" type="design">
+Use the `DocsHeading` component to create a page heading and link to the respective Design Guidelines page if one exists.
```
aaronmaturen, line 151: we can now do this with markdown: 

`[{ComponentName}]("/design/{component} 'api')`

README.md
aaronmaturen: I guess we talk about this further down 😂

aaronmaturen:
Made two comments... I don't think either of them would stop me from approving it :D

PR #108, opened 2021-01-22
docs(left nav): add github link

joasmi723:


orion-cengage (approved):
Link looks and works as expected.

Issue #34, opened 2020-12-16
Timepicker FF/NVDA screenreader issue


agenova-cengage:
**Is your feature request related to a problem? Please describe.**
Screenreader is not reading subsequent values when adjusting the timepicker value(hour/minute) fields.
**Screencast**: NVDA(2020.3)+FF: http://somup.com/cYl6QR2tKm
**Note**: Issue is **NOT** reproducible with JAWS+Chrome combination.

**Describe the solution you'd like**
NVDA on FF should readout subsequent values when adjusting the timepicker value(hour/minute) fields.

**Describe alternatives you've considered**
N/A

rcgillet:
The issue seems to be that there is a `placeholder` on the `input`. The `input` is of type `number` which does not appear to be supported: https://www.deque.com/blog/accessible-forms-the-problem-with-placeholders/.

We are going to need some discussion on what we want to do here. Is there an option where we can show something that looks like a placeholder? Should we get rid of the placeholder altogether? etc.

gaussant:
Try putting `aria-live="assertive"` on the `<input>` that changes the time value...
`<input type="text" disabled="" value="05:02 AM" aria-invalid="false" style="background:#FFFFFF;border-color:#8f8f8f;cursor:default;width:144px" class="css-n7k5b0" id="3f9b0c18-6859-4a3a-b477-abdabd422a07" aria-live="assertive">`

joasmi723:
The suggestion from @gaussant didn't solve the issue in FF/NVDA.  Still looking for a solution.

tasnook:
Hi All, 

I have confirmed that removing the placeholder attribute resolves the issue (NVDA/FF) without the need to add `aria-live`. NVDA/Chrome and NVDA/Edge work as is with the placeholder text.

### **Three Possible Remediations:**

1. Add an `onfocus` event listener that calls the `removeAttribute` method to remove the placeholder attribute entirely upon user interaction. You'd have to decide whether or not to replaced it `onBlur` if the user does *not* make a selection. This solution is experimental and will need to be validated with NVDA testing.
2. You could take an approach somewhat similar to the popular floating text labels that swap CSS classes after user interaction. In this case, the element should always have `aria-hidden="true"` and the class during/after user interaction should apply `{ display: none; }`
[...]

PR #109, opened 2021-01-22
fix(timepicker): add aria-live to timepicker inputs

joasmi723:
This is the fix suggested from Perkins for https://github.com/cengage/react-magma/issues/34

Not tested on Windows yet (NVDA and Jaws), only tested on Mac VoiceOver.


aaronmaturen:
didnt work on nvda in firefox

PR #125, opened 2021-01-27
Feat/select styles

joasmi723:


orion-cengage:
Feedback:

1. I don't see anything about the left-aligned label in the Docs preview URL, but I do see it in the Storybook.
2. The dropdown should align with the left and right sides of the input, not the label
image
3. Perhaps connected, I'm not seeing a large gap between the input and the dropdown on the regular top-label selects
image

packages/react-magma-dom/src/components/Input/index.tsx
```diff
@@ -3,7 +3,7 @@ import styled from '../../theme/styled';
 import { InputBase, InputBaseProps, InputSize } from '../InputBase';
 import { InputMessage } from './InputMessage';
 import { Label } from '../Label';
-import { useGenerateId } from '../../utils';
+import { useGenerateId, Omit } from '../../utils';
```
aaronmaturen, line 6: is `Omit` being used?

packages/react-magma-dom/src/components/Select/index.tsx
```diff
@@ -8,9 +8,10 @@ import {
 
 import { Select as InternalSelect } from './Select';
 import { MultiSelect } from './MultiSelect';
-import { InputMessage } from '../Input/InputMessage';
+//import { InputMessage } from '../Input/InputMessage';
```
aaronmaturen, line 11: should we just delete this line?

packages/react-magma-dom/src/components/Input/index.tsx
joasmi723: Yeah, it's in the props interface.  I noticed an error in that file, and it looked like importing `Omit` was what was missing.

joasmi723: Yeah, it's in the props interface. I noticed an error in that file, and it looked like importing Omit was what was missing.

packages/react-magma-dom/src/components/Select/index.tsx
joasmi723: Yep, we sure should!  Will do that.

joasmi723:
> Feedback:
> 
> 1. I don't see anything about the left-aligned label in the Docs preview URL, but I do see it in the Storybook.  Is that expected?

This has now been added

> 2. The dropdown should align with the left and right sides of the input, not the label
>    image

This is hard to do because of how the component is structured.  As discussed, we will look into doing this as a future enhancement.

> 3. Perhaps connected, I'm seeing a large gap between the input and the dropdown on the regular top-label selects
>    image

Fixed

aaronmaturen (approved):

orion-cengage (approved):

PR #128, opened 2021-02-01
Feature/new icons

aaronmaturen:


aaronmaturen:
	All the icon names have to  be unique and there were two Android icons so I just deleted one
	A react component can’t start with a number so the ones that did (like 4K and 5G) I preface with the category
	We could preface all the icons with a category to get rid of collisions across groups
	We’ve got to update docs/storybook to use icons that actually exist
	Since we have to make changes for the new icon sizes, maybe we want to just update dom to use the new icons in lieu of aliases

orion-cengage:
Looking really good so far.  Here's a little feedback:
1. Not sure why the "Av" section is displaying first.  Alphabetically it should come after the "Alert" section.
2. What's the reason for appending "Icon" to the end of the names?  Is that really necessary?
3. I see the underscores were removed from the original file names.  Is that necessary?  Some of those icon names have two or three words, plus "Icon" at the end, and not having a separator like an underscore makes it way harder to read or scan quickly.

orion-cengage:
Looks like the holes in the "P" and "R" for "WlPracticeItIcon" and "WlReadyIcon" are missing, making them more filled in than they should be.

image

Do you need different svg files for those?  I'm not sure what the problem with the existing ones might be.

aaronmaturen:
> 1. Not sure why the "Av" section is displaying first.  Alphabetically it should come after the "Alert" section.

That's an easy fix to sort them :D 

> 2. What's the reason for appending "Icon" to the end of the names?  Is that really necessary?

They're all icons and react components ... for something like `<Report>` I'd assume that was a report component ... not an icon. 

> 3. I see the underscores were removed from the original file names.  Is that necessary?  Some of those icon names have two or three words, plus "Icon" at the end, and not having a separator like an underscore makes it way harder to read or scan quickly.

Since they're all react components they have to be `PascalCase` or `SCREAMING_SNAKE_CASE`

orion-cengage:
I see what you're saying about the icon naming.  There must be a way to display a friendly name in order to make the documentation more friendly.  Maybe we could look at how Carbon is doing it for a future update to the docs?  They're using React for the icon itself, but displaying a friendly name.

image

https://www.carbondesignsystem.com/guidelines/icons/library

joasmi723:
> I see what you're saying about the icon naming. There must be a way to display a friendly name in order to make the documentation more friendly. Maybe we could look at how Carbon is doing it for a future update to the docs? They're using React for the icon itself, but displaying a friendly name.
> 
> https://www.carbondesignsystem.com/guidelines/icons/library

For what it's worth, I agree with Aaron about the naming.  I really like having `Icon` as a part of the names to differentiate them from other components, and I agree about the casing.  When I'm looking at documentation, I like the name on the screen to be the actual component name, just so I can easily copy-and-paste.  That's more intuitive to me than using an icon like in the Carbon example.  Just my $0.02.

orion-cengage:
Ah, but they also make it easy to copy the actual code with a single click.  Seems pretty slick to me.  Anyway, this is maybe something we can look at in the future.  

For now, once Aaron sorts the groups so the "AV" section comes after "Alert", I think the new icons are ready to implement into the rest of the components.

packages/react-magma-icons/src/icons/Aliases/AlertIcon.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+import { IconProps } from '../../IconProps';
+import { ExpandMoreIcon } from '../Navigation/ExpandMoreIcon';
+export const AlertIcon = (props: IconProps) => {
+  console.warn('AlertIcon has been deprecated');
```
avanslaars, line 5: These all need to be wrapped in a check for the process env and only run in dev mode. This will prevent the warnings in production mode and in test output and will allow rollup to tree-shake this console statement out of production builds

orion-cengage:
Breadcrumb component: The margin on the left and right of the chevron needs to be 4px instead of 8px.

orion-cengage:
Alerts components: The icons on success, danger, and warning are incorrect.  They should be, "CheckCircleIcon", "WarningIcon" and "ErrorIcon".

image

orion-cengage:
The top and bottom padding on **Banner alerts** needs to be 12px instead of 16px.

orion-cengage:
The X in the auto-dismissing toast close button isn't properly centered within the ring.  

image

orion-cengage:
The sort arrows on tables aren't quite vertically aligned properly.  Moving the icon down another 2px seems to help.  Or some flexbox could probably fix it too.

image

orion-cengage:
Tabs component feedback:  Icon needs to be 20px so it matches the line-height of the text label

orion-cengage:
On search and date picker inputs, can we please make the icon on the right side 24px instead of 20px?  This should go for all inputs with icons, that way they are consistent with buttons, which they often get paired with.

orion-cengage:
Date picker feedback:

- Close button should be medium size instead of small
- The arrow icons should use "ArrowBackIcon" and "ArrowForwardIcon"
- Month + Year label should be 16px/24px
- Help icon should be small, not medium
- Arrow buttons should be closer to the edges so the edge of their containers align with calendar below.
image

orion-cengage:
Time picker input: Please move the clock icon to the left so it is 8px from the left edge instead of 12px.  And I think the time itself would move over a little too as a result.

image

orion-cengage:
> Time picker input: Please move the clock icon to the left so it is 8px from the left edge instead of 12px. And I think the time itself would move over a little too as a result.
> 
> image

Verified this is now fixed.

orion-cengage:
> Date picker feedback:
> 
> * Close button should be medium size instead of small
> * The arrow icons should use "ArrowBackIcon" and "ArrowForwardIcon"
> * Month + Year label should be 16px/24px
> * Help icon should be small, not medium
> * Arrow buttons should be closer to the edges so the edge of their containers align with calendar below.
>   image

All changes have been verified except the close button.  It's still set as small, and needs to be medium/default size.

orion-cengage:
Please remove the margins from the icon on icon-only tabs

image

orion-cengage:
I found a small issue with the Date Picker.  The close button and right arrow button are now overlapping.  There needs to be 8px of space between the bottom of the close button and the top of that button.

image

PR #141, opened 2021-02-11
chore: changing exports

ccedrone:
I believe I covered all of the components appropriately, but I need the experts.


joasmi723 (approved):

packages/react-magma-dom/src/components/Breadcrumb/index.tsx
```diff
@@ -1,61 +1,2 @@
...
-          <StyledList isInverse={isInverse} ref={ref} theme={theme}>
-            {children}
-          </StyledList>
-        </nav>
-      </BreadCrumbContext.Provider>
-    );
-  }
-);
+export * from './Breadcrumb';
+export * from './Item';
```
aaronmaturen, line 2: Not sure if this matters in any schema of things... should we do `Item` or `BreadcrumbItem` ?

joasmi723: `BreadcrumbItem` would be preferable, but that would be breaking change. We should change it to BreadcrumbItem for v3.

joasmi723: Actually, I _do_ think it should be `BreadcrumbItem`, so I think you're correct.  We'll look at that.

joasmi723: Ah, the component is named BreadcrumbItem, which is how it should be and is how it gets exported.  But the file is named Item.tsx, so I guess that should change.

PR #147, opened 2021-02-16
Feat/table row color

joasmi723:


packages/react-magma-dom/src/components/Table/Table.test.js
```diff
@@ -67,30 +67,30 @@ describe('Table', () => {
...
-
-    expect(getByTestId('row2')).toHaveStyleRule(
-      'background',
-      magma.colors.tone02,
-      {
-        target: ':hover',
-      }
-    );
-  });
+  // it('should render table with hover styles', () => {
```
aaronmaturen, line 70: should this test get deleted?

packages/react-magma-dom/src/components/Table/Table.test.js
```diff
@@ -148,49 +148,49 @@ describe('Table', () => {
...
-
-    expect(getByTestId('row2')).toHaveStyleRule(
-      'background',
-      magma.colors.tint,
-      {
-        target: ':nth-of-type(even)',
-      }
-    );
-  });
+  // it('should render table with inverse styles', () => {
```
aaronmaturen, line 151: diitto?

packages/react-magma-dom/src/components/Table/Table.test.js
joasmi723: Oops, no.  It shouldn't have been checked in commented out.  Will fix.

PR #176, opened 2021-02-26
Feat/flex

joasmi723:


packages/react-magma-dom/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,219 @@
...
+   */
+  direction?: FlexDirection;
+  /**
+   * If true, the component will have the flex container behavior. You should be wrapping items with a container.
+   */
+  isContainer?: boolean;
+  /**
+   * If true, the component will have the flex item behavior. You should be wrapping items with a container.
+   */
+  isItem?: boolean;
```
avanslaars, line 71: Do we need both `isItem` and `isContainer`? I know an item could be both, but could we determine `item` status based on the ancestor nodes or something? so `isContainer` is container behavior only, but if its parent is a Flex container, it also acts as an item? If that isn't reasonable, then I'm wondering if it would make sense to have an enum with 3 options, Container, Item, Both (naming could be improved, but you get the idea).

joasmi723: I can't figure out a good way to make looking at the ancestor nodes work, but I like the idea of an enum.  I'm struggling with naming though.  Is `behavior` a good name for the prop? And if so should the options be `container`, `item` and `both`, or  `container`, `item` and `containeritem`?

Another idea would be to create two components, one called `FlexContainer` and one called `FlexItem`.  The `FlexItem` could also take a boolean `isContainer`.  What are your thoughts on that approach?

For what it's worth, MUI uses the two booleans `item` and `container`.  Several other libraries have two components, usually called `Row` and `Column`.  I haven't found an example of one that uses an enum.

joasmi723: I updated the PR to use the enum `behavior` with options `container`, `item` and `both`.  I used `both` in another enum, so I figured this would keep some consistency.

aaronmaturen (approved):

PR #188, opened 2021-03-10
fix(form group, radio group): make label styles consistent with other…

joasmi723:
… inputs


orion-cengage (approved):
Looks great from my side.

PR #190, opened 2021-03-10
Feature/inverse context

joasmi723:
The Inverse Context allows the `isInverse` prop to be set on a containing component, and will be provided to all components inside of it (unless specified otherwise).  This also introduces a new `Container` component and applies the inverse context to the Forms.


packages/react-magma-dom/src/inverse/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+interface InverseInterface {
+  isInverse?: boolean;
+}
+
+export const InverseContext = React.createContext<InverseInterface>({
+  isInverse: false,
+});
+
+export function getIsInverse(
```
aaronmaturen, line 11: we can probably use the context here instead of making the devs pass it in

aaronmaturen (changes requested):

packages/react-magma-dom/src/inverse/index.ts
joasmi723: Great feedback!  Will change this.

packages/react-magma-dom/src/components/Alert/Alert.stories.tsx
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
 import { Alert, AlertProps } from './index';
-import { Story, Meta } from '@storybook/react/types-6-0';
 import { AlertVariant } from '../AlertBase';
+import { Card } from '../Card';
```
aaronmaturen, line 4: _should_  be able to do `import { Card, CardBody } from '../Card';`

aaronmaturen (approved):

PR #195, opened 2021-03-16
chore: spacing for solid split dropdowns

ccedrone:
Added margin for the solid background split dropdown component. The icon only state padding was changed as well.


packages/react-magma-dom/src/components/Dropdown/DropdownMenuItem.tsx
```diff
@@ -70,7 +70,7 @@ const StyledItem = styled.div<{
 const IconWrapper = styled.span`
   color: ${props => props.theme.colors.neutral03};
   display: inline-flex;
-  margin-right: ${props => props.theme.spaceScale.spacing05};
+  margin-right: '${props => props.theme.spaceScale.spacing05}';
```
joasmi723, line 73: This looks incorrect.  Is this style getting correctly applied?

joasmi723 (approved):

PR #197, opened 2021-03-17
chore: updating hex values throughout application

ccedrone:
Changed hex values from #727272 to #707070 for accessibility.


orion-cengage:
The rgb value on the color card needs to be updated to rgb(112,112,112)

image

orion-cengage (approved):
Looks good.

PR #206, opened 2021-03-23
Feature/icon sizing update

ccedrone:
Updated icon sizing to 16px for clear / close icons where applicable in ticket.


orion-cengage:
Looks like the update you made to the alert also needs to be made to the banner alert.

image

Can we adjust the right-padding on the chip inside the multi-select and combobox?  Change it to 4px instead of 8px.  That way it looks more like this
image
instead of this
image

orion-cengage (approved):
Updates look good to me.

PR #209, opened 2021-03-24
feat: updating styling for inverse errors

ccedrone:
Changes to styling to accommodate inverse errors.

### DIFF
--- a/packages/react-magma-dom/src/components/Input/InputMessage.tsx
+++ b/packages/react-magma-dom/src/components/Input/InputMessage.tsx
@@ -17,10 +17,12 @@ const Message = styled.div<InputMessageProps>`
   align-items: center;
   border-radius: ${props => props.theme.borderRadius};
   color: ${props =>
-    props.isInverse
+    props.isInverse && props.hasError
       ? props.theme.colors.dangerInverse
       : props.hasError
       ? props.theme.colors.danger
+      : props.isInverse && !props.hasError
+      ? props.theme.colors.neutral08
       : props.theme.colors.neutral03};
   display: flex;
   font-size: ${props => props.theme.typeScale.size02.fontSize};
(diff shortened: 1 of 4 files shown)


packages/react-magma-dom/src/components/Input/InputMessage.tsx
```diff
@@ -15,12 +15,11 @@ export interface InputMessageProps
 
 const Message = styled.div<InputMessageProps>`
   align-items: center;
-  background: ${props =>
-    props.isInverse && props.hasError ? props.theme.colors.danger : 'none'};
+  background: none;
```
joasmi723, line 18:
```suggestion
```

joasmi723: I think we can get rid of this line entirely, since the browser default is no background

packages/react-magma-dom/src/components/Input/InputMessage.tsx
```diff
@@ -32,10 +31,7 @@ const Message = styled.div<InputMessageProps>`
       ? props.theme.spaceScale.spacing03
       : props.theme.spaceScale.spacing02};
   min-height: ${props => props.theme.spaceScale.spacing06};
-  padding: ${props =>
-    props.isInverse && props.hasError
-      ? `${props.theme.spaceScale.spacing02} ${props.theme.spaceScale.spacing03}`
-      : '0'};
+  padding: 0;
```
joasmi723, line 34:
```suggestion
```

joasmi723: Same here -- We should be able to get rid of this line entirely I think.

joasmi723 (approved):

orion-cengage:
Helper text isn't supposed to be pink.  Only error messages.

image

orion-cengage (approved):
Updates look good to me.

PR #210, opened 2021-03-24
fix(tabs)

joasmi723:
- Fix a11y error by removing scroll buttons completely for screen readers
- Adjust tab scrolling so items aren't obscured by scroll buttons as much
- Fix circular dependencies for tabs

### DIFF
--- a/packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
+++ b/packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
@@ -2,9 +2,9 @@ import * as React from 'react';
 import styled from '@emotion/styled';
 import { jsx } from '@emotion/core';
 
-import { NavTabsContext } from './';
+import { NavTabsContext } from './NavTabs';
 import { StyledTabsChild, StyledIcon, TabStyles } from '../Tabs/Tab';
-import { TabsIconPosition } from '../Tabs';
+import { TabsIconPosition } from '../Tabs/Tabs';
 import { TabsOrientation } from '../Tabs/shared';
 import { ThemeContext } from '../../theme/ThemeContext';
 import { omit, XOR } from '../../utils';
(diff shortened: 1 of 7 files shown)


orion-cengage (approved):
Tabs are working way better and the a11y issue is gone.  Joanne is going to remove the 4px offset from the scroll buttons and then we're good.

packages/react-magma-dom/src/components/Tabs/utils.ts
```diff
@@ -13,6 +13,8 @@ export function useTabsMeta(theme, orientation, backgroundColor, isInverse) {
   const scrollStart = vertical ? 'scrollTop' : 'scrollLeft';
   const clientSize = vertical ? 'clientHeight' : 'clientWidth';
 
+  const approxTabSize = vertical ? 80 : 120;
```
aaronmaturen, line 16: should these be in the theme?

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
```diff
@@ -2,9 +2,9 @@ import * as React from 'react';
 import styled from '@emotion/styled';
 import { jsx } from '@emotion/core';
 
-import { NavTabsContext } from './';
+import { NavTabsContext } from './NavTabs';
 import { StyledTabsChild, StyledIcon, TabStyles } from '../Tabs/Tab';
-import { TabsIconPosition } from '../Tabs';
+import { TabsIconPosition } from '../Tabs/Tabs';
```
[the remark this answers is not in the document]
joasmi723: Yep, I think you're right!  Will change.

packages/react-magma-dom/src/components/Tabs/utils.ts
joasmi723: Oh!  Good call.  I hated hard-coding them, but thought it would be too much to calculate them dynamically.  Putting them in the theme might be a good compromise.

PR #224, opened 2021-04-05
Feature/isloading

aaronmaturen:


website/react-magma-docs/src/pages/api/button.mdx
```diff
@@ -140,6 +140,24 @@ export function Example() {
 }
 ```
 
+## Loading
+
+When a button is in a loading state, the label will be replaced with a loading spinner.
```
joasmi723, line 145:
```suggestion
When a button is in a loading state, the label can be replaced with a loading spinner using the `isLoading` prop.
```

joasmi723: Can you add a similar example to the IconButton mdx page as well?

packages/react-magma-dom/src/components/StyledButton/index.tsx
```diff
@@ -17,11 +17,13 @@ import {
...
 import { ThemeContext } from '../../theme/ThemeContext';
-import { ButtonProps, ButtonType } from '../Button';
+import { ButtonType, ButtonProps, ButtonSize, ButtonVariant } from '../Button';
+import { Spinner } from '../Spinner';
 
-interface StyledButtonProps extends ButtonProps {
+export interface StyledButtonProps extends ButtonProps {
   href?: string;
   iconOnly?: boolean;
+  isLoading?: boolean;
```
joasmi723, line 26: I _think_ the `isLoading` prop should be specified in the `BaseButtonProps` or `ButtonStyles` interface instead and extended from here. 

This seems to be working though (but I'm not sure why), except for the fact that the prop isn't showing up in the props table.

joasmi723: Yes, I double checked.  This prop should be removed here and added to the `ButtonStyles` interface in Button/index.tsx.  When you move it there, make sure you add a comment above it that describes the prop for the props table.  For example:

```
/**
  *  Set the button to a loading state
  * @default false
  */
  isLoading?: boolean;
```

Thanks!

jamiesmyln: Ok, it is now in Button/index.tsx.

packages/react-magma-dom/src/components/Button/index.tsx
```diff
@@ -56,6 +56,11 @@ export interface ButtonStyles {
    * Defines the border radius
    * @default ButtonShape.fill
    */
+  /**
```
joasmi723, line 59: Looks like you broke up the description of the `shape` prop with the prop itself.

joasmi723 (approved):

PR #226, opened 2021-04-08
Feature/pagination component

ccedrone:
Checking this in for Orion's styling review.


orion-cengage (changes requested):
1. When the arrow buttons are disabled, it looks like the background might be fine, but the icon isn't changing to its much lighter disabled color
image

2. For the inverse, the background color of the component should not be set on the outlined buttons.  They should be using "outline" so the backgrounds are transparent, therefore the color of the background comes through.
image

3.  On inverse, it doesn't look like the arrow buttons are disabling at all
image

Link for inverse buttons: https://react-magma.cengage.com/version/2.3.9/api/button/#inverse

orion-cengage:
Another update:
On the large size, the font-size of the numbers should be 20px with a line-height of 32px (body/large)

packages/react-magma-dom/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,238 @@
...
+  /**
+   * Default Page allows the user to toggle which Pagination button is active by default.
+   */
+  defaultPage?: number;
+  /**
+   * isInverse allows for inversing the theme pallette.
+   */
+  isInverse?: boolean;
+
+  onPageChange?: (pageNum: number) => void;
```
joasmi723, line 30:
```suggestion
  /**
   * Event that fires when the active page changes
   */
  onPageChange?: (pageNum: number) => void;
```

packages/react-magma-dom/src/components/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,135 @@
...
+    box-shadow: ${hoverBoxShadowColor};
+    outline: 0 !important;
+    outline-offset: 0;
+    overflow: visible;
+    z-index: 1;
+  }
+  &:focus:before {
+    content: '';
+    border: ${props => props.theme.spaceScale.spacing01} solid
+      ${props => props.theme.colors.focus};
```
joasmi723, line 109: The color of the focus style should change, so that we use the `focusInverse` color when the `isInverse` prop is true

packages/react-magma-dom/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,238 @@
...
+   */
+  count: number;
+  /**
+   * Default Page allows the user to toggle which Pagination button is active by default.
+   */
+  defaultPage?: number;
+  /**
+   * isInverse allows for inversing the theme pallette.
+   */
+  isInverse?: boolean;
```
joasmi723, line 28: You can get rid of the description for the `isInverse` prop, because if it's left blank a standard default one gets applied.

orion-cengage (changes requested):
Requested styling fixes/tweaks

1. Font size on regular size pagination should be 16px, not 14px
2. Disabled arrow button should use standard disabled button styling
https://docs-preview-242--upbeat-sinoussi-f675aa.netlify.app/api/button/#disabled
image

image

3. The number on the active page for the inverse version should be primary blue, not secondary dark gray
image

packages/react-magma-dom/src/components/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+      props.theme.colors.primary
+    )}`;
+  }
+}
+
+const StyledPageButton = styled(Button)`
+  border: none;
+  border-top: ${BuildBorder};
+  border-right: ${BuildBorder};
+  border-bottom: ${BuildBorder};
```
joasmi723, line 53: Instead of setting border to none and then calling a `BuildBorder` function on each of the sides, could you get the same effect by doing `border-width: 2px 2px 2px 0;` ?  I think if we do that, we wouldn't need the `BuildBorder` function at all.

orion-cengage (approved):
All requested updates look great.

PR #231, opened 2021-04-12
Feature/input focus

jamiesmyln:


packages/react-magma-dom/src/components/InputBase/index.tsx
```diff
@@ -83,19 +83,31 @@ export interface InputBaseProps
...
 }
 
-const InputWrapper = styled.div`
+const InputWrapper = styled.div<InputBaseProps>`
   align-items: center;
   display: flex;
   flex-shrink: 0;
   position: relative;
+  border: 1px solid;
+  border-color: ${props =>
```
joasmi723, line 92: Moving the styles here seems to work everywhere except for the time picker.  It looks like you will have to make some adjustments to that component so that it works like it did before.

joasmi723: Thanks! The focus state looks better on the date picker now.  However, there still is some weirdness on the date picker error and inverse state.

PR #241, opened 2021-04-15
Feature/input focus

jamiesmyln:
Refactored and improved inputs


packages/react-magma-dom/src/components/TimePicker/index.tsx
```diff
@@ -326,13 +316,14 @@ export const TimePicker = React.forwardRef<HTMLInputElement, TimePickerProps>(
...
             inputStyle={{
               background: `${theme.colors.neutral08}`,
-              borderColor: `${
-                errorMessage
-                  ? theme.colors.danger
-                  : isInverse
-                  ? theme.colors.neutral08
-                  : theme.colors.neutral04
-              }`,
+              // border: '1px solid',
```
joasmi723, line 319: Can we remove these commented lines?  Other than that, this looks good to me.

jamiesmyln: Sure thing

joasmi723 (approved):

orion-cengage:
Can we please make the following adjustments to the SHOW button in order to optimize how the space within the input is being used?

Basically I think changing the button to the small variant will help the most, and maybe some small tweaks to spacing.  Below is an illustration of where the component is now, and where I'd like it to get.

Password Input - Button update

orion-cengage (changes requested):
I requested some changes in the comments.

PR #242, opened 2021-04-19
Feature/dark mode

joasmi723:


orion-cengage:
Here are some styling tweaks I'd like for us to try:

1. Update all borders to rgba(255,255,255, 0.25)
image

2.

PR #245, opened 2021-04-22
fix(input): fix style in input base that was causing the search box i…

joasmi723:
…nside the header to be too big

### DIFF
--- a/.changeset/clean-houses-sniff.md
+++ b/.changeset/clean-houses-sniff.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/clean-houses-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
aaronmaturen, line 2: This should probably be `"react-magma-dom": patch`

aaronmaturen (approved):

.changeset/clean-houses-sniff.md
joasmi723, line 2:
```suggestion
 "react-magma-dom": patch
```

PR #246, opened 2021-04-22
Feature/updating error styles for radios

ccedrone:


packages/react-magma-dom/src/components/Radio/index.tsx
```diff
@@ -81,18 +81,27 @@ function buildRadioIconColor(props) {
...
-    }
     return props.theme.colors.neutral08;
   }
   if (props.isChecked) {
     return props.color;
   }
+  return props.theme.colors.neutral02;
+}
+
+function buildErrorBorder(props) {
```
joasmi723, line 92: The formatting in here gets a little funky.  If you waned to avoid ternaries, you could do something like this:

```
export function buildErrorBorder(props) {
  if (props.hasError) {
    if (props.isInverse) {
      return `0 0 0 2px ${props.theme.colors.dangerInverse}`;
    }
    return `0 0 0 2px ${props.theme.colors.danger}`;
  }
  return `inherit`;
}```

joasmi723 (approved):

PR #252, opened 2021-04-30
feature/transitions

aaronmaturen:


packages/react-magma-dom/src/components/Transition/Transition.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+        />
+      </Transition>
+      <Button onClick={() => setIsOpen(!isOpen)}>
+        {isOpen ? 'hide' : 'show'}
+      </Button>
+    </div>
+  );
+};
+
+export const Testing = ()  => {
```
rcgillet, line 38: what does this do? I couldn't figure it out on the preview.

packages/react-magma-dom/src/components/Transition/Transition.test.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
rcgillet, line 1: This feels a bit empty. Nothing really being tested here.

rcgillet (changes requested):

packages/react-magma-dom/src/components/Transition/useTransition.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
rcgillet, line 1: is the hook used anywhere?

rcgillet (changes requested):

packages/react-magma-dom/src/components/Transition/Transition.tsx
```diff
@@ -0,0 +1,179 @@
...
+ */
+export interface BaseTransitionProps extends HTMLMotionProps<'div'> {
+  /**
+   * If `true`, the element will unmount when `in={false}` and animation is done
+   */
+  unmountOnExit?: boolean;
+  /**
+   * If `true`, the content will animate in
+   */
+  in?: boolean;
```
joasmi723, line 19: The prop name `in` is kind of vague to me -- is it standard in other libraries?

packages/react-magma-dom/src/components/Transition/Transition.tsx
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * If `true` the children will have the `fade` transition applied from the theme.
+   * @deafult false
+   */
+  fade?: boolean;
+  /**
+   * Custom styles to apply to the transition component div.
+   */
+  style?: React.CSSProperties;
```
joasmi723, line 28: Do we need the `style` prop, or do we get it by extending attributes from `div`?

website/react-magma-docs/src/pages/api/transition.mdx
```diff
@@ -0,0 +1,599 @@
...
+    </div>
+  );
+}
+```
+
+# Transition Props
+
+**Any other props supplied will be provided to the wrapping `div` element.**
+
+<TransitionProps />
```
joasmi723, line 599: Props table is not rendering in the preview.  :(

website/react-magma-docs/src/pages/api/transition.mdx
```diff
@@ -0,0 +1,599 @@
...
+      </Transition>
+      <Button onClick={() => setIsOpen(!isOpen)}>
+        {isOpen ? 'hide' : 'show'}
+      </Button>
+    </div>
+  );
+}
+```
+
+### NudgeTop
```
joasmi723, line 264: These examples behave a little oddly and not consistently with each other. The div seems to be growing from just one side when transitioning in.  The NudgeBottom example is always on the screen and the NudgeRight example doesn't show up at all.

I also think it would be helpful to add an explanation of what each transition type actually does.

website/react-magma-docs/src/pages/api/transition.mdx
```diff
@@ -0,0 +1,599 @@
...
+      </div>
+      <Button onClick={() => setIsOpen(!isOpen)}>
+        {isOpen ? 'hide' : 'show'}
+      </Button>
+    </div>
+  );
+}
+```
+
+### SlideTop
```
joasmi723, line 380: Are the Slide transitions always based on fixed positioning?  If so, I would specify this in the docs here.

website/react-magma-docs/src/pages/api/transition.mdx
```diff
@@ -0,0 +1,599 @@
...
+    </div>
+  );
+}
+```
+
+### NudgeTop
+
+```typescript
+import React from 'react';
+import { Transition } from 'react-magma-dom';
```
joasmi723, line 268: Make sure you're importing `Button` from react-magma-dom in all of these examples.

website/react-magma-docs/src/pages/api/transition.mdx
```diff
@@ -0,0 +1,599 @@
...
+
+export function Example() {
+  const [isOpen, setIsOpen] = React.useState(false);
+  return (
+    <div>
+      <div style={{ height: '250px', width: '250px', overflow: 'hidden' }}>
```
joasmi723, line 390: Will the consumer need to add these styles to the outer div?  If so, I would describe this in the docs.  If not, I would leave these styles out.

packages/react-magma-dom/src/components/Transition/Transition.tsx
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * If `true` the children will have the `scale` transition applied from the theme.
+   * @deafult false
+   */
+  scale?: boolean;
+  /**
+   * If `true` the children will have the `collapse` transition applied from the theme.
```
joasmi723, line 35: Is this description correct here?

packages/react-magma-dom/src/components/Transition/Transition.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { AnimatePresence, HTMLMotionProps, motion } from 'framer-motion';
+import { MotionVariants } from '../../theme/components/transition';
+
+import { ThemeContext } from '../../theme/ThemeContext';
+import { ThemeInterface } from '../../theme/magma';
+
+/**
+ * @children required
+ */
+export interface BaseTransitionProps extends HTMLMotionProps<'div'> {
```
joasmi723, line 11: Do we need to let consumers know what these `HTMLMotionProps` props are?

packages/react-magma-dom/src/components/Transition/Transition.stories.tsx
aaronmaturen: That was me leaving stuff in that shouldn't be :(

packages/react-magma-dom/src/components/Transition/Transition.test.js
aaronmaturen: It's actually just the generated tests from plop 😂 I don't know what an actual good test would be 🤔 we could add a check for the css properties for all of them... but wouldn't snapshots already have that same exact case covered...

joasmi723 (approved):

rcgillet (approved):

PR #253, opened 2021-05-03
chore(icons): move out of the monorepo

aaronmaturen:


aaronmaturen:
Just a note: there's no actions for the new icons repo. We'll need those before adding more icons :D

PR #257, opened 2021-05-10
Sliding drawer transition

jamiesmyln:
This PR includes: 

- the Drawer component 
- Adding Transition to Modal for the animation 
-  Datepicker modal styling fix.


packages/react-magma-dom/src/components/Drawer/Drawer.test.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { axe } from 'jest-axe';
+import { Drawer } from '.';
+import { render } from '@testing-library/react';
+
+const TEXT = 'Test Text';
+
+describe('Drawer', () => {
```
joasmi723, line 8: These just look like the default tests, there should probably be more.

packages/react-magma-dom/src/components/Drawer/useDrawer.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
joasmi723, line 1: If we're not using this file, we probably should just delete it.

packages/react-magma-dom/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @children required
+ */
+export interface DrawerProps extends ModalProps {
+  testId?: string;
+  /**
+   * Style properties for the drawer
+   */
+  style?: React.CSSProperties;
```
joasmi723, line 30: I wonder if we should call this property `drawerStyle` to distinguish it from `containerStyle`.  I believe that's what we do in other components that take multiple style props.

packages/react-magma-dom/src/components/Modal/Modal.tsx
```diff
@@ -372,89 +342,87 @@ export const Modal = React.forwardRef<HTMLDivElement, ModalProps>(
...
-                isInverse={false}
-                ref={ref}
-                theme={theme}
-              >
+    return (
+      <>
+        <Global
+          styles={css`
+            html {
+              overflow: hidden;
```
joasmi723, line 350: It looks like this is getting applied even when the modal or drawer isn't open.  As a result, I can't scroll on the modal and drawer docs pages.

website/react-magma-docs/src/pages/api/drawer.mdx
```diff
@@ -0,0 +1,66 @@
...
+A container with a transition
+
+## Basic Usage
+
+```typescript
+import React from 'react';
+import { Drawer } from 'react-magma-dom';
+
+export function Example() {
+  return <Drawer>Drawer content</Drawer>;
```
joasmi723, line 19: These examples don't have a way to trigger the drawer.  Can you add a button like for the storybook stories?

Also, we should probably show an example where we set the drawer position.

joasmi723 (changes requested):

packages/react-magma-dom/src/components/Drawer/Drawer.test.js
jamiesmyln: Added tests to cover the 4 positions

aaronmaturen:
⚠️ No Changeset found 😱

PR #262, opened 2021-05-11
feat(native select): adding the native select component

ccedrone:


packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { useIsInverse } from '../../inverse';
+import { I18nContext } from '../../i18n';
+import { useGenerateId } from '../../utils';
+
+/**
+ * @children required
+ */
+export interface NativeSelectProps extends InputProps {
+  disabled?: boolean;
+  hasError?: boolean;
```
joasmi723, line 21: You shouldn't have to specify a lot of these props here (helperMessage, isInverse, etc) -- they should already be included when you extend the `InputProps`.   

We also shouldn't need a consumer to provide a `hadError` prop.  We should be able to determine this if by whether the `errorMessage` prop is empty.

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -0,0 +1,92 @@
...
+        isInverse={isInverse}
+        ref={ref}
+        style={containerStyle}
+        theme={theme}
+      >
+        <Label htmlFor={id} isInverse={isInverse}>
+          {labelText}
+        </Label>
+
+        <SelectTriggerButton toggleButtonProps={''} disabled={disabled}>
```
joasmi723, line 70: The inverse version isn't getting the correct focus style, and the red border isn't showing up for the error message.  I think you might need to add `isInverse={isInverse}` and `hasError={!!errorMessage}` (or something similar) here.

packages/react-magma-dom/src/components/NativeSelect/useNativeSelect.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
joasmi723, line 1: If you're not using this file, you can just delete it.

packages/react-magma-dom/src/index.ts
```diff
@@ -117,6 +117,7 @@ export {
   LoadingIndicator,
   LoadingIndicatorProps,
 } from './components/LoadingIndicator';
+export { NativeSelect } from './components/NativeSelect';
```
joasmi723, line 120:
```suggestion
export { NativeSelect, NativeSelectProps } from './components/NativeSelect';
```

website/react-magma-docs/src/pages/api/native-select.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+```typescript
+import React from 'react';
+import { NativeSelect } from 'react-magma-dom';
+
+export function Example() {
+  return (
+    <NativeSelect
+      labelText="Select this"
+      hasError
```
joasmi723, line 56: The consumer shouldn't need to specify `hasError` because they're including an `errorMessage` prop.

website/react-magma-docs/src/pages/api/native-select.mdx
```diff
@@ -0,0 +1,187 @@
...
+          <option>Green</option>
+          <option>Blue</option>
+        </NativeSelect>
+      </CardBody>
+    </Card>
+  );
+}
+```
+
+## Inverse with error
```
joasmi723, line 132:
```suggestion
## Inverse with Error
```

website/react-magma-docs/src/pages/api/native-select.mdx
```diff
@@ -0,0 +1,187 @@
...
+          <option>Green</option>
+          <option>Blue</option>
+        </NativeSelect>
+      </CardBody>
+    </Card>
+  );
+}
+```
+
+## Inverse disabled
```
joasmi723, line 111:
```suggestion
## Inverse Disabled
```

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+export const Disabled = Template.bind({});
+Disabled.args = {
+  ...Default.args,
+  disabled: true,
+};
+
+export const HasError = Template.bind({});
+HasError.args = {
+  ...Default.args,
+  hasError: true,
```
joasmi723, line 48: User shouldn't need to specify `hasError` if they are also providing an errorMessage.

joasmi723 (changes requested):

website/react-magma-docs/src/pages/api/native-select.mdx
```diff
@@ -0,0 +1,187 @@
...
+          <option>Red</option>
+          <option>Green</option>
+          <option>Blue</option>
+        </NativeSelect>
+      </CardBody>
+    </Card>
+  );
+}
+```
+
```
joasmi723, line 182: We might also want to show an example with an onChange event, to demonstrate getting the select's value.

website/react-magma-docs/src/pages/api/native-select.mdx
```diff
@@ -0,0 +1,226 @@
...
+# NativeSelect Props
+
+**Any other props supplied will be provided to the wrapping `div` element.**
```
joasmi723, line 224:
```suggestion
**Any other props supplied will be provided to the wrapping `select` element.**
```

joasmi723 (approved):

orion-cengage (approved):
Looks good.

PR #269, opened 2021-05-18
Feature/accordion

joasmi723:


orion-cengage (changes requested):
- Looks like the container around the icon is 30px tall, which is throwing off the vertical centering.
image

- I think it would be helpful on disabled items to use the same disabled cursor we use on disabled buttons
- On the Inverse accordion, the dividing lines should be white at 40% opacity.  This will match the borders we use on inverse tables.
image

orion-cengage (approved):

orion-cengage (changes requested):
- Change bottom-padding of content area from 8px to 12px.  But leave the top padding at 8px.
image

- Add hand cursor on hover to the clickable area (when not disabled) for additional feedback that the area is clickable
- Let's change the weight of the text in the clickable headers to semi-bold for a little bit clearer hierarchy.
image

- Let's force a bit of space between the end of the title and the chevron for situations where they get close to each other.  Probably the same space that's between the chevron and text when the chevron is on the left side.
image

image

jamiesmyln (approved):

orion-cengage (approved):
Looks great

packages/react-magma-dom/src/components/Accordion/useAccordion.ts
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * If true, multiple accordion items may be expanded at once
+   * @default true
+   */
+  isMulti?: boolean;
+  /**
+   * Event that fires when an accordion button is clicked
+   */
+  onExpandedChange?: (event: any) => void;
+  testId?: string;
```
rcgillet, line 32: Don't believe this is needed in the hook because it doesn't deal with rendering.

packages/react-magma-dom/src/components/Accordion/useAccordionButton.ts
```diff
@@ -0,0 +1,131 @@
...
+    onExpandedChange,
+    setExpandedIndex,
+  } = React.useContext(AccordionContext);
+
+  const { buttonId, index, isDisabled, isExpanded, panelId } = React.useContext(
+    AccordionItemContext
+  );
+
+  const handleClick = () => {
+    if (typeof onExpandedChange === 'function') {
```
rcgillet, line 33: can we use this the same way we do elsewhere?  `x && typeof x === 'function' && x()`

packages/react-magma-dom/src/components/Accordion/useAccordionButton.ts
```diff
@@ -0,0 +1,131 @@
...
+        }
+      } else {
+        if (isExpanded) {
+          setExpandedIndex(null);
+        } else {
+          setExpandedIndex(index);
+        }
+      }
+    }
+  };
```
rcgillet, line 52:
```suggestion
    if (!isControlled) {
  const expandedIndex =
    isMulti && isArray(expandedIndex)
      ? isExpanded
        ? expandedIndex.filter((item) => item !== index)
        : expandedIndex.concat([index])
      : isExpanded
      ? null
      : index;
}
```

rcgillet: I didn't test this, but I _think_ this would clean up the `if/else` nesting

packages/react-magma-dom/src/components/Accordion/useAccordionButton.ts
```diff
@@ -0,0 +1,131 @@
...
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    const arrLength = buttonRefArray.current.length;
+
+    switch (event.key) {
+      case 'ArrowDown': {
+        if (index === arrLength - 1) {
+          focusFirst();
+        } else {
+          focusNext();
+        }
```
rcgillet, line 81:
```suggestion
        index === arrLength - 1 ? focusFirst() : focusNext();
```

packages/react-magma-dom/src/components/Accordion/useAccordionButton.ts
```diff
@@ -0,0 +1,131 @@
...
+        break;
+      }
+      case 'ArrowUp': {
+        if (index === 0) {
+          focusLast();
+        } else {
+          focusPrev();
+        }
```
rcgillet, line 89:
```suggestion
        index === 0 ? focusLast() : focusPrev();
```

packages/react-magma-dom/src/components/Accordion/useAccordionItem.ts
```diff
@@ -0,0 +1,66 @@
...
+  testId?: string;
+}
+
+interface AccordionItemContextInterface {
+  buttonId?: string;
+  index?: number;
+  isDisabled?: boolean;
+  isExpanded?: boolean;
+  panelId?: string;
+  setIsExpanded?: any;
```
rcgillet, line 22: there is a type for state updates. 
```suggestion
  setIsExpanded?: React.Dispatch<React.SetStateAction<boolean>>;
```

packages/react-magma-dom/src/components/Accordion/useAccordionItem.ts
```diff
@@ -0,0 +1,66 @@
...
+    isDisabled: false,
+    isExpanded: false,
+    setIsExpanded: () => {},
+  }
+);
+
+export function useAccordionItem(props: UseAccordionItemProps) {
+  const { index, isDisabled } = props;
+
+  const [isExpanded, setIsExpanded] = React.useState(false);
```
rcgillet, line 36:
```suggestion
  const [isExpanded, setIsExpanded] = React.useState<boolean>(false);
```

packages/react-magma-dom/src/components/Accordion/useAccordionItem.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const buttonId = `${idPrefix}_btn`;
+  const panelId = `${idPrefix}_panel`;
+
+  React.useEffect(() => {
+    if (isMulti) {
+      setIsExpanded(isArray(expandedIndex) && expandedIndex.includes(index));
+    } else {
+      setIsExpanded(expandedIndex == index);
+    }
```
rcgillet, line 49:
```suggestion
    const newIsExpanded = isMulti
      ? isArray(expandedIndex) && expandedIndex.includes(index) 
      : expandedIndex == index;
    
    setIsExpanded(newIsExpanded);
```

rcgillet (changes requested):

packages/react-magma-dom/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,161 @@
...
+   */
+  defaultIndex?: number[];
+  /**
+   * Not used in uncontrolled accordion
+   */
+  index?: never;
+  /**
+   * If true, multiple accordion items may be expanded at once
+   */
+  isMulti?: true;
```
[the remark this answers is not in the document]
joasmi723: Actually, I think it is because true is the default.  It'll be required when it's false though.  I will double check.

packages/react-magma-dom/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * @internal
+   */
+  theme?: ThemeInterface;
```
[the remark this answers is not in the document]
joasmi723: Yes, good call!  I'll put the testID in there too.

packages/react-magma-dom/src/components/Accordion/useAccordion.ts
joasmi723: Good call.  I'll put it in the new BaseAccordionProps interface you suggested.

packages/react-magma-dom/src/components/Accordion/useAccordionButton.ts
joasmi723: We sure can!

packages/react-magma-dom/src/components/Accordion/AccordionButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+`;
+
+const TextWrapper = styled.span`
+  flex-grow: 1;
+`;
+
+export const AccordionButton = React.forwardRef<
+  HTMLButtonElement,
+  AccordionButtonProps
+>((props, forwardedRef) => {
```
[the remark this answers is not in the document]
joasmi723: What naming would you suggest as a more descriptive name in the hook?  It seems like we use the `forwardedRef` naming in other places where we use `useForkedRef` but I'm open to other suggestions.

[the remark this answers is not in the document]
rcgillet: I saw that, I back down from this comment :)

rcgillet (approved):

Issue #266, opened 2021-05-13
Сan't remove value in DatePicker if it was already set once time


Maks19:
We can't remove value in **`DatePicker`** if it was already set once time.

Sandbox example - https://codesandbox.io/s/react-magma-example-forked-6ennh?file=/example.tsx:334-343

rcgillet:
For this we should probably add an `isClearable` prop to the `InputBase` component that adds a clear icon to the right of the input and is used to empty out the value. This functionality can then be used in any `input` fields we have in the future. 

Things to consider: 
 - controlled vs. uncontrolled changing of value. (what state the current input is in)
 - custom clear icons
 - clear icon placement in an input that has another icon in it (i.e. Search component)

PR #280, opened 2021-05-26
feat(input): added isClearable prop to input base

jamiesmyln:
re #266

### DIFF
--- a/packages/react-magma-dom/src/components/InputBase/index.tsx
+++ b/packages/react-magma-dom/src/components/InputBase/index.tsx
@@ -241,7 +241,7 @@ const IconWrapper = styled.span<{
     props.iconPosition === 'left' ? props.theme.spaceScale.spacing03 : 'auto'};
   right: ${props =>
     props.iconPosition === 'right' ? props.theme.spaceScale.spacing03 : 'auto'};
-  position: ${props => (props.isClearable ? 'relative' : 'absolute')};
+  position: absolute;
   top: ${props => props.theme.spaceScale.spacing03};
 
   ${props =>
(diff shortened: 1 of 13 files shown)


orion-cengage (changes requested):
We need to make sure the clear button styling is consistent across all the inputs.  The Combobox currently used the round button:
image

While the Search input is using the square button:
image

Since the Date Picker and Search inputs already use square buttons for their main action, let's just use square buttons for the Clear button as well.  This means updating Combobox.

packages/react-magma-dom/src/components/InputBase/index.tsx
```diff
@@ -387,6 +431,28 @@ export const InputBase = React.forwardRef<HTMLInputElement, InputBaseProps>(
             />
           </IconButtonContainer>
         )}
+
+        {isClearable && value && (
+          <IsClearableContainer theme={theme}>
```
joasmi723, line 436: When I tab through the Search field, I notice the Search button gets focused before the X button, which doesn't match what I'm seeing visually.  Can we switch the order of these elements in the dom so it matches the visual?

packages/react-magma-dom/src/components/InputBase/index.tsx
```diff
@@ -387,6 +431,28 @@ export const InputBase = React.forwardRef<HTMLInputElement, InputBaseProps>(
...
+            <IconButton
+              aria-label={'Clear Input'}
```
joasmi723, line 438: We should move this message into the i18n config.

joasmi723 (changes requested):

packages/react-magma-dom/src/components/InputBase/index.tsx
joasmi723: Awesome, this looks great!  

One other thing I hadn't noticed before is that after the clicking the 'clear' icon, nothing is in focus. I think we should return the focus to the search box.  

This PR also needs a changeset.  Everything else looks good to me.

PR #287, opened 2021-06-08
Pagination refactor and datagrid

rcgillet:


orion-cengage:
1. In the section for "Disabled Selectable Row", I don't see any disabled rows in the example

image


2. In the "InverseContext" section, the pagination isn't showing inversed.  I'm pretty sure we had that working before, but if not, I can get a visual design worked up for that quickly.

image

packages/react-magma-dom/scripts/generateJSON.js
```diff
@@ -83,9 +83,11 @@ const filterJson = () => {
 
     return sortObject(
       children
+        .filter(a => a)
         .filter(filterReactDefinitions)
         .filter(filterMotionDefinitions)
         .filter(filterEmotion)
+        .filter(filterExtendedDefinitions(['BasePaginationProps']))
```
aaronmaturen, line 90: I know we only care about this one for now... but we should move it into a tag for the Interface... we do that for ```@children true``` already

packages/react-magma-dom/src/components/Datagrid/Datagrid.tsx
```diff
@@ -72,48 +88,99 @@ export interface DatagridProps extends TableProps {
...
+      defaultSelectedRows = [],
       onHeaderSelect,
       onRowSelect,
       onSelectedRowsChange,
-      pagination,
+      paginationProps = {},
       rows,
-      selectedRows: controlledSelectedRows,
+      selectedRows: selectedRowsProp,
+      hasPagination = true,
```
aaronmaturen, line 129: 😆

packages/react-magma-dom/src/components/Pagination/PageButton.tsx
```diff
@@ -9,7 +9,7 @@ export interface PaginationProps extends ButtonProps {
   isSelected?: boolean;
 }
 
-function typeSize(props) {
+export function typeSize(props) {
```
aaronmaturen, line 12: should this have a more specific name... like `paginationTypeSize` or `pageButtonTypeSize` ?

packages/react-magma-dom/src/hooks/usePagination.ts
```diff
@@ -0,0 +1,204 @@
...
+   */
+  showFirstButton?: boolean;
+  /**
+   * If true, shows the last page button
+   * @default false
+   */
+  showLastButton?: boolean;
+}
+
+export function usePagination(props: Partial<UsePaginationProps> = {}) {
```
aaronmaturen, line 60: This hook should be in `src/components/pagination` ... it's the logic for building a pagination component :D even where it's used in TablePagination... that's really just an example of using the hook to create a custom pagination example

`src/hooks` should be nice generic hooks like `useDataPagination` `useControlled` etc... either to manipulate data... or something more at a global app level like media queries or clipboard management... things that stand alone without being the logic to a component 🤔

aaronmaturen:
also, changeset

orion-cengage (approved):

Issue #260, opened 2021-05-11
add new color: primaryInverse


orion-cengage:
Purpose: This new color will serve as a variation of primary blue that we can use on dark backgrounds.  The first applications of this color will be on the new Tags component, and inverse hyperlinks.

1. New color: #70CDFF
2. On the docs page for Colors, let's create a new group for "Inverse Colors".  Then we can include this new color there, and move **dangerInverse** and **successInverse** to this section as well.
3. Remove line from Feedback colors description: "The successInverse and dangerInverse are designed to be used against dark backgrounds."
4. Description for new Inverse section can be: "The inverse colors are designed to be used against dark backgrounds."

ccedrone:
This was added into the development of Tag and all set: https://github.com/cengage/react-magma/pull/289

PR #289, opened 2021-06-09
feat(tag)

ccedrone:


orion-cengage (changes requested):

1. I'm not seeing a section on the API docs page for the small sizes

2. Change the labels on the examples to just: Default, Primary, Low Contrast, High Contrast
image

3. Change labels on examples to just: Disabled, Disabled Low Contrast
image

4. Change labels on examples to just: Inverse Default, Inverse Primary, Inverse Low Contrast, Inverse High Contrast
image

5. Change labels on examples to just: Inverse Disabled, Inverse Disabled Low Contrast
image

6. I think the titles of these sections should be "Tag With Icon" and "Deletable Tag With Icon"
image
image

7. Small tag with delete button: The right margin on the close button icon should be 4px, not 2px
image

orion-cengage:
@ccedrone I forgot one update to make.  The API Docs page needs the button in the upper-right that goes to the Design Guidelines page.

packages/react-magma-dom/src/i18n/default.ts
```diff
@@ -225,6 +225,9 @@ export const defaultI18n: I18nInterface = {
     nextButtonLabel: 'Scroll tabs forward',
     previousButtonLabel: 'Scroll tabs back',
   },
+  tag: {
+    deleteAriaLabel: 'Delete tag',
```
rcgillet, line 229: This should be more descriptive. We do something in `Combobox` to use the `labelText`.

`'reset selection for {labelText}. {selectedItem} is selected',`. So here we could do `delete {labelText} tag`. Then in the `Tag`  component file you can do something like this ```const deleteAriaLabel = i18n.tag.deleteAriaLabel.replace(/\{labelText\}/g, labelText)```

website/react-magma-docs/src/pages/api/tag.mdx
```diff
@@ -0,0 +1,307 @@
...
+    </I18nContext.Provider>
+  );
+}
+```
+
+## Tag Props
+
+**Any other props supplied will be provided to the wrapping `div` element.**
+
+<DeleteableTagProps />
```
rcgillet, line 306: Above each of the `props` tables put `### Deletable Tag Props` or `### Clickable Tag Props` so they know what they are looking at.

rcgillet (changes requested):

orion-cengage (approved):
Styling is looking good.

rcgillet (approved):

ccedrone:
In the development of this component, pimaryInverse was needed. This ticket was resolved as part of the work: https://github.com/cengage/react-magma/issues/260

PR #290, opened 2021-06-11
Feature/victory line

joasmi723:


orion-cengage:
### Styling Feedback

- I'm not sure we want the title for the viz to be an H5, because that may not make sense within the structure of the page it's on.  Perhaps we should instead use paragraph or span and style it accordingly.
- Would it be possible to make the popover with multiple data points only appear when you hover over the actual line. Currently, the hoverable areas are so large, that the chart is almost always obscured by a popover.
image

- Please change the text within the popovers to normal instead of semi-bold
image

- Too much space after the last item.  Should only be 8px.
image

- Can we increase the width of the popover to avoid things wrapping so much?  It would be nice if it could grow as necessary with a max-width of 275px.  But if we have to just explicitly set the width, then let's go with 275px. 
- Also, can we top align the color block and corresponding text instead of middle align?
image

- Can we include the color block even when it's only showing one?
image

- There's some weird indentation happening here
image

- The color blocks on the Spending Revenue example are weird sizes
image

- The popover on the History of Texas example isn't following the cursor
image

- Can we have a consistent 16px of space between the y-axis title and labels?  Currently it looks like that area is set to a specific value so the gap is inconsistent.
image
image

rcgillet:
> ### Styling Feedback

> * Would it be possible to make the popover with multiple data points only appear when you hover over the actual line. Currently, the hoverable areas are so large, that the chart is almost always obscured by a popover.

Tried this many times. This is not possible with the container given.

> * Can we increase the width of the popover to avoid things wrapping so much?  It would be nice if it could grow as necessary with a max-width of 275px.  But if we have to just explicitly set the width, then let's go with 275px.
> * There's some weird indentation happening here
>   image
> * The color blocks on the Spending Revenue example are weird sizes
>   image

Grouping the last few together. This all has to do with the width. We can put it at 275, but even that is not big enough and it will cause multi-line tooltips.

> * Can we have a consistent 16px of space between the y-axis title and labels?  Currently it looks like that area is set to a specific value so the gap is inconsistent.
>   image
>   image

We do not choose the gap, That is built internally to the tool used basesd on their calculations.

orion-cengage:
@rcgillet Thanks for the responses.

The response to the first one is kind of a bummer.  I think I'll have to run this by some people and determine whether the behavior is acceptable.

Regarding the second one, I totally get that 275px won't result in the lines never wrapping, but it would be nice if it didn't happen quite so often.

rcgillet:
> Regarding the second one, I totally get that 275px won't result in the lines never wrapping, but it would be nice if it didn't happen quite so often.

The nice thing is, people can customize labels... so they can make them shorter if it gets really weird!

orion-cengage (approved):
I fully reviewed all of the requested styling fixes and the bug regarding the data point color in the popover.  All looks good to me at this point.

packages/charts/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.log
```
aaronmaturen, line 1: this file shouldn't be needed with `.gitignore` at the top level of the repo?

packages/charts/README.md
```diff
@@ -0,0 +1,40 @@
+# React Magma Schema Renderer
```
aaronmaturen, line 1: This isn't a schema renderer...

This is what gets shown in NPM

packages/charts/package-lock.json
```diff
@@ -1,64 +1,25 @@
 {
```
aaronmaturen, line 1: this file probably shouldn't exist... it gets added when you do `npm i something` inside the package... remove it and do an install at the top level to get the correct hoisting

packages/react-magma-dom/src/components/Tooltip/index.tsx
```diff
@@ -72,7 +75,7 @@ const ToolTipArrow = styled.span<{ position?: any; isInverse?: boolean }>`
   }
 `;
 
-const StyledTooltip = styled.div<{
+export const StyledTooltip = styled.div<{
```
aaronmaturen, line 78: we're doing this for `StyledButton` too where `@react-magma/charts` is using the internal version... should we be?

packages/react-magma-dom/src/theme/magma.ts
```diff
@@ -563,6 +563,25 @@ export const magma = {
     },
   },
 
+  charts: {
+    line: {
+      colors: [
```
aaronmaturen, line 568: I'm guessing these don't match up with `magma.colors`?

aaronmaturen: are the colors for `line` charts different than `bar` or `pie` or any other chart? 

Is it more important that these are colors, or for charts (or even for line charts directly)? 

If it's more important that they're a set of colors that work with the theme... but aren't the theme colors... maybe it makes more sense to just have an `iterableColors` array in the magma theme... then any component that just needs a set of colors that work with the theme can use it...

packages/charts/src/components/DataViz/DataVizTabs.tsx
```diff
@@ -0,0 +1,48 @@
...
+    return (
+      <div ref={ref} {...other}>
+        <Heading level={3}>Simple Line</Heading>
+        <Paragraph>
+          Data table description - Lorem ipsum dolor sit amet, consectetur
+          adipiscing elit. Donec ullamcorper quam dolor, at consequat mi pretium
+          in. Duis iaculis ligula nibh, sit amet.
+        </Paragraph>
+        <TabsContainer>
+          <Tabs aria-label="Line Chart Demo">
```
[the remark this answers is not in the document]
rcgillet: Deleting. Old files

orion-cengage:
Feedback from Gary at Perkins:

- We don't need to use dashes because we already have sufficient ways for the user to get information rather than the color.  This includes hovering over items in the legend, hovering over points in the chart, and the data table.
- If we want to, using different shapes other than the circles could be helpful, but probably not necessary.
- The way the screen reader takes control of the arrows is probably fine and made sense to him.
-- He did suggest we display keyboard instructions, but only when someone tabs from the Tab to the Chart.  Kind of like the Skip Link component.  So when you tab, you would see this:
image

- One thing we need to fix is the tooltips need to dismiss when you press the Escape key.  That's a newer WCAG rule that is pretty important.

PR #311, opened 2021-07-06
fix dropdown onClose when clicking outside menu

rcgillet:


aaronmaturen:
Add a changeset.

.changeset/popular-knives-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-magma": patch
```
aaronmaturen, line 2: react-magma-dom

aaronmaturen (changes requested):

.changeset/popular-knives-confess.md
rcgillet, line 2:
```suggestion
"react-magma-dom": patch
```

aaronmaturen:
My guess is that it's not working because the `handleMenuBlur` is just setting the `isOpen` state directly and not using the helper function that calls the `onClose` prop after `setIsOpen`

Not sure why we even want a global event listener when we can just handle it with the blurs...

aaronmaturen (approved):

PR #326, opened 2021-07-20
fix(modal): add useFocusLock hook. fix weird focus states

rcgillet:


packages/react-magma-dom/src/components/Modal/Modal.tsx
```diff
@@ -1,11 +1,11 @@
 import * as React from 'react';
 import styled from '@emotion/styled';
 import { Global, css } from '@emotion/core';
-import {
-  getTrapElements,
-  getTrapElementsAndFocus,
-  getFocusedElementIndex,
-} from './utils';
+// import {
```
aaronmaturen, line 4: delete the comments 😈

packages/react-magma-dom/src/hooks/useFocusLock.ts
```diff
@@ -0,0 +1,107 @@
+import * as React from 'react';
```
aaronmaturen, line 1: should we create a docs page for this? Makes sense that other teams (or components external to dom) would want this too

aaronmaturen: in the same vein... we should export it in `./src/index.ts`

packages/react-magma-dom/src/components/Modal/Modal.tsx
aaronmaturen: probably delete that code from utils also...

packages/react-magma-dom/package.json
```diff
@@ -50,6 +50,7 @@
   "devDependencies": {
     "@emotion/core": "^10.1.1",
     "@emotion/styled": "^10.0.27",
+    "@testing-library/user-event": "^13.2.0",
```
aaronmaturen, line 53: testing-library is at the root level... and if they're here, versions must match

aaronmaturen: this is why package-lock showed up

aaronmaturen (changes requested):

jamiesmyln (approved):
I didn't notice any strange behavior with components that use modals.

aaronmaturen (approved):

PR #329, opened 2021-07-27
fix(modal): stopped background from scrolling when a modal is open

jamiesmyln:
Whatever was behind the modal was scrollable so I changed how the CSS is applied.


orion-cengage (approved):
Looks good to me.  The background isn't scrolling anymore, and I also verified that the modal scrolls correctly if it's really long.

rcgillet (approved):

PR #347, opened 2021-08-17
Fix/disable input buttons

aaronmaturen:


orion-cengage:
Here are a couple of styling issues I see:

- The border of the disabled input should be neutral05
- The background of the disabled button within the input should not be white.  It should be transparent so you see the light gray background of the container behind the button.
image
- The Search button and Clear button are in the wrong order
image

aaronmaturen:
closes #348

PR #379, opened 2021-10-01
docs: page templates

orion-cengage:
Added placeholder page to test page templates


ccedrone:
see: https://github.com/cengage/react-magma/pull/422

PR #383, opened 2021-10-04
fix(tooltip): allow for hover in to tooltip message

rcgillet:


orion-cengage (approved):
Tooltip is working great now.

ccedrone (approved):

PR #384, opened 2021-10-05
chore: updating padding and icon sizes per Orion

ccedrone:


orion-cengage (approved):
Tags update looks good.

packages/react-magma-dom/src/components/Tag/Tag.tsx
```diff
@@ -215,16 +215,33 @@ function buildTagPadding(props) {
...
-      return `${props.theme.spaceScale.spacing02}`;
+      return `${props.theme.spaceScale.spacing02} 6px`;
+  }
+}
+
+function buildLabelPadding(props) {
+  if (props.icon) {
+    switch (props.size) {
+      case 'small':
+        return `0 4px`;
```
aaronmaturen, line 235: this should probably be spacingScale.spacing02

packages/react-magma-dom/src/components/Tag/Tag.tsx
```diff
@@ -215,16 +215,33 @@ function buildTagPadding(props) {
...
+      default:
+        return `0 8px`;
```
aaronmaturen, line 237: this should probably be spacingScale.spacing03

PR #411, opened 2021-11-12
feat: Predictive Search

ccedrone:
Adding in new prop within InputBase to allow for the Predictive Search feature.


orion-cengage (approved):
Looks good from UX design perspective.

aaronmaturen (approved):

Issue #355, opened 2021-08-23
Update help icon placement on inputs


orion-cengage:
Please update the placement of the Help icon on all form inputs.  All of the functionality that currently exists for the help icon should remain unchanged or adapt to the new placement.

![image](https://user-images.githubusercontent.com/24397282/136267131-fb280325-8f98-471d-a979-223a9b0cb6c2.png)

If the input label is long enough to wrap, the button should remain in the same position, relative to the container of the input.

![image](https://user-images.githubusercontent.com/24397282/136267204-8b820418-ea0a-4969-aa4c-dcacd564c3da.png)

### Specs

- Help icon should use the small, icon-only button with no visible container.
- There is a 4px gap underneath the button
- There is 16px of space between the text label and button to prevent them from running into each other.

![image](https://user-images.githubusercontent.com/24397282/136267837-ca44b719-aed7-4ee3-a6af-72c0b11d3d4f.png)

![image](https://user-images.githubusercontent.com/24397282/136267887-1bb0f415-404c-4997-b154-b1ec322ff8c4.png)

#### Inverse

![image](https://user-images.githubusercontent.com/24397282/136267931-7acffb48-061b-44db-a672-31c6c08bdc52.png)

ccedrone:
See: https://github.com/cengage/react-magma/pull/424

PR #424, opened 2021-12-14
feat: input updates to support additional icon positioning

ccedrone:
Added a enum item so that icons may go above the input field.

https://github.com/cengage/react-magma/issues/355


rcgillet (approved):

orion-cengage (changes requested):
Looks like you're using the medium button which is causing some overlap on hover:
image

Please use the small button as stated in the original ticket.  In the illustration below, the pink box shows the invisible boundary of the button, so you can see how it aligns on the right side of the input, and there's 4px of space below the button.
image

orion-cengage (approved):

PR #428, opened 2022-01-04
feat: Inputs update

ccedrone:
Removed bottom margin per ticket on Form Field Container.

### DIFF
--- a/.changeset/few-wasps-cry.md
+++ b/.changeset/few-wasps-cry.md
@@ -1,0 +1,7 @@
+---
+"react-magma-dom": patch
+---
+
+feat: Inputs update
+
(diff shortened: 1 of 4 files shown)


orion-cengage (changes requested):
I can see the bottom margin has been removed but the reserved area for helper text and errors is still there and taking up space even when those elements aren't being used.  Please see the original issue for full details.

image

orion-cengage (changes requested):
Is there a way the top 4px margin on the input message container can only show if there's actually help or error text present?  Otherwise, we still have 4px of artificial space below the input.

image

orion-cengage (approved):

PR #429, opened 2022-01-12
feat: updating datepicker keyboard instructions

ccedrone:
Modified the existing placement from a modal outside of the datepicker dropdown to a modal within it.


jamiesmyln (approved):
Looks good! (Needs a changeset, though.)

orion-cengage (changes requested):
Please make sure we're using the same Close button in both places.  It should be the one you see in the default view first.

image

image

orion-cengage (approved):

PR #607, opened 2022-03-18
Fix/drawer issues

chris-cedrone-cengage:
Fixing placement of drawer.

### DIFF
--- a/.changeset/fuzzy-crews-perform.md
+++ b/.changeset/fuzzy-crews-perform.md
@@ -1,0 +1,5 @@
+---
+"react-magma": patch
+---
+
+Fix/drawer issues


packages/react-magma-dom/src/components/TimePicker/TimePicker.tsx
```diff
@@ -140,7 +143,10 @@ export const TimePicker = React.forwardRef<HTMLInputElement, TimePickerProps>(
             type="number"
             value={hour}
           />
-          <Divider> : </Divider>
+          <Divider isInverse={isInverse} theme={theme}>
```
silvalaura, line 146: I think these changes snuck out in here, but are already a part of another PR. Should we revert?

greenervigil: I do not see this on main.  I think we should keep this in.

greenervigil (approved):

packages/react-magma-dom/src/components/TimePicker/TimePicker.tsx
silvalaura: The changes are in this open PR: https://github.com/cengage/react-magma/pull/605/files

greenervigil (changes requested):
@ccedrone The TimePicker is in another PR.  Should we remove it from this one?

packages/react-magma-dom/src/components/TimePicker/TimePicker.tsx
ccedrone: Oops, is there a fancy way to just remove the timepicker file?

greenervigil (approved):

Issue #318, opened 2021-07-13
RM 3 (4) - New home page


orion-cengage:
This ticket is only for the new version 3 of the React Magma docs.

New home page to go with the rest of the updates.  Designed to inform visitors more efficiently and accurately about what React Magma is and how to get started, and designed with future Dark Mode in mind.

**inVision:** https://cengage.invisionapp.com/share/Z3RH81FTNMC

### AC's

- Not much in terms of functionality.  The copy and design can be pulled from the inVision prototype.
- Designing and Developing sections have buttons leading to the pages to get started
-- Start Designing: links to "Getting Started" page under "Designing" section.
-- Start Developing: links to "Getting Started" page under "Developing" section.
- Responsive behavior (see prototype)
-- 768px wide and below, the Designing and Developing sections stack
-- 600px wide and below, the Feature cards stack on top of each other
-- Might have to make some other tweaks at other breakpoints based on what happens to the content.

### RM Logo
[rm-logo-large.png.zip](https://github.com/cengage/react-magma/files/6810893/rm-logo-large.png.zip)

### Screenshot

![image](https://user-images.githubusercontent.com/24397282/125480502-2c348960-8fef-4ccf-9027-073a93be8c69.png)

silvalaura:
This work was merged in https://github.com/cengage/react-magma/pull/352 and then reverted https://github.com/cengage/react-magma/commit/0f60e622b6578c20be1078645176a3153cd856a6. Not sure why, but we can certainly use this as a starting point.

PR #643, opened 2022-04-13
docs(home page): updating home page to have new designs

silvalaura:
#318

**This will stay a draft until we have the new colors and we can revisit these changes**

### DIFF
--- a/patterns/header/src/components/header/Header.tsx
+++ b/patterns/header/src/components/header/Header.tsx
@@ -72,8 +72,6 @@ export const Header = (props: HeaderProps) => {
 
   const isInverse = useIsInverse(props.isInverse);
 
-  console.log(theme.iterableColors);
-
   return (
     <AppBar
       isCompact={isCompact}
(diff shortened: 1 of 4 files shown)


greenervigil (approved):

orion-cengage (changes requested):
1. Looks like the old yellow button is being used here and not the new Tertiary yellow button.
image

2. On these two cards, can we please go with a background color of primary-500, and go borderless?
image

3. This button at the bottom (which is a hyperlink styled like a button) doesn't appear to have all of the correct styles.  Text is too heavy, and the there aren't any hover/active states.
image

4. Can we make the background of the footer white (neutral-100) to match the rest of the pages.
image

5. Please make sure the padding of these content areas is done consistently so they line up properly.
image

6. And then add an extra 24px of left padding to the div with the title and subtitle.
image

7. Can we add nbsp; between Design and System so you don't end up just getting the work System on a second line?
image

orion-cengage:
We've got another alignment issue further down the page.  Please make sure the various cards are the same distance from the left and right sides.  It's causing the gutter that runs through the middle of the page to be misaligned.

image

image

ccedrone (approved):

orion-cengage:
Styling suggestions:
1. Can we put a max-width of 1200px on the `article` of the page, and center on the page?
2. On the div containing the main title of the page, can we restrict its width to 66% ? This seems to help with the Magma logo shrinking to a weird size.
  2.1. But then at 600px and below, you have to remove that width so the title and subtitle correctly take up the entire width.

orion-cengage (approved):

Issue #305, opened 2021-06-30
RM 3 (1) - New header


orion-cengage:
**Purpose:** Part of RM Docs design to coincide with update to RM 3

**inVision:** https://cengage.invisionapp.com/share/Z3RH81FTNMC

- Update header to use the RM Header pattern
-- Use the compact prop
-- React Magma logo goes on the left
--- file: "rm-logo-full-light.svg"
--- Logo links to home page of that version of documentation.
-- Small/primary color Tag is displayed to the right of the logo.
--- Clicking this tag will take you to the new RM Versions page (to be handled in a separate ticket)
-- Button linking to GitHub on the right
-- Header pins to the top of the viewport
- Breakpoint 1024px and below:
-- Main navigation on left is suppressed
-- Button with menu icon is displayed to the left of the RM Logo
--- Clicking the button opens a drawer from the left
---- The drawer will contain the main navigation handled in a separate ticket.
- Breakpoint 600px and below:
-- RM Logo changes to "rm-logo-mark.svg"

**Screenshots**

Logo SVG files: [rm-logos.zip](https://github.com/cengage/react-magma/files/6743777/rm-logos.zip)

![image](https://user-images.githubusercontent.com/24397282/124025153-c0176a00-d9b5-11eb-820b-b9db495d6632.png)

![image](https://user-images.githubusercontent.com/24397282/124025208-cc032c00-d9b5-11eb-8cf5-ba2c49f0046e.png)

![image](https://user-images.githubusercontent.com/24397282/124025249-d4f3fd80-d9b5-11eb-8564-e7e9deb5f405.png)

![image](https://user-images.githubusercontent.com/24397282/124124447-0d3f1e80-da3e-11eb-98a6-1d057fe487b6.png)

orion-cengage:
Here's the Github icon.  Needs to be added to icon family.

[Github icon.zip](https://github.com/cengage/react-magma/files/7195889/Github.icon.zip)

PR #645, opened 2022-04-15
Feature/header v3

ccedrone:
re #305


silvalaura (approved):

orion-cengage (changes requested):
1. The github nav item on the right should be using a link/button component so the icon uses the primary color and you get hover/active states.
image

website/react-magma-docs/src/components/Masthead/index.js
```diff
@@ -1,18 +1,19 @@
 import React from 'react';
 import { Link } from 'gatsby';
-import { Logo } from '../Logo';
+import { Logo, SmallLogo } from '../Logo';
+import { ButtonVariant, IconButton } from 'react-magma-dom';
 import styled from '@emotion/styled';
 import { Header } from '@cengage-patterns/header';
-import { magma, useMediaQuery, AppBarPosition, Spacer } from 'react-magma-dom';
+import { magma, AppBarPosition } from 'react-magma-dom';
```
greenervigil, line 7: line 4 & 7 can go together.

silvalaura: addressed in [ff537d3](https://github.com/cengage/react-magma/pull/645/commits/ff537d347b7a929a91636d72b3fddc05696d520e)

orion-cengage (changes requested):
1. Because the header is now using a white background, the focus state on the logo/home button needs to be updated to the focus state for light backgrounds.
image

2. Same thing with the "Skip Navigation" button.  The style needs to be updated and needs to be positioned with the boundaries of the header.
image

orion-cengage (changes requested):
Looks like in the last update, the color of the text in the header's logo got changed to a lighter gray.  This should be neutral-700.

Also, please remove the change in color that happens on hover.

image

ccedrone:
I resoundingly approve of this impressive PR!

PR #646, opened 2022-04-15
docs(versions.mdx): new versions page in react-magma-docs

silvalaura:


ccedrone (approved):

orion-cengage (changes requested):
Can we please override the min-width on these instances of the tables to prevent them from scrolling horizontally?  I think all of the content within the tables should be able to fit on small screens.

orion-cengage (approved):

Issue #648, opened 2022-04-18
RM3 / Color system


orion-cengage:
Purpose: Create color system and naming convention that not only allows us to apply our brand, but also provide a scalable system  for use by other brands.

inVision: https://cengage.invisionapp.com/share/R312LU2EX4GN

### AC

1.0 Allow groupings of color, each based on a base color or starting point.  Each group can then consist of up to 9 shades and tints including that base color.  The base color takes the 500-position.
- 1.1 Define the following groups:
  - 1.1.1 Primary
  - 1.1.2 Secondary
  - 1.1.3 Tertiary
  - 1.1.4 Neutral
  - 1.1.5 Info
  - 1.1.6 Danger
  - 1.1.7 Warning
  - 1.1.8 Success

2.0 For our use of the groups, define the following colors within each group for the new Higher Ed branding
- 2.1 Primary
  - 2.1.1 primary100 - #E8E9F8
  - 2.1.2 primary200 - #BABDE9
  - 2.1.3 primary300 - #8B91DA
  - 2.1.4 primary400 - #5D65CB
  - 2.1.5 primary500 - #3942B0
  - 2.1.6 primary600 - #292F7C
  - 2.1.7 primary700 - #1A1E51
- 2.2 Secondary
  - 2.2.1 secondary500 - #FEE449
  - 2.2.2 secondary600 - #FEDA0D
  - 2.2.3 secondary700 - #ECC901
- 2.3 Tertiary
  - 2.3.1 tertiary500 - #CDDEFF
  - 2.3.2 tertiary600 - #A2C1FF
  - 2.3.3 tertiary700 - #79A5FF
- 2.4 Neutral
  - 2.4.1 neutral100 - #FFFFFF
  - 2.4.2 neutral200 - #F5F5F5
  - 2.4.3 neutral300 - #D4D4D4
  - 2.4.4 neutral400 - #8D8D8D
  - 2.4.5 neutral500 - #707070
  - 2.4.6 neutral600 - #5A5A5A
  - 2.4.7 neutral700 - #454545
  - 2.4.8 neutral800 - #2D2D2D
  - 2.4.9 neutral900 - #000000
- 2.5 Info
  - 2.5.1 info100 - #E8F5FC
  - 2.5.2 info200 - #A6DEFF
  - 2.5.3 info300 - #2FB3FF
  - 2.5.4 info400 - #009AF3
  - 2.5.5 info500 - #0074B7
  - 2.5.6 info600 - #005F96
[...]

PR #650, opened 2022-04-19
feat(magmacolors.ts): Adding new rebrand colors

silvalaura:
Adding the new rebrand colors per #648.

For now, I am leaving the legacy colors in the file until we transition all components to the new colors to avoid breaking anything. 

Colors can be accessed like: `theme.colors.primary` (which defaults to 500) or `theme.colors.primary600`

### DIFF
--- a/.changeset/rebrand-colors.md
+++ b/.changeset/rebrand-colors.md
@@ -1,0 +1,5 @@
+---
+"react-magma-dom": major
+---
+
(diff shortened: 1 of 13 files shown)


orion-cengage:
I'm not sure how I can best review this from my non-engineering perspective.  I'll leave it to the experts on this one.  If there are any issues, I'm sure they'll surface as we apply these new colors to the actual components.

PR #664, opened 2022-04-27
docs: update typography docs

orion-cengage:
Updated images and copy because of Work Sans, and made some tweaks to the docs styling.


website/react-magma-docs/src/components/MainNav/index.js
```diff
@@ -29,20 +29,19 @@ const activeStyleInverse = {
...
-    ${magma.spaceScale.spacing04};
+  padding: ${magma.spaceScale.spacing03} ${magma.spaceScale.spacing06};
 `;
 
 const Heading3 = styled.h3`
   color: ${props =>
-    props.isInverse ? magma.colors.neutral07 : magma.colors.neutral};
-  font-size: ${magma.typeScale.size04.fontSize};
-  font-weight: 700;
+    props.isInverse ? magma.colors.neutral06 : magma.colors.neutral700};
```
silvalaura, line 42: Can we replace `neutral06` with one of the new colors here?

silvalaura (commented):
In the future, please don't check in the package-lock (unless a new dependency has been added).

website/react-magma-docs/src/components/MainNav/index.js
orion-cengage: Yeah, you bet.  Sorry about the package-lock.  I didn't have that problem in the past, but I'll make sure I don't include it from now on.

silvalaura (approved):

Issue #590, opened 2022-02-17
RM3 / Branding refresh / Modal / Dark background


orion-cengage:
**Purpose:** Update the styling of the Modals for the new branding.

**inVision:** https://cengage.invisionapp.com/share/VN12E4BDPHBE

#### Prerequisites:
- Color palette in place
- Typography updated
- Card container updated
- Buttons updated
- Work Sans available

![image](https://user-images.githubusercontent.com/24397282/154516040-391b4886-45d5-46f9-a74a-eec44716be60.png)

## AC

- If still necessary, change font from Open Sans to Work Sans

### 1.0 Modal background overlay
- 1.1 Color = neutral-900, opacity 0.60

### 2.0 Modal container
- 2.1 Use updated raised card component for dark background

### 3.0 Modal title
- 3.1 The title will remain an H1 semantically but should use the styles of Heading/Small/Productive + white + semi-bold

![image](https://user-images.githubusercontent.com/24397282/154516170-badbdb5d-ad02-456b-a10e-07d8dcd962e8.png)

### 4.0 Close button
- 4.1 Use medium/text only/primary for dark background

![image](https://user-images.githubusercontent.com/24397282/154516241-056f917a-ffdd-4150-a0fe-a5bc665fa3d1.png)

Issue #589, opened 2022-02-17
RM3 / Branding refresh / Modal / Light background


orion-cengage:
**Purpose:** Update the styling of the Modals for the new branding.

**inVision:** https://cengage.invisionapp.com/share/U212E3OEADF4

#### Prerequisites:
- Color palette in place
- Typography updated
- Card container updated
- Buttons updated
- Work Sans available

![image](https://user-images.githubusercontent.com/24397282/154504000-f4a25e1c-fe7f-4fdb-b1dc-719b3ab4a109.png)

## AC

- If still necessary, change font from Open Sans to Work Sans

### 1.0 Modal background overlay
- 1.1 Color = neutral-900, opacity 0.60

### 2.0 Modal container
- 2.1 Use updated raised white card component

### 3.0 Modal title
- 3.1 The title will remain an H1 semantically but should use the styles of Heading/Small/Productive + semi-bold

![image](https://user-images.githubusercontent.com/24397282/154504651-39be1634-4d30-4ddf-89d2-89766aac9f76.png)

### 4.0 Close button
- 4.1 Use medium/text only/primary

![image](https://user-images.githubusercontent.com/24397282/154504595-8dbe3e37-2e9d-4dc8-82eb-7de4968551ee.png)

PR #681, opened 2022-05-09
feat(modal): updating Modal with rebrand styles

silvalaura:
Adds support for `isInverse`.

re #589 & #590

### DIFF
--- a/.changeset/rebrand-modal.md
+++ b/.changeset/rebrand-modal.md
@@ -1,0 +1,5 @@
+---
+"react-magma-dom": major
+---
+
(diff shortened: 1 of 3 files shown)


ccedrone (approved):

greenervigil (approved):

orion-cengage (changes requested):
Something is making all of the text within the modal #707070, which is too light.  The default color for all text should be neutral-700 (#454545) unless changed to something else like subdued text, alerts, inverse, etc.
image

orion-cengage (approved):

Issue #561, opened 2022-02-16
RM3 / Branding refresh / File Uploader / Dark background


orion-cengage:
**Purpose:** Update the styling of the File Uploader for the new branding.

**inVision:** https://cengage.invisionapp.com/share/N812DWWSAKTG

#### Prerequisites:
- Color palette in place
- Buttons already updated
- Card container updated
- Work Sans available

![image](https://user-images.githubusercontent.com/24397282/164549911-44733045-7ec7-4faf-b121-5a6b8f3fb5f3.png)

![image](https://user-images.githubusercontent.com/24397282/164549943-efda3e52-00ed-4e02-b23c-120c39e75543.png)

## AC

- If still necessary, change font from Open Sans to Work Sans

### 1.0 Buttons
- 1.1 Make sure the button in the non-drop variant and in the dropzone are using the new inverse primary/solid/medium button
![image](https://user-images.githubusercontent.com/24397282/154309021-176d0b9a-9a8b-4180-b5c8-478831190837.png)

- 1.2 Make sure the uploaded items are using the new primary/medium/text button with just the icon
![image](https://user-images.githubusercontent.com/24397282/154309185-283a2aee-c271-4a6e-a4c8-f21e544004ef.png)

### 2.0 Overall colors
- 2.1 All of the colors essentially need to be swapped out to use the new palette, even though some of them are the same hex value as the original.  Below are some major ones to call out:
- 2.2 Dropzone
-- 2.2.1 Background = neutral-900, opacity 0.25
-- 2.2.2 Border = neutral-100, opacity 0.50
- 2.2 Error messages, icons, and borders = danger-200
- 2.3 Icons on uploaded items = neutral-100

Issue #560, opened 2022-02-16
RM3 / Branding refresh / File Uploader / Light background


orion-cengage:
**Purpose:** Update the styling of the File Uploader for the new branding.

**inVision:** https://cengage.invisionapp.com/share/SP12DVR34FY9

#### Prerequisites:
- Color palette in place
- Buttons already updated
- Card container updated
- Work Sans available

![image](https://user-images.githubusercontent.com/24397282/164549201-bf6db380-15c1-4c66-9d96-1e2cf38f77dd.png)

![image](https://user-images.githubusercontent.com/24397282/164549794-e98a625f-1b05-4f55-b03d-91bcea3a7bff.png)

## AC

- If still necessary, change font from Open Sans to Work Sans

### 1.0 Buttons
- 1.1 Make sure the button in the non-drop variant and in the dropzone are using the new primary/solid/medium button
![image](https://user-images.githubusercontent.com/24397282/154306378-b8ef3243-c4cf-4a6f-af93-23eff56e101c.png)

- 1.2 Make sure the uploaded items are using the new primary/medium/text button with just the icon
![image](https://user-images.githubusercontent.com/24397282/154306444-2f7288fd-baa6-45ee-a4b1-0c10e9cb9fbe.png)

### 2.0 Overall colors
- 2.1 All of the colors essentially need to be swapped out to use the new palette, even though some of them are the same hex value as the original.  Below are some major ones to call out:
- 2.2 Dropzone
-- 2.2.1 Background = neutral-200
-- 2.2.2 Border = neutral-400
- 2.3 Error messages, icons, and borders = danger-500
- 2.4 Icons on uploaded items
-- 2.4.1 Generic, audio, image, video, zip = neutral-500
-- 2.4.2 PDF = danger-500
-- 2.4.3 Word = info-500
-- 2.4.4 Excel = success-500
-- 2.4.5 Powerpoint = warning-500

PR #683, opened 2022-05-09
feat(dropzone): updating dropzone colors

silvalaura:
re #560 & #561


greenervigil (approved):

orion-cengage (changes requested):
Please change the red border on the error-items from 2px to 1px.

image

When I upload an .svg file, it doesn't look like it's getting a thumbnail or the generic image associated with it.

image

image

ccedrone (approved):

PR #692, opened 2022-05-13
feat:(Dropdown) updates to v3 styling

ccedrone:


packages/react-magma-dom/src/components/Dropdown/DropdownMenuItem.tsx
```diff
@@ -71,13 +94,17 @@ const StyledItem = styled.div<{
...
 }>`
   ${MenuItemStyles}
 `;
 
-export const IconWrapper = styled.span`
-  color: ${props => props.theme.colors.neutral03};
+export const IconWrapper = styled.span<{ isInverse?: boolean }>`
+  color: ${props =>
+    props.isInverse
+      ? props.theme.colors.focusInverse
```
silvalaura, line 106: Can we use `info200` instead of `focusInverse` here, since this is not referring to the focus?

silvalaura: Actually, I think this should probably the white?

packages/react-magma-dom/src/components/Dropdown/Dropdown.tsx
```diff
@@ -57,7 +60,9 @@ export interface DropdownProps extends React.HTMLAttributes<HTMLDivElement> {
   width?: string | number;
 }
 
-const Container = styled.div`
+const Container = styled.div<{
+  isInverse?: boolean;
+}>`
```
silvalaura, line 65: Do we still need this addition?

packages/react-magma-dom/src/components/Dropdown/DropdownDivider.tsx
```diff
@@ -17,8 +25,17 @@ const StyledHr = styled.hr`
...
 
   const theme = React.useContext(ThemeContext);
-  return <StyledHr {...other} data-testid={testId} theme={theme} />;
+  const context = React.useContext(DropdownContext);
+
+  return (
+    <StyledHr
+      {...other}
+      data-testid={testId}
+      isInverse={context.isInverse}
```
silvalaura, line 37: Should `isInverse` come from context or props? If context, we don't need the change in line 28.

packages/react-magma-dom/src/components/Dropdown/DropdownMenuItem.tsx
```diff
@@ -36,12 +36,30 @@ export interface DropdownMenuItemProps
   value?: string | number;
 }
 
+export function menuColors(props) {
+  if (props.isInverse) {
+    return props.theme.colors.neutral100;
+  }
+  if (props.disabled) {
+    props.theme.colors.disabledText;
+  }
```
silvalaura, line 45: Disabled should be the first check, in case it's inverse AND disabled.

packages/react-magma-dom/src/components/Dropdown/DropdownMenuItem.tsx
```diff
@@ -36,12 +36,30 @@ export interface DropdownMenuItemProps
...
+  return props.theme.colors.neutral700;
+}
+
+export function menuBackground(props) {
+  if (props.isInverse) {
+    return props.theme.colors.primary600;
+  }
+  if (props.disabled) {
+    ('none');
```
silvalaura, line 54:
```suggestion
    return 'none';
```

silvalaura: Again, disabled should be the first check here

orion-cengage (changes requested):
Looks like there's some severe z-index issues with the dropdowns.
image

packages/react-magma-dom/src/components/Dropdown/DropdownDivider.tsx
ccedrone: Context, it was removed from 28 on the previous commit.

packages/react-magma-dom/src/components/Dropdown/DropdownMenuItem.tsx
ccedrone: Removed.

silvalaura: Looks like the return statement for the disabled case still needs to be updated

orion-cengage (approved):

orion-cengage (changes requested):
Looks like the disabled item isn't using the appropriate styling.

image

silvalaura (approved):

Issue #716, opened 2022-05-17
RM3 / Date picker / Dark background


orion-cengage:
**Purpose:** Update date picker with new brand styles for dark backgrounds.  Does not exist today.

![image](https://user-images.githubusercontent.com/24397282/168902351-309bc77d-b345-4658-97c2-12f840fe1cb9.png)

## AC

1.0 All buttons within the calendar component use inverse primary link variant.
2.0 Color for all the normal text and numbers is netural-100
3.0 Current day indicator (triangle) is secondary-500
4.0 Border around each day in the calendar is primary-200 
5.0 Container for the dropdown is same as dropdown component - primary-500 background with border.

silvalaura:
@orion-cengage can I get colors for this screen, please?
<img width="307" alt="image" src="https://user-images.githubusercontent.com/91160746/169140447-90c78aa9-f84e-47ce-9e0a-337e9e99402c.png">

orion-cengage:
@silvalaura Here you go!

![image](https://user-images.githubusercontent.com/24397282/169146437-2967d39d-e954-4d4f-ba5d-ef1b7cde6526.png)

Issue #715, opened 2022-05-17
RM3 / Date picker / Light background


orion-cengage:
**Purpose:** Update date picker with new brand styles

![image](https://user-images.githubusercontent.com/24397282/168901099-30d421f4-729d-4078-8fc2-31593c6c59ff.png)

## AC

1.0 All buttons within the calendar component use primary link variant.
2.0 Color for all the normal text and numbers is netural-700
3.0 Current day indicator (triangle) is primary-500
4.0 Border around each day in the calendar is netural-300

PR #719, opened 2022-05-19
feat(datepicker): update DatePicker with new colors

silvalaura:
re https://github.com/cengage/react-magma/issues/715 & https://github.com/cengage/react-magma/issues/716

### DIFF
--- a/.changeset/rebrand-datepicker.md
+++ b/.changeset/rebrand-datepicker.md
@@ -1,0 +1,5 @@
+---
+"react-magma-dom": major
+---
+
+feat(DatePicker): Update DatePicker with new colors. Add `isInverse` support.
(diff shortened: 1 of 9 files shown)


greenervigil (changes requested):
The keyboard shortcuts are not using the correct inverse text.  This should be addressed here as well.  

Captura de Pantalla 2022-05-19 a la(s) 22 59 19

orion-cengage (changes requested):
The border on the dropdown container looks too bright.  It should be using the same border styling as the dropdown component.

image

orion-cengage (changes requested):
Looks like this bug got introduced in 2.5.7.  The dropdown on the date picker is incorrectly overlapping the input instead of appearing 2px below it.

image

greenervigil (approved):

PR #730, opened 2022-05-24
docs(breakpointscontainer): update breakpoints container examples

### LINKED ISSUE #701: Bug/Breakpoints container - color on example
On some of the examples, we are showing a light blue background with white text.  This does not meet the contrast ratio we need for accessibility.

image

Solution: Everywhere we use this color as a background on the page, make the text primary-700 (#1A1E51).

image

silvalaura:
fix #701

### DIFF
--- a/.changeset/docs-breakpoints.md
+++ b/.changeset/docs-breakpoints.md
@@ -1,0 +1,5 @@
+---
+"react-magma-docs": major
+---
+
+docs: update breakpoints container examples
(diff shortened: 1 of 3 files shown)


orion-cengage (changes requested):
Looks like the text is changing to neutral-700 instead of **primary**-700.  Please do primary-700 because it's still not passing accessibility.

silvalaura:
> Looks like the text is changing to neutral-700 instead of **primary**-700. Please do primary-700 because it's still not passing accessibility.

I have replaced tertiary700 with warning500 as the background color, so it can pass accessibility with white text (using Card inverse) instead of customizing font color just for these examples.

orion-cengage (approved):

ccedrone (approved):

PR #741, opened 2022-06-01
fix: Toast X center alignment

ccedrone:
Fixing alignment for circle animation when text breaks to a second line.


silvalaura (commented):
Can we also add an example of a Toast with longer text to Storybook so we can catch this faster in the future?

silvalaura:
Looks great 👍  Can we just add a few more examples in Storybook? One with Toast text that takes up two lines, and maybe some edge cases where the toast text is even longer to ensure it still aligns correctly?

silvalaura (approved):

PR #779, opened 2022-06-13
fix(tooltip): fix tooltip position bug

### LINKED ISSUE #752: Bug: Tooltip hiding behind other objects
Description: The Tooltip Right example is underneath the other buttons to show the other positions.  It's my understanding that Tooltips should never really be able to be hidden behind anything.  They should always appear above any other object on a page.  We need to determine why this is happening.

image

silvalaura:
fix #752

Also, I noticed a console warning that the following icons are deprecated, so I figured we should update our docs to use the non-deprecated versions.
```
  CircleRightIcon,
  CircleLeftIcon,
  CircleDownIcon,
  CircleUpIcon
```

### DIFF
--- a/.changeset/tooltip-position.md
+++ b/.changeset/tooltip-position.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': major
+---
+
+feat(tooltip): Fix tooltip position bug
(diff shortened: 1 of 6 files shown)


ccedrone (approved):

orion-cengage (approved):
Great work on the tooltips, and it was really helpful to see all the different scenarios to test out in Storybook.

Issue #773, opened 2022-06-10
Docs: Pagination


ccedrone:
Handle Page Change errors out

![blarg](https://user-images.githubusercontent.com/77400920/173105540-5469a8c7-1445-48e0-bcc5-784754d68c0c.png)

Issue #771, opened 2022-06-10
Docs: Table - min-width example


ccedrone:
Min-Width overflowing from example (overflow-y on div needed

![example](https://user-images.githubusercontent.com/77400920/173104210-657c5b32-5f90-4445-ba34-089644646f65.png)
)

Issue #765, opened 2022-06-10
Docs: Tabs > Inverse


silvalaura:
**Describe the bug**
Under the Tabs docs, there are isInverse examples not using the inverse correctly

**To Reproduce**
1. Go to https://next--upbeat-sinoussi-f675aa.netlify.app/api/tabs/#inverse
2. Under Applied to Tabs Container, click About Us
3. Notice the content is set to `isInverse={false}`

**Expected behavior**
All Inverse examples should use the Inverse

![image](https://user-images.githubusercontent.com/91160746/173100361-fddb0be9-83d9-4bb8-adf0-5e7baa1f5eff.png)

On another note, I wonder if we can remove the separate sections `Applied to Tab Panel`, `Applied to Tab Panels Container`, `Applied to Tabs Container` considering that people will use isInverse for all parts of tabs and not individually.

Issue #764, opened 2022-06-10
Docs: Tags > Internationalization


silvalaura:
**Describe the bug**
Why is there a section for Internationalization under Tags?

**To Reproduce**
1. Go to https://next--upbeat-sinoussi-f675aa.netlify.app/api/tag/#internationalization

**Expected behavior**
This section shouldn't exist, but if it does need to be there, we should update the code to use the `<Tag>` component, instead of `<tag>`

PR #782, opened 2022-06-14
fix: fixing a variety of documentation examples

### LINKED ISSUE #772: Docs: Spinner - Inverse example needs padding on the left
padding

silvalaura:
fix #772 & #765 & #773 & #771 & #764


orion-cengage (changes requested):
It looks like most or all of the table examples are now breaking when the browser window gets smaller.  Here is just one example showing the contents spilling out of the container.  This is also causing the entire page to scroll horizontally.

image

orion-cengage (approved):

greenervigil (approved):

Issue #754, opened 2022-06-10
Bug: RM2 - All alerts, close button misalignment


orion-cengage:
Issue found in 2.5.13

Description: The close buttons are misaligned.  Most likely a result of a previous update.

![image](https://user-images.githubusercontent.com/24397282/173091259-93f172de-16c7-4bc1-b1db-9b0bb644f8f9.png)

![image](https://user-images.githubusercontent.com/24397282/173091406-8e30ca1a-650c-4e1e-a9ff-294ce31741e5.png)

PR #786, opened 2022-06-16
fix: Close button alignment on Alerts

ccedrone:
Regarding: [754](https://github.com/cengage/react-magma/issues/754)

### DIFF
--- a/.changeset/patch-alerts-fix.md
+++ b/.changeset/patch-alerts-fix.md
@@ -2,4 +2,4 @@
 "react-magma-dom": patch
 ---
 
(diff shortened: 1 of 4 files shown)


silvalaura (approved):
Approved, but let's hold off on merging. If we make progress on the VoiceOver bug, then we could get both out together.

orion-cengage (approved):

silvalaura:
@ccedrone Now that this fix has been released as a patch, we need to copy these changes to the dev version with a new PR against dev.

Issue #180, opened 2021-03-02
Tree View Component


orion-cengage:
**Prototype:** https://cengage.invisionapp.com/share/FWZS3HZAZYR
**Inspect:** https://cengage.invisionapp.com/d/main#/console/15897083/440993701/preview

Inspiration: https://material-ui.com/components/tree-view/

Requirements

- Component can support unlimited number of parents and children
- Each item must have a text label
-- Text label can wrap if there isn't enough room
- Each item can display an optional icon to the left of the label
- Parent items can be expanded and collapsed
-- You can expand or collapse the parent by clicking/tapping anywhere on the item
-- Option to have any parent expanded by default
-- Option to show every level expanded with no collapsing
-- Each item will get the appropriate amount of padding/margin on the left to ensure each level is properly aligned
--- Some scenarios might get more complex when certain items don't have an arrow or icon like their peers

silvalaura:
Accessibility reference: https://www.w3.org/WAI/ARIA/apg/patterns/treeview/

Laura's notes:
<img width="304" alt="image" src="https://user-images.githubusercontent.com/91160746/225409543-34fe6db6-5bcd-4234-88c0-353215503d42.png">
<img width="675" alt="image" src="https://user-images.githubusercontent.com/91160746/225410318-1d0952ff-bbd9-4b4d-96e9-5b192d967a4c.png">
<img width="716" alt="image" src="https://user-images.githubusercontent.com/91160746/225412124-9bcd41f2-0df0-408e-a173-16a65d0db89a.png">
<img width="247" alt="image" src="https://user-images.githubusercontent.com/91160746/225412547-a380e7c7-2c65-48c0-b20e-087e906b4c54.png">
<img width="173" alt="image" src="https://user-images.githubusercontent.com/91160746/225412641-8713df73-046a-40bb-9cbe-47ea28a9ab86.png">
<img width="526" alt="image" src="https://user-images.githubusercontent.com/91160746/225415131-e1bb6602-9d28-4beb-a1a0-ba97050079cc.png">

silvalaura:
More Laura notes:
- hasIcons: boolean. will have default icons (noted above) for branches and leaves, but can be customized. this should prevent some branches having icons and leaves not having them
- isSelectable: single, multi (checkboxes), off (default)

- Make an example for a zebra view with items on the right, and examples of how it could be restyled
- Examples to build: filters/batch edit (selectable) and document selector
- Make sure all the examples are complex (more than one level)
[...]

PR #787, opened 2022-06-21
feat(TreeView): TreeView and TreeItem components

acoopa:
Issue: #180 

## What I did

Created a TreeView component.

## Screenshots:

image
image
image

## Checklist 

## How to test

Requirements and testing notes found in Notion: https://laura-silva.notion.site/TreeView-1fa9fc9d40654f7dba969a7fbaa492c6


packages/react-magma-dom/src/components/TreeView/useTreeView.ts
```diff
@@ -0,0 +1,43 @@
+import * as React from "react"
+
+export enum ExpandInitialOptions {
+  all = "all",
+  first = "first",
+  none = "none",
+}
+export interface UseTreeViewProps {
+  expandInitial?: ExpandInitialOptions;
```
silvalaura, line 9: Can we add a description for these props, and outline the default value?

packages/react-magma-dom/src/components/TreeView/index.ts
```diff
@@ -0,0 +1,3 @@
+// export * from "./useTreeView"
```
silvalaura, line 1: Do we need to uncomment this line?

silvalaura (commented):
Looking great so far, nice work! 👏 

A couple other notes:
- I noticed that the expand buttons don't look clickable (the cursor doesn't change to the pointer). Also, the accessibility check on storybook calls out `Buttons must have discernible text` so I think they are just missing the aria label attribute
- We try not to commit the package-lock unless something has changed in our packages

ccedrone (approved):

silvalaura:
Although there are some outstanding TODOs, there will be a followup PR to clean that up.

PR #821, opened 2022-07-06
Fix/table border conflict

ccedrone:
Moving the table border to a wrapping container which fixes overflow issues.

### DIFF
--- a/packages/react-magma-dom/src/components/Table/Table.tsx
+++ b/packages/react-magma-dom/src/components/Table/Table.tsx
@@ -118,7 +118,7 @@ export const TableContainer = styled.div<{
             : props.theme.colors.neutral300
         }`
       : 0};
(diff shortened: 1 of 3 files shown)


packages/react-magma-dom/src/components/Table/Table.tsx
```diff
@@ -121,8 +118,14 @@ const StyledTable = styled.table<{
             : props.theme.colors.neutral300
         }`
       : 0};
-  border-radius: ${props =>
-    props.hasSquareCorners ? 0 : props.theme.borderRadius};
+  overflow-x: ${props => (props.minWidth ? 'auto' : 'visible')};
```
silvalaura, line 121: This used to be `overflow`. Was it intentionally changed to `overflow-x`?

silvalaura (approved):

PR #824, opened 2022-07-08
Fix/data grid edges

ccedrone:
This is a proposed change to Table which requires an added container by the user in order to include border styling on the outer Table.

**<TableContainer><Table>Table Thangs</Table></TableContainer>**

The previous implementation excluded the pagination aspect of the Table and requires a new approach.

If reverted back to Card, the CSS property of overflow:hidden is needed to hide the table edges. This isn't ideal seeing how widely used the component is. Hence this may be a solution, however it's potentially not ideal.

See "With Table Container", "Pagination Inverse", and "Row Colors Inverse" in Storybook for the examples.

Doc site examples haven't been added due to the tenuous nature of this proposal.

-Chris


ccedrone:
Cards will be recommended for border styling in the meantime.

PR #827, opened 2022-07-14
800 fix focus state

### LINKED ISSUE #800: Fix focus state management on dropdown + modal
Perkins CENG_2486, CENG_2491

These are similar and probably related issues between the dropdown and modal components.

**Dropdown**
When making a selection from the dropdown (or pressing the Escape key to close it), focus returns to the dropdown button but there is not a visible focus.

**Modal**
When closing the modal visible focus is not evident on the button that triggers the modal.

greenervigil:
fixes #800

### DIFF
--- a/packages/react-magma-dom/src/components/Dropdown/Dropdown.stories.tsx
+++ b/packages/react-magma-dom/src/components/Dropdown/Dropdown.stories.tsx
@@ -95,7 +95,6 @@ Default.args = {};
 export const OnClickOutside = Template.bind({});
 OnClickOutside.args = {
   ...Default.args,
-  onClose: event => alert('I am closing'),
 };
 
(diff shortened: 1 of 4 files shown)


.changeset/healthy-ties-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-magma-docs": patch
+---
+
+800 fix focus state
```
silvalaura, line 5: This changes the Dropdown `closeDropdown` functionality, so I'd prefer this get tagged under `"react-magma-dom": major`. The description of the change should say that and maybe mention that modal examples were updated as well.

silvalaura (commented):
Can we remove the package-lock change as it's not needed for this specific PR?

silvalaura (approved):

orion-cengage (approved):

Issue #608, opened 2022-03-23
Tabs: a11y bug / aria-label


orion-cengage:
The following bug was found in 2.5.8 and reported by Shawn Haley.

![image](https://user-images.githubusercontent.com/24397282/159716785-90503ba8-bc14-432d-b641-c361b56f4f84.png)

Uditsingh67:
The issue is fixed for Tabs components.
<html>
<body>


OS | Version | Result
-- | -- | --
Win 10 (Chrome) | 102.0.5005.63 | PASS (http://somup.com/c3jbivu9Y8)
Win 10 (Firefox) | 101.0 | PASS (http://somup.com/c3jbiau9r3)
Win 10 (Edge) | 103.0.1264.37 | PASS (http://somup.com/c3jbi4u9rI)
MAC (safari) | 15.3 | PASS (http://somup.com/c3jbipu9rH)
Ipad (safari) | v15.1 | PASS (http://somup.com/c3jbiGu93f)


</body>
</html>

Uditsingh67:
Issue does not exist.
See Screen-cast : http://somup.com/c3jwfIU2oC

PR #846, opened 2022-07-25
Fix/tabs a11y

ccedrone:
Regarding: [608](https://github.com/cengage/react-magma/issues/608)

### DIFF
--- a/.changeset/bug-tabs.md
+++ b/.changeset/bug-tabs.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': minor
+---
+
+fix(Tabs): Removing unnecessary aria-label from wrapping div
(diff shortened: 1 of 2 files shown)


.changeset/bug-tabs.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': minor
```
silvalaura, line 2: This is a bug fix, so it should go under `patch` instead of minor. 

For reference,
```
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backwards compatible manner
PATCH version when you make backwards compatible bug fixes
```

silvalaura (approved):

greenervigil (approved):

Issue #506, opened 2022-02-14
a11y_NativeSelect> Helper Message Inverse> A11y> 1 violations is appearing


priya-rajput:
**Describe the bug**
NativeSelect>> "1 violations" is appearing for "Helper Message Inverse". 

**To Reproduce**
1. Go to 'https://storybook-preview-290--upbeat-sinoussi-f675aa.netlify.app/?path=/story/nativeselect--helper-message-inverse'
2. Click on 'Accessibility" add on
3. "1 violations" is appearing

**Expected behavior**
"0 violations" should appear for NativeSelect> Helper Message Inverse.

**Screenshots**
https://somup.com/c3n2QfZW9F

**Desktop:**
 - OS: [Win 10]
 - Browser [chrome]
 - Version [Version 98.0.4758.82 (Official Build) (64-bit)]

Uditsingh67:
The issue is fixed.
See Screen-cast : http://somup.com/c3jffjuCpQ
Thanks!

ccedrone:
This was fixed via: [854](https://github.com/cengage/react-magma/pull/854)

Uditsingh67:
Issue does not exist.
See Screen-cast : http://somup.com/c3jvi9UldN

Issue #830, opened 2022-07-19
Bug_NativeSelect > Default /Has Error /Helper Message /Inverse /Has Error Inverse /Helper Message Inverse > User is not able to launch the dropdown menu by clicking on dropdown icon.


priya-rajput:
**Describe the bug:**
Dropdown menu is not launching when user clicks the dropdown icon.

**To Reproduce:**
1. Go to 'https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/iframe.html?id=nativeselect--inverse&args=&viewMode=story'
2. Click on dropdown icon.
3. Verify that dropdown menu is not launching.

**Expected behavior:**
Dropdown menu should launch when user clicks the dropdown icon.

**Screenrecording:**
https://somup.com/c3iqI6u3kH

**Desktop:**
OS: [Win 10],Browser [chrome][Version 98.0.4758.82 (Official Build) (64-bit)]
OS: [Mac 12], Browser [Safari][Version 15.2]
OS: [iOS Device],Browser [Safari][Version 15.5]

Uditsingh67:
The Issue is fixed.
See Screen-cast : http://somup.com/c3jffMuC5f
Thanks!

Uditsingh67:
Issue does not exist.
See Screen-cast : http://somup.com/c3jvjVUlKl

Issue #330, opened 2021-07-29
Nested interactive elements for NativeSelect


aaronmaturen:
```
FAIL  packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
  ● NativeSelect › Does not violate accessibility standards

    expect(received).toHaveNoViolations(expected)

    Expected the HTML found at $('.css-1gckytq-StyledButton-inputBaseStyles-inputWrapperStyles') to have no violations:

    <div data-testid="selectTriggerButton" role="button" style="cursor: default;" tabindex="0" class="css-1gckytq-StyledButton-inputBaseStyles-inputWrapperStyles e1h9drla1">

    Received:

    "Ensure interactive controls are not nested (nested-interactive)"

    Fix any of the following:
      Element has focusable descendants

    You can find more information on this issue here:
    https://dequeuniversity.com/rules/axe/4.2/nested-interactive?application=axeAPI

      32 |
      33 |     return axe(container.innerHTML).then(result => {
    > 34 |       return expect(result).toHaveNoViolations();
         |                             ^
      35 |     });
      36 |   });
      37 | });

      at packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js:34:29
```

axe doesn't like that the `select` is nested inside the `toggleButton` ... I'm assuming they should be siblings not parent/child

also, if the plan is to move the downshift components out of dom, we probably don't want to use the `toggleButton` from select anyways 🤔

silvalaura:
Review after #830

silvalaura:
No longer an issue ✅ 

![Image](https://github.com/cengage/react-magma/assets/91160746/1b482699-720a-47f1-bfdd-5dbbf6e0a99e)

PR #854, opened 2022-07-28
fix: Native Select - dropdown menu

ccedrone:
Reworking component by removing SelectTriggerButton which should fix: [830](https://github.com/cengage/react-magma/issues/830) & [330](https://github.com/cengage/react-magma/issues/330).

### DIFF
--- a/.changeset/bug-native-select.md
+++ b/.changeset/bug-native-select.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+Removing SelectTriggerButton to fix dropdown caret issues.
--- a/.changeset/bug-native-select.md
+++ b/.changeset/bug-native-select.md
@@ -2,4 +2,4 @@
 'react-magma-dom': patch
 ---
 
-Removing SelectTriggerButton to fix dropdown caret issues.
+fix(Native Select): Fixed dropdown caret which wasn't selecting the options on click.
--- a/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
+++ b/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
@@ -53,29 +53,37 @@ describe('NativeSelect', () => {
 
   it('should render an error state', () => {
     const testId = 'test-id';
-    const { getByTestId } = render(
+    const { getByTestId, getByText } = render(
       <NativeSelect
         errorMessage="This is an error"
         testId={testId}
       ></NativeSelect>
     );
+
     const nativeselect = getByTestId(testId).parentElement;
     expect(nativeselect).toHaveStyleRule('border-color', magma.colors.danger);
+
+    const errorMessage = 'This is an error';
+    expect(getByText(errorMessage)).toBeInTheDocument();
   });
 
   it('should render an inverse error state', () => {
     const testId = 'test-id';
-    const { getByTestId } = render(
+    const { getByTestId, getByText } = render(
       <NativeSelect
         errorMessage="This is an error"
         isInverse
         testId={testId}
       ></NativeSelect>
     );
+
     const nativeselect = getByTestId(testId).parentElement;
     expect(nativeselect).toHaveStyleRule(
       'border-color',
       magma.colors.danger200
     );
+
+    const errorMessage = 'This is an error';
+    expect(getByText(errorMessage)).toBeInTheDocument();
   });
 });
--- a/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
+++ b/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
@@ -60,4 +60,32 @@ describe('NativeSelect', () => {
       transparentize(0.6, magma.colors.neutral100)
     );
   });
+
+  it('should render an error state', () => {
+    const testId = 'test-id';
+    const { getByTestId } = render(
+      <NativeSelect
+        errorMessage="This is an error"
+        testId={testId}
+      ></NativeSelect>
+    );
+    const nativeselect = getByTestId(testId).parentElement;
+    expect(nativeselect).toHaveStyleRule('border-color', magma.colors.danger);
+  });
+
+  it('should render an inverse error state', () => {
+    const testId = 'test-id';
+    const { getByTestId } = render(
+      <NativeSelect
+        errorMessage="This is an error"
+        isInverse
+        testId={testId}
+      ></NativeSelect>
+    );
+    const nativeselect = getByTestId(testId).parentElement;
+    expect(nativeselect).toHaveStyleRule(
+      'border-color',
+      magma.colors.danger200
+    );
+  });
 });
--- a/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
+++ b/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
@@ -51,6 +51,30 @@ describe('NativeSelect', () => {
     );
   });
 
+  it('should render a default border', () => {
+    const testId = 'test-id';
+    const { getByTestId } = render(
+      <NativeSelect testId={testId}></NativeSelect>
+    );
+
+    expect(getByTestId(testId)).toHaveStyleRule(
+      'border',
+      `1px solid ${magma.colors.neutral500}`
+    );
+  });
+
+  it('should render a default inverse border', () => {
+    const testId = 'test-id';
+    const { getByTestId } = render(
+      <NativeSelect isInverse testId={testId}></NativeSelect>
+    );
+
+    expect(getByTestId(testId)).toHaveStyleRule(
+      'border',
+      `1px solid ${transparentize(0.5, magma.colors.neutral100)}`
+    );
+  });
+
   it('should render an error state', () => {
     const testId = 'test-id';
     const errorMessage = 'This is an error';
--- a/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
+++ b/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
@@ -15,16 +15,6 @@ describe('NativeSelect', () => {
     expect(getByTestId(testId)).toBeInTheDocument();
   });
 
-  it('should show an error message', () => {
-    const labelText = 'Label';
-    const errorMessage = 'This is an error';
-    const { getByText } = render(
-      <NativeSelect labelText={labelText} errorMessage={errorMessage} />
-    );
-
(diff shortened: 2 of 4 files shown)


.changeset/bug-native-select.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+Removing SelectTriggerButton to fix dropdown caret issues.
```
silvalaura, line 5: Think of this changelog as what our adopters see when updating versions. They likely don't care about the implementation (they don't know what SelectTriggerButton is), but they care **what** bug is fixed. So I'd recommend writing these messages in terms of what was fixed instead of how. In addition, it's helpful to note what component is affected.

Consider something along the lines of: `fix(component): Description of bug fix`

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
```diff
@@ -34,4 +36,28 @@ describe('NativeSelect', () => {
...
+    const { getByTestId } = render(
+      <NativeSelect disabled isInverse testId={testId}></NativeSelect>
+    );
+    const nativeselect = getByTestId(testId);
+    expect(nativeselect).toHaveStyleRule(
+      'color',
+      transparentize(0.6, magma.colors.neutral100)
+    );
+  });
 });
```
silvalaura, line 63: Can we add a test (or add to the error test) the red input outline?

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
```diff
@@ -34,4 +36,56 @@ describe('NativeSelect', () => {
...
+    const testId = 'test-id';
+    const { getByTestId } = render(
+      <NativeSelect
+        errorMessage="This is an error"
+        testId={testId}
+      ></NativeSelect>
+    );
+    const nativeselect = getByTestId(testId).parentElement;
+    expect(nativeselect).toHaveStyleRule('border-color', magma.colors.danger);
+  });
```
silvalaura, line 74: This is a almost a duplicate of lines 18-23. They can be condensed to 1

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -22,19 +23,38 @@ export interface NativeSelectProps
...
 }
-const StyledNativeSelect = styled.select<{
+const StyledNativeSelectWrapper = styled.div<{
+  disabled?: boolean;
+  hasError?: boolean;
+  isInverse?: boolean;
   theme: ThemeInterface;
+}>`
+  align-items: center;
+  display: flex;
```
silvalaura, line 36: These two styles are already in `inputWrapperStyles`. 

I would suggest having `inputWrapperStyles` first, and then below any new style that is missing or needs to be overwritten.

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
silvalaura: This is missing `expect(getByText(errorMessage)).toBeInTheDocument();` (and the inverse test below as well)

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -22,19 +23,36 @@ export interface NativeSelectProps
...
   isInverse?: boolean;
+  theme: ThemeInterface;
 }>`
+  margin: -1px 0;
   ${inputBaseStyles};
-  background: inherit;
-  > option {
-    background: ${props =>
-      props.isInverse ? props.theme.colors.neutral600 : 'none'};
-  }
```
silvalaura, line 37: We need to keep this. This was added as part of [fix(nativeselect): add background color to the options for isInverse](https://github.com/cengage/react-magma/commit/ab69379bb8dda7094eba22034e738387e3228a40) 

image

silvalaura: We should maybe even add a comment so we don't accidentally remove it
```
// Required for Windows && Chrome support
```

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -22,14 +23,37 @@ export interface NativeSelectProps
...
+    z-index: 1;
+  }
+`;
+
+const StyledNativeSelect = styled.select<{
   isInverse?: boolean;
+  theme: ThemeInterface;
 }>`
   ${inputBaseStyles};
+  height: 38px;
```
silvalaura, line 55: Do we need to hard code the height? If so, where did this number come from? Usually our numbers are multiples of 4, so just wondering if this needs to be 40px instead. I'll defer to @orion-cengage

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -22,14 +23,37 @@ export interface NativeSelectProps
...
+  // Required for Windows && Chrome support
   background: inherit;
```
silvalaura, line 57: Is this still affecting the border styling?

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
```diff
@@ -34,4 +26,64 @@ describe('NativeSelect', () => {
...
+    );
+
+    const nativeselect = getByTestId(testId).parentElement;
+    expect(nativeselect).toHaveStyleRule(
+      'border-color',
+      magma.colors.danger200
+    );
+
+    const errorMessage = 'This is an error';
+    expect(getByText(errorMessage)).toBeInTheDocument();
```
silvalaura, line 87: **This is style feedback and not required changes. Take it or leave it**

1. I would suggest declaring the `errorMessage` variable at the top of the test so you can use it in multiple places instead of retyping `'This is an error'` twice. This way, if we wanted to change what that message says we would only need to do it once.

2. Typically if I'm only going to use a value once, I don't create a variable (unless it would really help for readability). I would probably just have this expect in a single line `expect(getByTestId(testId).parentElement).toHaveStyleRule(.......);`

```
    const testId = 'test-id';
    const errorMessage = 'This is an error';

    const { getByTestId, getByText } = render(
      <NativeSelect
        errorMessage={errorMessage}
        isInverse
        testId={testId}
      ></NativeSelect>
    );

    expect(getByTestId(testId).parentElement).toHaveStyleRule(
      'border-color',
      magma.colors.danger200
    );

    expect(getByText(errorMessage)).toBeInTheDocument();
```

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
ccedrone: Removed fixed height.

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
ccedrone: Moved border to <select> to help with spacing.

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -22,14 +23,41 @@ export interface NativeSelectProps
...
+const StyledNativeSelect = styled.select<{
   isInverse?: boolean;
+  theme: ThemeInterface;
 }>`
   ${inputBaseStyles};
+  border: 1px solid
+    ${props =>
+      props.isInverse
+        ? transparentize(0.5, props.theme.colors.neutral100)
+        : props.theme.colors.neutral500};
```
silvalaura, line 59: The border should be red when there's an error, which makes me wonder how the unit tests passed 🤔

ccedrone:
This may also fix: https://github.com/cengage/react-magma/issues/506

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -69,23 +111,26 @@ export const NativeSelect = React.forwardRef<HTMLDivElement, NativeSelectProps>(
...
+        <StyledNativeSelectWrapper
           disabled={disabled}
           hasError={!!errorMessage}
           isInverse={isInverse}
-          toggleButtonProps={''}
+          theme={theme}
         >
           <StyledNativeSelect
+            aria-label={nativeSelectLabel}
             {...other}
```
silvalaura, line 122: This works, but it may be nice to be a little more explicit about the aria label.

Under props, we should declare the aria-label (`'aria-label': ariaLabel,`) and then we can have the aria-label conditionally have either the one passed by the user if it exists, or fall back on the default:
```
{...other}
aria-label={ariaLabel || nativeSelectLabel}
```

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -54,9 +93,12 @@ export const NativeSelect = React.forwardRef<HTMLDivElement, NativeSelectProps>(
       ...other
     } = props;
     const theme = React.useContext(ThemeContext);
+    const i18n = React.useContext(I18nContext);
     const isInverse = useIsInverse(isInverseProp);
     const id = useGenerateId(defaultId);
 
+    const nativeSelectLabel = i18n.select.placeholder;
```
silvalaura, line 100: Accessibility wise, should the aria-label be the placeholder (in this case `Select...`) or the `labelText`?

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
ccedrone: Removed from component as deemed unnecessary by Leslie.

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
ccedrone: Removed from component as deemed unnecessary by Leslie.

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -53,8 +91,13 @@ export const NativeSelect = React.forwardRef<HTMLDivElement, NativeSelectProps>(
       testId,
       ...other
     } = props;
+
+    console.log(other);
+
```
silvalaura, line 96: 🤭

silvalaura (approved):
🌟

PR #856, opened 2022-07-28
feat(datagrid): Add `isSortableBySelected` prop to DataGrid

silvalaura:
This fix was released as part of 2.6.0 (https://github.com/cengage/react-magma/pull/841) so just copying those changes to v3.


orion-cengage (changes requested):
1. I think we should make the other two columns in the example sortable as well so there's an example of how different sortable columns work together.
2. The focus style on the arrow needs to be solid instead of dashed
image
image

3. When the column is not currently sorted and has the double arrow, it's supposed to be a lighter gray (neutral-500) than when it is sorted (neutral-700).  _See regular table component for example._
image
image

image

orion-cengage (approved):

ccedrone (approved):

greenervigil (approved):

PR #864, opened 2022-08-08
fix: updating Storybook example to pass a11y checks

ccedrone:


silvalaura:
In storybook, I'm still seeing `1 Violation` for all examples if I refresh the page:
image
image

It seems to be related to the color contrast for the alert (we get the same violation under the [Alerts stories](https://storybook-preview-864--upbeat-sinoussi-f675aa.netlify.app/?path=/story/alert--default)). @orion-cengage thoughts?

ccedrone:
Strange, I can't reproduce for the Inverse example, however the others show an error, which is reading a different color for the foreground on each page load.

silvalaura:
> Strange, I can't reproduce for the Inverse example, however the others show an error, which is reading a different color for the foreground on each page load.

To reproduce the Inverse violation, go directly to https://storybook-preview-864--upbeat-sinoussi-f675aa.netlify.app/?path=/story/container--inverse (instead of going to Default first and then Inverse)

ccedrone:
Still doesn't reproduce the error for me.

ccedrone:
PR unnecessary due to Storybook issues.

PR #868, opened 2022-08-09
fix: Input adding aria-label to clickable examples for a11y pass in Storybook

ccedrone:


packages/react-magma-dom/src/components/Input/Input.stories.tsx
```diff
@@ -109,12 +109,14 @@ export const ClickableIcon = Template.bind({});
 ClickableIcon.args = {
   ...Default.args,
   icon: <NotificationsIcon />,
+  iconAriaLabel: 'Bell Icon',
```
silvalaura, line 112: Shouldn't the aria labels explain to the user what the button does instead of provide a description? If it's an "X" icon, we would say "Close" instead of "X Icon" right? Wonder what would be the appropriate label for this bell.

I know it doesn't _really_ matter in storybook, but I'm just trying to learn best practices and make sure we apply them in all examples.

ccedrone: Do you think "Clickable icon" would work better, something along those lines?

silvalaura: I think it's unclear in this case because who knows why there would be a Bell inside that input. In this example, it makes sense that the EventIcon would have the `Calendar` label. 
```
      <Input
        labelText="Large input with clickable icon on right"
        iconAriaLabel="Calendar"
        inputSize="large"
        onIconClick={() => {
          alert('Icon clicked');
        }}
        icon={<EventIcon />}
      />
```

I would probably guess that the Bell means notifications, so I would label it `Notifications` to follow the pattern above.

silvalaura (approved):

Issue #317, opened 2021-07-12
react-magma-docs - Update display of props


orion-cengage:
This update is only for the new version 3 of the React Magma docs.

**Purpose:** The props are currently displayed in tables, and due to the content of these tables, they aren't providing a very good experience on anything other than the largest of screens.  We're going to try a new approach.

**Figma:** https://www.figma.com/file/pdcWRIHYjeu8y09XzAT1SB/React-Magma-Docs?node-id=17%3A12983

### AC's

- Each prop will get its own two-column layout.
-- Start with the prop name
-- Two-column section consists of the attribute title on the left (Type, Default, Description) and the full details of that attribute on the right.
![image](https://user-images.githubusercontent.com/24397282/183144109-01fe0519-8d6a-4869-b273-494103e61226.png)

- Required props called out with **small primary tag** instead of the previous asterisk method.
![image](https://user-images.githubusercontent.com/24397282/183144215-ba2a7213-1d5c-4aaa-a739-323f6997f669.png)

- I'm hoping we can keep this same two-column layout all the way down to 320px - 375px.  We might need to tweak once we see it in action.
![image](https://user-images.githubusercontent.com/24397282/183421586-a91eba8e-0a02-447b-8393-b33e021008bf.png)

-- It's possible we may need to go to a single column layout on small screens.
![image](https://user-images.githubusercontent.com/24397282/183421704-5e5bbd06-cdf4-468b-98c7-3cf520bc297c.png)

jamiesmyln:
This is ready on feature/docs-props-table but we're waiting to merge it until v3 is ready for it.

PR #869, opened 2022-08-10
docs(simplepropstable): updating layout of props tables

silvalaura:
#317

We don't currently have any deprecated props, so here's an example of how that would look:
image

### DIFF
--- a/.changeset/docs-props-table.md
+++ b/.changeset/docs-props-table.md
@@ -1,5 +1,5 @@
 ---
-'react-magma-dom': patch
+'react-magma-docs': patch
 ---
 
 docs: Updating layout of props tables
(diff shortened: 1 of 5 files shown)


orion-cengage (changes requested):

- Reduce the margin below the prop title from 24px to 16px
image

- Change font-weight of prop title to bold
image

- On pages, like Accordion, where we're using H2, H3, and H4 close to each other, it doesn't feel right.  The current H3 feels unnecessary because we're repeating those words in the H4.  We'll be better off removing the current H3s, and then changing the current H4s to H3s.
image
image

- Please add an additional 32px of bottom margin to the bottom of the props tables.  This will help with multiple tables by increasing the space between the bottom of the table and the next heading.
image

- On small screen, the total space between the title and value should only be 4px.  It's currently 16px.  But maintain the 16px of space between each group of title and value.
image
image

orion-cengage (approved):

greenervigil (approved):

Issue #870, opened 2022-08-15
react-magma-docs: Replace sidebar accordion with our own component


silvalaura:
Replaced the Magma site sidebar accordion component with our own.

Previous accordion: https://react-magma.cengage.com/version/3.1.1/

Our accordion: https://next--upbeat-sinoussi-f675aa.netlify.app/

Uditsingh67:
We Verified this ticket.
See Screen-cast : http://somup.com/c3QIhDUvp2
Thanks!

PR #899, opened 2022-08-29
Docs/side bar accordion

ccedrone:
Issue: # [870](https://github.com/cengage/react-magma/issues/870)

## What I did
Updated docs sidebar to use our Accordion

## Checklist 
- [n/a] New and existing unit tests pass locally with my changes
- [n/a] I have added tests that prove my fix is effective or that my feature works


silvalaura:
Removing the animation looks fine when changing between pages, but looks odd when just opening and closing the Accordions 😕  I wonder if we can tweak it some more for the collapsing state?

ccedrone:
Agreed! I'll keep tweaking.

silvalaura (approved):

silvalaura:
Since we are removing the only usage of the `react-accessible-accordion` package, we should also delete it from the package.json

orion-cengage (approved):

Issue #891, opened 2022-08-19
Inputs enhancement: Character counter


orion-cengage:
**Purpose:** Define and add character counter functionality to text-based inputs.  This will provide valuable feedback to users when there are limits put on how many characters are allowed for submission.

**Figma:** https://www.figma.com/file/VoWYiiOKKppcYm5QLgPyDC/Form-Inputs?node-id=1%3A12778

**Inspiration:** https://designsystem.digital.gov/components/character-count/

Note: I have a feeling this will not simply use the maxlength attribute for inputs.  If that was the case, then it would stop you from entering more than the allowed characters and that's not exactly what we're going for here.

### AC

1.0 Verify when Character Counter is enabled, it appears below the input with the same styling as standard Helper Text, and that it gets the character limit from the [maxlength] attribute.
![image](https://user-images.githubusercontent.com/24397282/185651126-8683073d-b56e-44ef-b4b4-c63ace5bf520.png)
-- 1.1 Verify you can also still enable Helper Text and it appears below the Character Counter.
     ![image](https://user-images.githubusercontent.com/24397282/185651170-b6386240-7a19-4d18-901e-28ea1e226cf7.png)

2.0 Verify as you type characters into the input, the Character Counter updates and reflects how many allowed characters are left.
![image](https://user-images.githubusercontent.com/24397282/185651235-72178176-e3db-4ed6-adaa-3a5681227834.png)

3.0 Verify once you start typing characters over the limit, the input reacts by showing its error state, and the character counter's message reflects this as well.
     ![image](https://user-images.githubusercontent.com/24397282/185969744-b88a987b-f82b-4376-9d92-0afa91c102d6.png)
     ![image](https://user-images.githubusercontent.com/24397282/185969821-d2517f07-adf5-4240-871f-9833eafd7b2c.png)

-- 3.1 Verify the input can support multiple error messages in addition to the character counter.
     ![image](https://user-images.githubusercontent.com/24397282/185969929-c7cbc737-9cd6-4476-84ab-9d274f7d02bc.png)

4.0 Verify the character counter can be enabled on the following components:

- Input
- Text area

5.0 Correct grammar - Depending on the number of characters referenced in the message, we need to make sure the message is grammatically correct.
-- 5.1
--- 1 character allowed
--- 20 characters allowed
--- 0 characters left
--- 1 character left
--- 5 characters left
--- 1 character over limit
--- 40 characters over limit 

6.0 Accessibility
[...]

PR #917, opened 2022-09-23
Feat/character counter

ccedrone:
Issue: # [891](https://github.com/cengage/react-magma/issues/891)

## What I did

Character Counter component to be used within Input and Textarea.

## Checklist 

## To Dos
- [x] Accessibility testing - Mac
- [x] Accessibility testing - Windows
- [x] Unit test code coverage
- [x] Cross browser testing (Safari, FF)


packages/react-magma-dom/src/components/Input/index.tsx
```diff
@@ -54,11 +74,13 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
             descriptionId ? descriptionId : props['aria-describedby']
           }
           aria-invalid={!!errorMessage}
-          hasError={!!errorMessage}
+          hasError={!!errorMessage || characterLength > maxLength}
           iconPosition={iconPosition}
           id={id}
           inputSize={inputSize}
           isInverse={isInverse}
+          onChange={maxLength ? handleChange : props.onChange}
```
silvalaura, line 82: I think we can handle this like this instead of having the ternary on this call:

https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/InputBase/index.tsx#L508

https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/InputBase/index.tsx#L477-L483

packages/react-magma-dom/src/components/Input/index.tsx
```diff
@@ -54,11 +74,13 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
+          onClear={maxLength ? handleClear : props.onClear}
```
silvalaura, line 83: Same with onClear. It's important to type check and ensure that `props.onClear` is actually a function before we make a call for it

```
function handleOnClear() {
   onClear && typeof onClear === 'function' && onClear();
   if (maxLength) {
      setCharacterLength(null);
   }
}
```

Also, why set the character length to `null`?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { FormFieldContainerProps } from '../FormFieldContainer';
+import { ThemeContext } from '../../theme/ThemeContext';
+import { ThemeInterface } from '../../theme/magma';
+import { InputSize } from '../InputBase';
+import { I18nContext } from '../../i18n';
+
+export interface CharacterCounterProps
+  extends Omit<FormFieldContainerProps, 'fieldId'>,
+    React.HTMLAttributes<HTMLDivElement> {
+  activeStying?: boolean;
```
silvalaura, line 13:
```suggestion
  activeStyling?: boolean;
```

silvalaura: Although, why is this prop?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,129 @@
...
+  hasError?: boolean;
+  inputSize?: InputSize;
```
silvalaura, line 15: Does CharacterCounter really need to know the `inputSize`?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,129 @@
...
+    ...rest
+  } = props;
+
+  const theme = React.useContext(ThemeContext);
+
+  const i18n = React.useContext(I18nContext);
+
+  const activeStyling = inputTotal < maxLength || inputTotal == maxLength;
+
+  const errorStyling = inputTotal > maxLength;
```
silvalaura, line 78: If these are opposite, do we need both variables?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,232 @@
...
+import { axe } from '../../../axe-helper';
+import { Input } from '../Input';
+import { CharacterCounter } from '.';
+import { transparentize } from 'polished';
+import { magma } from '../../theme/magma';
+import { render, fireEvent } from '@testing-library/react';
+import { I18nContext } from '../../i18n';
+import { defaultI18n } from '../../i18n/default';
+
+const testId = 'test-id';
```
silvalaura, line 11: Since you use `const labelText = 'Character Counter';` in every test, you can also declare that up here instead of recreating the variable in each test.

silvalaura: Same with `characterLeft` and `characterOver` since you're pulling that from i18n and won't change

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,232 @@
...
+    expect(onChange).toHaveBeenCalled();
+    expect(getByText('2 ' + charactersOver)).toHaveStyleRule(
+      'color',
+      magma.colors.danger200
+    );
+
+    expect(getByText('2 ' + charactersOver)).toBeInTheDocument();
+  });
+
+  it('Shows the label "characters left" equal to the maxLength if the user clears the input', () => {
```
silvalaura, line 211: I think it would help with legibility of tests (and so that we can ensure we've covered every case we need to) to group tests based on characterLeft, characterOver, etc

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,139 @@
...
+      return props.theme.colors.danger200;
+    }
+    return transparentize(0.3, props.theme.colors.neutral100);
+  }
+  if (props.hasError) {
+    return props.theme.colors.danger;
+  }
+}
+
+const StyledCharacterCounter = styled.div<{
```
silvalaura, line 36: Looks like `StyledCharacterCounter` is the div wrapper around the words `X characters allowed`, not a wrapper for the input. You don't need border-radius or even the flex properties on this. Does the margin between the text and input need to differ depending on the input size?

ccedrone: `flex` is still needed for when the glyph appears while characters are over the limit.

silvalaura: Should we use the `<InputMessage>` component then? That seems to do the styling and show the icon, then we don't have to handle that here. Could be worth looking into.

- [Component](https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/Input/InputMessage.tsx)
- [Usage example](https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/RadioGroup/index.tsx#L178-L186)

ccedrone: Seems like it works just as well, switched `<InputMessage` in place.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,107 @@
...
+  inputLength?: number;
+  /**
+   * Sets the maximum amount of characters allowed.
+   */
+  maxLength?: number;
+  testId?: string;
+  /**
+   * @internal
+   */
+  theme?: ThemeInterface;
```
silvalaura, line 24: Theme does not need to be a prop of CharacterCounter (you don't pass theme to it when you call it in FormFieldContainer)

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+function buildFontWeight(props: CharacterCounterProps) {
+  if (
+    (props.inputLength < props.maxLength && props.inputLength >= 1) ||
+    props.inputLength === props.maxLength
+  ) {
+    return props.theme.typographyVisualStyles.headingXSmall.fontWeight;
+  }
+}
+const StyledWrapper = styled.div``;
```
silvalaura, line 35: This doesn't seem to be used. Can it be removed?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,107 @@
...
+  theme?: ThemeInterface;
+}
+
+function buildFontWeight(props: CharacterCounterProps) {
+  if (
+    (props.inputLength < props.maxLength && props.inputLength >= 1) ||
+    props.inputLength === props.maxLength
+  ) {
+    return props.theme.typographyVisualStyles.headingXSmall.fontWeight;
+  }
```
silvalaura, line 33: Should this have a default return?

packages/react-magma-dom/src/components/Input/Input.test.js
```diff
@@ -474,4 +476,216 @@ describe('Input', () => {
...
+
+        fireEvent.change(getByLabelText(labelText), {
+          target: { value: 'dddddd' },
+        });
+
+        expect(onChange).toHaveBeenCalled();
+        expect(getByText('2 ' + charactersOver)).toBeInTheDocument();
+      });
+
+      it('Shows the inverse label "characters over limit" as the user types over the limit', () => {
```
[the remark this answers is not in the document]
ccedrone: Redundant test, removed.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+    if (inputLength === maxLength) {
+      return `0 ${i18n.characterCounter.charactersLeft}`;
+    }
+  }
+
+  // Character Counter default "allowed" title states.
+  function characterTitle() {
+    if (inputLength > 0) {
+      return characterTitleCount();
```
silvalaura, line 84: Is there a reason why `characterTitleCount` is a separate function instead of having that logic directly in here? I guess we can argue that for legibility it may be nice to keep it all together. But, if the logic in `characterTitleCount` needed to be used in multiple places, a function makes more sense.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  return (
+    <StyledWrapper data-testid={testId} aria-live="polite">
+      <StyledInputMessage
+        hasError={isOverMaxLength}
+        isInverse={isInverse}
+        inputLength={inputLength}
+        maxLength={maxLength}
+        theme={theme}
```
silvalaura, line 100: Is `theme` needed here?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
ccedrone: `return 'inherit'` work?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
ccedrone: I'm using this for the JSX `<StyledWrapper data-testid={testId} aria-live="polite">`, should it be done another way?

silvalaura: It's not actually styled though, so why not just use `<div data-testid={testId} aria-live="polite">`?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
silvalaura: Yup!

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  it('Does not violate accessibility standards', () => {
+    const { container } = render(<CharacterCounter maxLength={22} />);
+
+    return axe(container.innerHTML).then(result => {
+      return expect(result).toHaveNoViolations();
+    });
+  });
+});
```
silvalaura, line 24: As far as CharacterCounter unit tests, we should write some that directly test this component. So for example, some of the tests that you wrote that belong here would be:

```
it('Shows the default label of "characters allowed" if maxLength is 0', () => {
        const { getByText } = render(
          <I18nContext.Provider
            value={{
              ...defaultI18n,
            }}
          >
            <CharacterCounter maxLength={0} />
          </I18nContext.Provider>
        );

        expect(getByText('0 ' + charactersAllowed)).toBeInTheDocument();
      });

      it('Shows the default label of "character allowed" if maxLength is 1', () => {
        const { getByText } = render(
          <I18nContext.Provider
            value={{
              ...defaultI18n,
            }}
          >
            <CharacterCounter maxLength={1} />
          </I18nContext.Provider>
        );

        expect(getByText('1 ' + characterAllowed)).toBeInTheDocument();
      });
      it('Shows the default label of "characters allowed" if maxLength is greater than 1', () => {
        const { getByText } = render(
          <I18nContext.Provider
            value={{
              ...defaultI18n,
            }}
          >
            <CharacterCounter maxLength={2} />
          </I18nContext.Provider>
        );

        expect(getByText('2 ' + charactersAllowed)).toBeInTheDocument();
      });
    });
```

Notice above that the change I made was call `<CharacterCounter maxLength={1} />` **instead of** `<Input maxLength={1} />`. In addition, we want to add tests that directly test the other CharacterCounter prop of `inputLength`. 

Then, any test that checks things like `onChange` or any Input prop should live under the Input tests files. In addition, we want to add tests similar to the Input ones to Textarea to test that the CharacterCounter works within that component as well. 

Let me know if you need further clarification.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
ccedrone: On it!

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
ccedrone: Not anymore, removed.

packages/react-magma-dom/src/components/Input/Input.test.js
```diff
@@ -474,4 +476,216 @@ describe('Input', () => {
...
+        fireEvent.change(getByLabelText(labelText), {
+          target: { value: 'dddddd' },
+        });
+
+        expect(onChange).toHaveBeenCalled();
+        expect(getByText('2 ' + charactersOver)).toBeInTheDocument();
+      });
+    });
+
+    it('Shows the label "characters left" equal to the maxLength if the user clears the input', () => {
```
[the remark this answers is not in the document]
ccedrone: Yup, wanted 'characters allowed', changed the title and added another test within this test.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,102 @@
...
+   */
+  inputLength?: number;
+  /**
+   * Sets the maximum amount of characters allowed.
+   */
+  maxLength?: number;
+  testId?: string;
+  /**
+   * @internal
+   */
```
silvalaura, line 22:
```suggestion
  /**
   * @internal
   */
   testId?: string;
```

website/react-magma-docs/src/pages/api/character-counter.mdx
```diff
@@ -0,0 +1,70 @@
+---
+pageTitle: CharacterCounter API
+title: CharacterCounter
+props:
+  - InputBaseProps
```
silvalaura, line 5: This should be `CharacterCounterProps`

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,102 @@
+import * as React from 'react';
+import styled from '../../theme/styled';
+import { FormFieldContainerProps } from '../FormFieldContainer';
+import { I18nContext } from '../../i18n';
+import { InputMessage } from '../Input/InputMessage';
+import { magma } from '../../theme/magma';
+
+export interface CharacterCounterProps
+  extends Omit<FormFieldContainerProps, 'fieldId'>,
```
silvalaura, line 9: Do we need `Omit<FormFieldContainerProps, 'fieldId'>,`?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+      labelText={labelText}
+      errorMessage={args.errorMessage}
+      helperMessage={args.helperMessage}
+      maxLength={args.maxLength}
+    />
+  );
+};
+TextArea.args = {
+  ...Default.args,
+};
```
silvalaura, line 102: We should exclude `isClearable` from the Textarea story because textareas are not clearable:

`TextArea.parameters = { controls: { exclude: ['isClearable'] } };`

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
ccedrone: It's a required prop that Character Counter doesn't need correct?

silvalaura: What this is saying is CharacterCounterProps will be `CharacterCounterProps + FormFieldContainerProps - fieldId`. But CharacterCounterProps doesn't need to extend FormFieldContainerProps because it doesn't need any of the FormFieldContainerProps like actionable, containerStyle, isLabelVisuallyHidden, etc.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
ccedrone: Moved into one function.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,123 @@
...
+
+  it('Does not violate accessibility standards', () => {
+    const { container } = render(<CharacterCounter maxLength={22} />);
+
+    return axe(container.innerHTML).then(result => {
+      return expect(result).toHaveNoViolations();
+    });
+  });
+  describe('Titles', () => {
+    const labelText = 'Character Counter';
```
silvalaura, line 39: `labelText` is not used in this file (only in Input because Inputs have labels)

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -0,0 +1,99 @@
...
+   */
+  inputLength?: number;
+  /**
+   * Sets the maximum amount of characters allowed.
+   */
+  isInverse?: boolean;
+  /**
+   * Enables the Character Counter and sets the maximum amount of characters allowed within the Input.
+   */
+  maxLength?: number;
```
silvalaura, line 20: Should `inputLength` and `maxLength` be required props for this component? Without those props this component would be useless, right?

I think we want them to be required for CharacterCounter but optional for Input, Textarea, FormField container.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,123 @@
...
+              ...defaultI18n,
+            }}
+          >
+            <CharacterCounter maxLength={1} />
+          </I18nContext.Provider>
+        );
+
+        expect(getByText('1 ' + characterAllowed)).toBeInTheDocument();
+      });
+      it('Shows the default label of "characters allowed"', () => {
```
silvalaura, line 69:
```suggestion
      it('Shows the default label of "characters allowed" if maxLength > 1', () => {
```

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,123 @@
...
+            <CharacterCounter maxLength={2} />
+          </I18nContext.Provider>
+        );
+
+        expect(getByText('2 ' + charactersAllowed)).toBeInTheDocument();
+      });
+    });
+
+    describe('Characters Left', () => {
+      it('Shows the label "characters left" as the user types', () => {
```
silvalaura, line 85: So here we aren't testing the user typing. This component has a prop `inputLength` that just gives it the length of the input. So a better description would be something like `Shows the label "characters left" when inputLength > 1`

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,123 @@
...
+        const { getByText } = render(
+          <CharacterCounter inputLength={2} maxLength={4} />
+        );
+        expect(getByText('2 ' + charactersLeft)).toBeInTheDocument();
+      });
+
+      it('Shows the label "characters left" if user types to zero', () => {
```
silvalaura, line 92: Here I would say something like `Shows the label "characters left" when maxLength is equal to inputLength`

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,123 @@
...
+        const { getByText } = render(
+          <CharacterCounter inputLength={4} maxLength={4} />
+        );
+        expect(getByText('0 ' + charactersLeft)).toBeInTheDocument();
+      });
+
+      it('Shows the label "character left" as the user types to one remaining character', () => {
```
silvalaura, line 99: Same as above, just tweak this description to refer to the component props instead of a user action

packages/react-magma-dom/src/components/Input/Input.test.js
```diff
@@ -474,4 +475,67 @@ describe('Input', () => {
...
+
+      fireEvent.change(getByLabelText(labelText), {
+        target: { value: '' },
+      });
+
+      expect(getByLabelText(labelText)).toHaveAttribute('value', '');
+      expect(getByText('4 ' + charactersAllowed)).toBeInTheDocument();
+    });
+
+    describe('styling', () => {
```
silvalaura, line 506: I think styling tests can move to CharacterCounter as they are not Input specific

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+          onClick={onHelpLinkClick}
+          type={ButtonType.button}
+          size={ButtonSize.small}
+          variant={ButtonVariant.link}
+        />
+      </Tooltip>
+    </Input>
+  );
+};
+
```
silvalaura, line 85: Adding the default args will make this story have a character counter by default
```suggestion
WithChildren.args = {
  ...Default.args,
}
```

packages/react-magma-dom/src/components/Input/Input.test.js
ccedrone: I did the same thing in 'Textarea.test.js', should that also be removed?

silvalaura: Yeah! This is the kind of test that we only need once if we put it in CharacterCounter.

silvalaura (commented):
The package-lock.json does not need to be committed unless changes were made to the package.json

website/react-magma-docs/src/pages/api/input.mdx
```diff
@@ -465,7 +478,7 @@ export function Example() {
 Passing in the `testId` prop will add the `data-testid` attribute to the input element for easier querying in tests.
 
 ```html noCodeSandbox
-  <Input labelText="Label" testId="test-id"/>
+<input labelText="Label" testId="test-id" />
```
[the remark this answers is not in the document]
ccedrone: On save, VS Code made it lowercase for some reason.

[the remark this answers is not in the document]
ccedrone: Prettier caused the issue, fixed.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,149 @@
...
+      });
+    });
+
+    describe('styling', () => {
+      it('Shows the error glyph if user exceeds maxLength prop', () => {
+        const icon = <ErrorIcon />;
+        const { container, rerender } = render(
+          <>
+            <CharacterCounter inputLength={3} maxLength={4} />
+            {icon}
```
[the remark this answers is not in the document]
ccedrone: Updated test.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
```diff
@@ -0,0 +1,141 @@
...
+              ...defaultI18n,
+            }}
+          >
+            <CharacterCounter maxLength={2} />
+          </I18nContext.Provider>
+        );
+
+        expect(getByText('2 ' + charactersAllowed)).toBeInTheDocument();
+      });
+    });
```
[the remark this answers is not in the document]
silvalaura: Also, `inputLength` is a required prop for `CharacterCounter` and therefore needs to be used in these tests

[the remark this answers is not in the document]
ccedrone: > Why are these test renders wrapped in `<I18nContext.Provider>` but the ones below are not?

I initially though they were needed to get the titles, but I later discovered they weren't. Removed.

silvalaura (approved):
⭐

greenervigil (approved):

silvalaura:
👍  on design guidelines

orion-cengage (approved):

Issue #909, opened 2022-09-06
Bug: Help icon on inputs not in the correct place


orion-cengage:
Version found: 3.x

https://react-magma.cengage.com/version/3.1.1/api/input/#additional_children_(e.g._help_link)

The help icon on inputs is displaying inside the input, but it is supposed to be above the input. 
![image](https://user-images.githubusercontent.com/24397282/188652892-f9efc33a-c8bd-4809-b916-783c9d8c0d23.png)
 
This is currently displaying correctly in 2.x
![image](https://user-images.githubusercontent.com/24397282/188652766-6fd8b155-7bbe-4f18-9c14-6f772f4f8848.png)

Possible solution: Instead of forcing it to appear above the input using CSS positioning, should the element itself render outside of the container?

silvalaura:
Good find. I think I missed this because this example of additional children is not in storybook. We should make sure we have this example there too.

silvalaura:
**How to test**
- Test all Inputs (Input, PasswordInput, DatePicker, Search, etc) and ensure they are behaving as expected
- Specifically, test Input > With Children and confirm that the component visually looks how we expect. Test medium, large, isClearable.

Uditsingh67:
The issue is fixed.
See Screen-cast : http://somup.com/c3611QvVjn
Thanks!

PR #918, opened 2022-09-26
fix(input): fix placement of `children` inside Input component 

silvalaura:
Issue #909

## What I did

Fix placement of `children` passed to Input component. Made it compatible with `isClearable` (which it wasn't previously)

**before**
image
image

**after**
image
image

## Checklist 

## How to test
- Test all Inputs (Input, PasswordInput, DatePicker) and ensure they are behaving as expected
- Specifically, test Input > With Children and confirm that the component visually looks how we expect. Test `medium`, `large`, `isClearable`.

### DIFF
--- a/.changeset/fix-helpicon.md
+++ b/.changeset/fix-helpicon.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Input): Fix placement of `children` inside Input component (ex: help icon).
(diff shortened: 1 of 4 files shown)


orion-cengage (changes requested):
Looks good except for one thing.  The icon should be a little smaller on the normal sized input - 20px instead of 24px.  It should be 24px on the large input.

image

orion-cengage (approved):

Issue #953, opened 2022-11-01
Multi Combobox: when used inside a form, deleting an item submits the form


silvalaura:
**Describe the bug**
After a user selects an item, if they were to delete it by clicking on it, this action would cause the form that the multiCombobox is within to submit. I think the issue is that the button that gets rendered when a user selects an item doesn’t have its “type” set to  “button” and defaults to “submit.”

**To Reproduce**
1. Go to https://codesandbox.io/s/react-magma-multi-combobox-issue-when-changing-item-list-forked-u2iwzu?file=/example.tsx
2. If you select the red item, the form gets submitted. This only seems to happen when there is more than one item selected

**Expected behavior**
Deleting an item does not submit the form

Uditsingh67:
@silvalaura 
Where should i test this ticket?

silvalaura:
@Uditsingh67 This storybook example is wrapped in a form
https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/combobox--full-page-example

Uditsingh67:
The issue is **fixed.**
**See Screen-cast :** http://somup.com/c3XffAvp7c
Thanks!

PR #954, opened 2022-11-01
fix(combobox): items have a default `type="button"` to prevent issues inside form

silvalaura:
Issue: #953 

## What I did

Added `type="button"` to the items inside a combobox. This should prevent the whole form submitting when an item gets removed.

_No visual change, just a change in the DOM_
image

## Checklist 

## How to test

- Confirm that after selecting more than one item, clicking one item to delete it does not submit the form

### DIFF
--- a/packages/react-magma-dom/src/components/Combobox/Combobox.stories.tsx
+++ b/packages/react-magma-dom/src/components/Combobox/Combobox.stories.tsx
@@ -5,9 +5,10 @@ import { SelectOptions } from '../Select';
 import { LabelPosition } from '../Label';
 import { Card } from '../Card';
 import { CardBody } from '../Card/CardBody';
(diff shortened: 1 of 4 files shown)


ccedrone (approved):

packages/react-magma-dom/src/components/Combobox/Combobox.stories.tsx
```diff
@@ -1204,59 +1205,68 @@ Typeahead.args = {
 };
 
 export const FullPageExample = args => {
+  function handleSubmit() {
+    alert('Form Submitted');
+  }
```
greenervigil, line 1210: This could be an action instead of an alert

greenervigil (approved):

PR #956, opened 2022-11-02
fix(navtabs): Added focus state on first NavTab for accessibility. New `isFocused` prop

ccedrone:
Issue: #

## What I did
Doc site, Added `ref` to `Introduction` link for keyboard navigation.
NavTabs, added focus state to first NavTab for accessibility fix.

## Checklist 

## How to test
Ensure after site loads that a focus goes to the next item below `Introduction` (Contribution Guidelines) on mobile.


packages/react-magma-dom/src/components/NavTabs/NavTabs.tsx
```diff
@@ -77,6 +77,7 @@ export const NavTabs = React.forwardRef<
       orientation={orientation || TabsOrientation.horizontal}
       ref={ref}
       theme={theme}
+      tabIndex={0}
```
silvalaura, line 80: I think this line is causing the focus to start on the Nav container, then each individual tab:
chrome-capture-2022-10-8

Is this okay for accessibility?

.changeset/docs-focus.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-docs': minor
```
silvalaura, line 2:
```suggestion
'react-magma-docs': patch
```

Also, if there are changes for dom as well, we should add a second changeset

packages/react-magma-dom/src/components/NavTabs/NavTabs.tsx
ccedrone: I moved that `tabIndex` to `Drawer` instead, seems to help with isolating the use case of a nav element without effecting the entire `NavTabs` component.

greenervigil (commented):
I ended up with two failed tests when running locally:

[ ] NavTabs.test.js:18:27
[ ] NavTabs.test.js:34:29

greenervigil (approved):

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
```diff
@@ -30,6 +30,10 @@ export interface BaseNavTabProps
    * If true, the component will display with the active/selected state
    */
   isActive?: boolean;
+  /**
+   * If true, sets a focus on the specified NavTab
+   */
+  isFocused?: boolean;
```
silvalaura, line 36: I may be missing something, but I'm wondering why this needs to be prop. `isFocused` is a boolean, not a tab index, so can we by default always focus the first tab, or is there a reason why we shouldn't?

ccedrone: This approach was used to satisfy the `cloneElement` so `NavTab` could pass something to `NavTabs`, aside from that, it's always going to focus the first one by default. Should that change?

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
```diff
@@ -89,6 +93,7 @@ const StyledTab = styled.a<{
   isActive?: boolean;
   isFullWidth?: boolean;
   isInverse?: boolean;
+  isFocused?: boolean;
```
silvalaura, line 96: I don't think this needs to be passed to TabStyles if TabStyles doesn't use it, right?

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
```diff
@@ -182,12 +196,14 @@ export const NavTab = React.forwardRef<any, NavTabProps>((props, ref) => {
           isInverse={isInverse}
           isFullWidth={isFullWidth}
           orientation={orientation}
+          ref={styledTabRef}
           theme={theme}
         />
       ) : (
         <StyledTab
           {...other}
-          ref={ref}
```
silvalaura, line 190: Does this ref need to be removed? I just want to make sure we aren't breaking any functionality that currently exists in case someone is using it.

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
silvalaura: I see. Does this work as we expect if someone passes `isFocused` to the second or third tab child?

.changeset/drawer-focus.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(drawer) Added tabIndex for focusing content when opened.
```
silvalaura, line 5: We should update this now that the implementation has changed

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
ccedrone: That might be an older pull, it should be showing `ref={styledTabRef}`.

.changeset/drawer-focus.md
ccedrone: "I see. Does this work as we expect if someone passes isFocused to the second or third tab child?" This works on any NavTab as well indeed.

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
silvalaura: Looks like `ref={ref}` was only on `StyledTab` but you've added `ref={styledTabRef}` to `StyledCustomTab`
image

ccedrone: In case the user wants to have the same focus with an icon based NavTab as well.

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
```diff
@@ -155,6 +160,15 @@ export const NavTab = React.forwardRef<any, NavTabProps>((props, ref) => {
     ? TabsIconPosition.left
     : TabsIconPosition.top;
 
+  const styledTabRef = React.useRef<HTMLAnchorElement>();
+
+  // Sets focus on first NavTab for accessibility
```
silvalaura, line 165: Sorry to nitpick, but isFocused can be any tab so we should remove `first` from this comment

packages/react-magma-dom/src/components/NavTabs/NavTab.tsx
ccedrone: This works on any NavTab as well indeed.

silvalaura (approved):

Issue #904, opened 2022-08-30
New Component: Toggle Button


orion-cengage:
**Purpose:** Create a new Toggle Button component one or more buttons that can be toggled on and off.

**Figma:** https://www.figma.com/file/VoWYiiOKKppcYm5QLgPyDC/Components---Inputs?node-id=75%3A10654

![image](https://user-images.githubusercontent.com/24397282/197209974-a765e374-ecaa-497c-8bff-3d61192d6093.png)

### AC

1.0 The component uses the "Subtle" button styles, which **_can't_** be changed.

2.0 The component can consist of a single button on its own, or as a group of buttons.  Styling of the buttons (space between, border-radius) needs to update automatically.
![image](https://user-images.githubusercontent.com/24397282/197210185-98a498c2-8771-41e6-9766-bb34f7d50d17.png)

![image](https://user-images.githubusercontent.com/24397282/197211076-c7ce8715-3c87-4910-9556-060d769b4409.png)

![image](https://user-images.githubusercontent.com/24397282/197210216-4a5c70fa-097d-4aff-bd2a-86083473bb98.png)

![image](https://user-images.githubusercontent.com/24397282/197211153-28542815-331f-4c3d-9912-9714f166f79e.png)

3.0 Buttons may contain text only, icon only, or text + icon

4.0 Each button within the component needs to turn on and off via click/tap

5.0 Button states

-- 5.1 When the button is "on", it uses the styling of the "pressed/active" Secondary button.
![image](https://user-images.githubusercontent.com/24397282/197210398-f4da8d58-b130-4cda-872c-55f43d5ddc8e.png)

-- 5.2 Regardless of whether the button is currently on or off, it gets the hover and pressed styling when triggered

-- 5.3 A button can be disabled, but when it is in this component, the border styling needs to be consistent with the rest of the buttons.
![image](https://user-images.githubusercontent.com/24397282/197210538-c0004d1e-21f2-411c-8351-918d8e34ef14.png)

6.0 With exclusive selection set, selecting one option deselects any other.

7.0 With multiple selection set, multiple buttons can be on at the same time.

8.0 You can set any button to be turned on by default.

9.0 You can enforce that **_at least one button_** in the group must be on.

10.0 You can set the size small, medium (default), and large.
![image](https://user-images.githubusercontent.com/24397282/197210654-aeef01a3-f500-4a54-90ae-70bae3e8b59a.png)

[...]

PR #985, opened 2022-12-20
feat(toggleButton): New ToggleButton & ToggleButtonGroup components

ccedrone:
Issue: #904

## What I did
Adding
- ToggleButton component
- ToggleButtonGroup component

image
image
image

## Checklist 

## How to test
- Navigate to the ToggleButton component in storybook. Test all the controls
- Navigate to the ToggleButtonGroup component in storybook. Test all the controls. Confirm that `enforced` and `exclusive` work as expected
- Verify the documentation pages for both ToggleButton and ToggleButtonGroup


packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,231 @@
...
+    }
+  };
+
+  return (
+    <>
+      {icon ? (
+        <StyledToggleButtonIcon
+          aria-checked={isChecked}
+          aria-label={ariaLabel}
+          aria-pressed={isChecked}
```
silvalaura, line 187: `The attribute aria-pressed is not supported by the role switch.`

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * @children required
+ */
+export interface ToggleButtonGroupProps
+  extends Omit<
+    React.HTMLAttributes<HTMLButtonElement>,
+    'isDefault' | 'isChecked'
+  > {
+  isChecked?: boolean;
+  isDefault?: boolean;
```
silvalaura, line 23: The ButtonGroup can't be checked or default

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+    >;
+
+    const disabledButton = item.props.disabled;
+
+    // const checkedButton = item.props.isChecked;
+
+    const handleClick = () => {
+      if (requiredSelect && !disabledButton && singleSelect) {
+        //Retains at least one selection for single select
+        setSelected(s => (s === index ? index : index));
```
silvalaura, line 108: This conditional isn't needed if it sets the value to the same thing whether it's true or false

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const theme = React.useContext(ThemeContext);
+
+  const [isSelected, setSelected] = React.useState(null);
+
+  const [isDefault, setDefault] = React.useState(null);
+
+  const ToggleButtons = React.Children.map(children, (child, index) => {
+    const item = child as React.ReactElement<
+      React.PropsWithChildren<ToggleButtonProps>
+    >;
```
silvalaura, line 99: Why is this needed over just `{children}`?

website/react-magma-docs/src/pages/api/toggle-button.mdx
```diff
@@ -0,0 +1,189 @@
...
+import { ToggleButton } from 'react-magma-dom';
+
+export function Example() {
+  return <ToggleButton>Toggle</ToggleButton>;
+}
+```
+
+## Toggle Button Group
+
+The `ToggleGroup` enables styling with an optional `noSpace` prop, and different selection states `singleSelect` and `requiredSelect`.
```
silvalaura, line 26: Should we show all these options in the example?

website/react-magma-docs/src/pages/api/toggle-button.mdx
```diff
@@ -0,0 +1,189 @@
...
+}
+```
+
+## Types
+
+Toggle Buttons come in three formats. Icon only, text only, or text with an icon.
+
+```typescript
+import React from 'react';
+import { SettingsIcon, ToggleButton } from 'react-magma-dom';
```
silvalaura, line 50: Missing `ToggleButtonGroup` import

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.test.js
```diff
@@ -0,0 +1,197 @@
...
+        <ToggleButton size={ButtonSize.large} icon={icon} testId={testId} />
+      );
+      const button = getByTestId(testId);
+
+      expect(button).toHaveStyleRule('width', magma.spaceScale.spacing11);
+      expect(button).toHaveStyleRule('height', magma.spaceScale.spacing11);
+    });
+  });
+
+  describe('Grouping', () => {
```
silvalaura, line 128: We should have a separate file for ToggleButtonGroup tests as they are two separate components.

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+   */
+  theme?: ThemeInterface;
+}
+
+export interface ToggleButtonGroupContextInterface {
+  isDefault?: boolean;
+  isInverse?: boolean;
+  requiredSelect?: boolean;
+  isChecked?: boolean;
+  singleSelect?: boolean;
```
silvalaura, line 56: I'm a little confused as to why some of these are part of the context. The context should only hold data that all the ToggleButtons need to know such as inverse and size, and maybe requiredSelect. Why do we need `isDefault` and `isChecked`? In your implementation you only use inverse and size.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,231 @@
...
+  return props.theme.spaceScale.spacing09;
+}
+
+//Sets the background colors for the Toggle Buttons.
+export function setBackgroundColor(props: ToggleButtonProps) {
+  if (
+    (props.isChecked && props.defaultChecked) ||
+    (!props.isChecked && props.isDefault)
+  ) {
+    return '';
```
silvalaura, line 114: Maybe we return `'none'` as we are returning a background color instead of blank?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,231 @@
...
+  const theme = React.useContext(ThemeContext);
+
+  const [isChecked, setChecked] = React.useState(false);
+
+  const handleClick = () => {
+    if (isChecked) {
+      setChecked(false);
+    } else if (!isChecked) {
+      setChecked(true);
+    }
```
silvalaura, line 178: When adopters use this, they are not only going to expect the button to change visually, but the buttons should likely trigger some logic as well. We need to add a call to their own button click function here. Also make sure to add an example of this in the docs/storybook to test it.

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+  React.createContext<ToggleButtonGroupContextInterface>({});
+
+const SingleSelectWrapper = styled.div<{
+  isDefault?: boolean;
+  isInverse?: boolean;
+  requiredSelect?: boolean;
+  isChecked?: boolean;
+  singleSelect?: boolean;
+  size?: ButtonSize;
+  value?: any;
```
silvalaura, line 70: ToggleButtonStyles only uses some of the props, so we can remove the extra ones from here.

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    const handleClick = () => {
+      if (requiredSelect && !disabledButton && singleSelect) {
+        //Retains at least one selection for single select
+        setSelected(s => (s === index ? index : index));
+      } else if (!disabledButton) {
+        //Allows single select to be deselected
+        setSelected(s => (s === index ? null : index));
+      }
+      if (handleClick && !disabledButton) {
```
silvalaura, line 113: handleClick is the name of this function, no need to check if it exists here

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const ToggleButtons = React.Children.map(children, (child, index) => {
+    const item = child as React.ReactElement<
+      React.PropsWithChildren<ToggleButtonProps>
+    >;
+
+    const disabledButton = item.props.disabled;
+
+    // const checkedButton = item.props.isChecked;
+
+    const handleClick = () => {
```
silvalaura, line 105: Why does the SingleSelectWrapper need an onClick function? The user is clicking the buttons not the wrapper

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
ccedrone: This was causing issues on docs and has been removed here.

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
ccedrone: Removed `isChecked`, but the rest are all used by `ToggleButton`.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,231 @@
...
+  /**
+   * Sets a disabled state for a button.
+   */
+  disabled?: boolean;
+  /**
+   * Icon which displays alongside text.
+   */
+  icon?: React.ReactElement<IconProps>;
+  isChecked?: boolean;
+  isDefault?: boolean;
```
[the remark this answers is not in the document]
ccedrone: Removed `isDefault` from that table.

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
ccedrone: This prevents unchecking a required singular selection, do you suggest reworking it?

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
silvalaura: You're right, I'm wrong. We need this :)

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,199 @@
...
+  //Sets active background color.
+  if (props.isChecked || props.defaultChecked || props.isDefault) {
+    if (props.isInverse) {
+      return transparentize(0.5, props.theme.colors.neutral900);
+    }
+    return transparentize(0.5, props.theme.colors.neutral300);
+  }
+}
+
+export const setButtonStyles = props => css`
```
silvalaura, line 87: Because these are Button styles, should they live in ToggleButton and then get imported here in ToggleGroup?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+    size: {
+      control: {
+        type: 'select',
+        options: ButtonSize,
+      },
+    },
+  },
+} as Meta;
+
+export const Default = Template.bind({});
```
silvalaura, line 42: Should we have a story for ToggleButtonIcon and one for ToggleButtonText?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,170 @@
...
+export interface ToggleButtonTextProps extends ButtonProps {
+  /**
+   * Enables a specified button to be active.
+   */
+  defaultChecked?: boolean;
+  /**
+   * Sets a disabled state for a button.
+   */
+  disabled?: boolean;
+  isChecked?: boolean;
```
silvalaura, line 27: Still a little confused as to why we need both `defaultChecked` and `isChecked` 🤔 I think ideally we figure out a way to combine these two and think through how adopters will use this.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const context = React.useContext(ToggleButtonGroupContext);
+
+  const theme = React.useContext(ThemeContext);
+
+  const [isChecked, setChecked] = React.useState(false);
+
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    props.onClick &&
+      typeof props.onClick === 'function' &&
+      props.onClick(event);
```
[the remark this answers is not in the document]
silvalaura: Also, if you destructure `onClick` in line 92, you can just use 
```
onClick && typeof onClick === 'function' && onClick(event);
```

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.test.js
```diff
@@ -0,0 +1,67 @@
...
+const TEXT = 'Test Text';
+
+const testId = 'test-id';
+
+const icon = <SettingsIcon />;
+
+describe('ToggleButtonGroup', () => {
+  it('Should find element by testId', () => {
+    const { getByTestId } = render(
+      <ToggleButton testId={testId}>{TEXT}</ToggleButton>
```
silvalaura, line 18:
```suggestion
      <ToggleButtonGroup testId={testId}>{TEXT}</ToggleButtonGroup>
```

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.test.js
```diff
@@ -0,0 +1,67 @@
...
+    );
+
+    expect(getByTestId(testId)).toBeInTheDocument();
+  });
+
+  it('Icon only buttons are compliant with accessibility', () => {
+    // const { container } = render(
+    //   <ToggleButton icon={icon} aria-label="Icon Button" />
```
silvalaura, line 26:
```suggestion
    //   <ToggleButtonGroup></ToggleButtonGroup>
```

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,199 @@
...
+  requiredSelect?: boolean;
+  singleSelect?: boolean;
+  size?: ButtonSize;
+}
+
+export const ToggleButtonGroupContext =
+  React.createContext<ToggleButtonGroupContextInterface>({});
+
+//Sets the icon width for icon only Toggle Buttons.
+export function setIconWidth(props: ToggleButtonProps) {
```
silvalaura, line 56: I'd also move this to ToggleButton

packages/react-magma-dom/src/components/ToggleButton/ToggleButtonGroup.tsx
silvalaura: As it's written, it is saying:
```
if (s === index) {
   return index;
} else {
   return index;
}
```
So it's returning index either way. Not sure if you need to have a conditional, or just `setSelected(index)`

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  // Sets the active state for the Toggle Button
+  const [selected, setSelected] = React.useState(null);
+
+  // Sets a specific selected state with the value prop.
+  const [selectedValue, setSelectedValue] = React.useState<string>(props.value);
+
+  React.useEffect(() => {
+    setSelectedValue(props.value);
+  }, [props.value]);
```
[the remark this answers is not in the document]
silvalaura: Same thing for line 99

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,203 @@
...
+  /**
+   * Icon which displays alongside text.
+   */
+  icon: React.ReactElement<IconProps>;
+  isChecked?: boolean;
+  isInverse?: boolean;
+  /**
+   * Shows the button in an active state.
+   */
+  selected?: boolean;
```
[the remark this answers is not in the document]
ccedrone: `isChecked` is looking for the manually set `selected` state by `value` then lets the user remove the active toggled state like the others. `selected` is the boolean for on and off.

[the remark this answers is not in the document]
silvalaura: Based on that, can we just have `isChecked` as the prop and `selected` continues to be the state (but is not a prop)?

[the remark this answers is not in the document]
silvalaura: Remove `selected` from the props table. It's a variable that you need for the component logic, but the user will not be setting it. Instead, users will set `isChecked` so that does need to be a prop.

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,182 @@
...
+          {item}
+        </StyledWrapper>
+      );
+    }
+    return child;
+  });
+
+  return (
+    <StyledToggleButtonGroup
+      aria-describedby="Toggle button group"
```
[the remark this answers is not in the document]
ccedrone: `aria-describedby={props['aria-describedby']}` Seem okay?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,203 @@
...
+  return (
+    <>
+      {icon ? (
+        <StyledToggleButtonIcon
+          aria-checked={selected}
+          aria-label={ariaLabel}
+          color={ButtonColor.subtle}
+          disabled={disabled}
+          theme={theme}
+          hasLabel={children ? true : false}
```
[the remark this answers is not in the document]
ccedrone: For some reason just using `children` as a check doesn't work with the styling.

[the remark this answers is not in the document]
silvalaura: Right. `hasLabel` is expecting a boolean, `children` is a node. You have to convert `children` to a boolean. One way would be to check if `children` has children (lol) like `React.Children.count(children) > 0` (source: https://stackoverflow.com/questions/69104115/how-to-check-if-a-child-element-exists-in-parent-in-react).

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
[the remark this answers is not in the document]
silvalaura: I guess we should first answer the question, what is the best solution for accessibility? then we can figure out the implementation.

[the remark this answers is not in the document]
ccedrone: It makes sense to have unique descriptions set by the user. We could also have a generated `id` as a default if necessary to mimic the other components with `aria-describedby`. Thoughts on thoughts?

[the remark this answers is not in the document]
silvalaura: Then yes, we can do it that way. Do we need to add that same ID to the [elements it's describing](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby)?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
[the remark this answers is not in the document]
ccedrone: Per discussion, keeping `hasLabel`.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+export interface ToggleButtonTextProps extends ButtonProps {
+  /**
+   * Sets a disabled state for a button.
+   */
+  disabled?: boolean;
+  /**
+   * @internal
+   */
+  isChecked?: boolean;
```
silvalaura, line 23: If someone is using a single ToggleButton that is initialized as check, how do we expect them to do that?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,203 @@
...
+export interface ToggleButtonIconProps extends ButtonProps {
+  /**
+   * Sets the aria-label which is required for icon only buttons.
+   */
+  'aria-label'?: string;
+  /**
+   * Sets a disabled state for a button.
+   */
+  disabled?: boolean;
+  hasLabel?: boolean;
```
silvalaura, line 44: Missing prop description

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,203 @@
...
+          icon={icon}
+          id={context.descriptionId}
+          isChecked={
+            context.selectedValue ? context.selectedValue === value : null
+          }
+          isInverse={context.isInverse || isInverse}
+          onClick={handleClick}
+          ref={ref}
+          enforced={context.enforced}
+          role="switch"
```
silvalaura, line 168: Is the role `switch` even when exclusive is true?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,203 @@
...
+          color={ButtonColor.subtle}
+          disabled={disabled}
+          theme={theme}
+          hasLabel={children ? true : false}
+          icon={icon}
+          id={context.descriptionId}
+          isChecked={
+            context.selectedValue ? context.selectedValue === value : null
+          }
+          isInverse={context.isInverse || isInverse}
```
silvalaura, line 164: Since you do this `context.isInverse || isInverse` check twice, consider making it a reusable variable

website/react-magma-docs/src/pages/api/toggle-button-group.mdx
```diff
@@ -0,0 +1,229 @@
...
+  return (
+    <ToggleButtonGroup singleSelect>
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+    </ToggleButtonGroup>
+  );
+}
+```
```
silvalaura, line 73: This can be removed as the `exclusive` example is below

website/react-magma-docs/src/pages/api/toggle-button-group.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+## Single Select
+
+Enforces one toggled state across a single select.
```
silvalaura, line 77: I like the description above a little better though: "Allows just one toggled button across the group in behavior like a radio."

website/react-magma-docs/src/pages/api/toggle-button-group.mdx
```diff
@@ -0,0 +1,229 @@
...
+        defaultChecked
+        aria-label="Check icon"
+        icon={<CheckIcon />}
+      />
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+    </ToggleButtonGroup>
+  );
+}
+```
```
silvalaura, line 119: It doesn't look like we have `defaultChecked` anymore

website/react-magma-docs/src/pages/api/toggle-button-group.mdx
```diff
@@ -0,0 +1,229 @@
...
+  return (
+    <ToggleButtonGroup noSpace singleSelect>
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
+    </ToggleButtonGroup>
+  );
+}
+```
```
silvalaura, line 140: dupe, can be replaced with `enforced` example

website/react-magma-docs/src/pages/api/toggle-button-group.mdx
```diff
@@ -0,0 +1,229 @@
...
+} from 'react-magma-dom';
+
+export function Example() {
+  return (
+    <Container style={{ padding: '12px' }}>
+      <ToggleButtonGroup>
+        <ToggleButton size={ButtonSize.small}>Toggle 1</ToggleButton>
+        <ToggleButton size={ButtonSize.medium}>Toggle 2</ToggleButton>
+        <ToggleButton size={ButtonSize.large}>Toggle 3</ToggleButton>
+      </ToggleButtonGroup>
```
silvalaura, line 195: Because this is the Group doc, I would show the example on the group: 
```suggestion
      <ToggleButtonGroup size={ButtonSize.small}>
        <ToggleButton>Toggle 1</ToggleButton>
        <ToggleButton>Toggle 2</ToggleButton>
        <ToggleButton size={ButtonSize.large}>Toggle 3</ToggleButton>
      </ToggleButtonGroup>
```

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,199 @@
...
+    testId,
+    ...rest
+  } = props;
+  const theme = React.useContext(ThemeContext);
+
+  // Sets the active state for the Toggle Button
+  const [selected, setSelected] = React.useState(null);
+
+  // Sets a specific selected state with the value prop.
+  const [selectedValue, setSelectedValue] = React.useState<string>(value);
```
silvalaura, line 107: What if the user wants to pre-select multiple buttons? Should this be an array instead of a string?

silvalaura: We can check with Orion if this is a requirement or not.

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    if (item.type === ToggleButton && props.exclusive) {
+      return (
+        <StyledWrapper
+          key={index}
+          onChange={handleChange}
+          onClick={handleClick}
+          isInverse={isInverse}
+          enforced={exclusive ? selected === index : null}
```
silvalaura, line 149: enforced has a boolean value. instead of `null`, could that be `false`?

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -0,0 +1,199 @@
...
+          descriptionId,
+          selectedValue,
+          isInverse,
+          enforced,
+          exclusive,
+          onChange: handleChange,
+          size,
+        }}
+      >
+        {console.log(descriptionId)}
```
silvalaura, line 194: Remove console log

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.test.js
```diff
@@ -0,0 +1,133 @@
...
+describe('ToggleButtonGroup', () => {
+  it('Should find element by testId', () => {
+    const { getByTestId } = render(
+      <ToggleButtonGroup testId={testId}></ToggleButtonGroup>
+    );
+
+    expect(getByTestId(testId)).toBeInTheDocument();
+  });
+
+  it('Icon only buttons are compliant with accessibility', () => {
```
silvalaura, line 25:
```suggestion
  it('Toggle Button Groups are compliant with accessibility', () => {
```

packages/react-magma-dom/src/components/ToggleButtonGroup/ToggleButtonGroup.test.js
```diff
@@ -0,0 +1,133 @@
...
+        'background',
+        transparentize(0.5, magma.colors.neutral300)
+      );
+
+      fireEvent.click(getByTestId(testId));
+
+      expect(getByTestId(testId)).toHaveStyleRule('background', 'none');
+    });
+  });
+});
```
silvalaura, line 133: Can we add a few more tests:
- check that `onChange` gets called
- verify `enforced` prop
- if we change `value` to a string, we should test with 1 value and then with multiple

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.test.js
```diff
@@ -0,0 +1,185 @@
...
+  });
+
+  it('Icon only buttons are compliant with accessibility', () => {
+    const { container } = render(
+      <ToggleButton icon={icon} aria-label="Icon Button" />
+    );
+    return axe(container.innerHTML).then(result => {
+      return expect(result).toHaveNoViolations();
+    });
+  });
```
silvalaura, line 32: Wouldn't hurt to also have an accessibility compliant test for non-icon toggle buttons

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
ccedrone: I didn't know if that was a requirement, but I added it in.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
ccedrone: This should be internal only I believe?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
ccedrone: Should it be something else you think?

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
ccedrone: There are a couple of other props with that similarity. Does it make sense to keep them as is? Seems like six in one half dozen in the other.

silvalaura: Your call! Variables are great for reusability so if in the future something changes, it only has to be changed in one place. Up to you to decide where/if it makes sense.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
silvalaura: I'm not 100% on the best accessibility behavior here, I'd triple check MUI/other similar components. Without researching, in my brain it's a "switch" when you can select multiple and "radio" when it's exclusive.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
silvalaura: If we don't expect users to use it, then it shouldn't be a prop and it should just be a variable within the component.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
silvalaura: What I meant was, what code do we expect users to use in order to start a single toggle button as checked. Is it something like this, or something else?

```
<ToggleButton isChecked>My Toggle</ToggleButton>
```

ccedrone: I made it work with `isChecked`.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
ccedrone: I don't see `role` being used in other component libraries really. I could just make it go between `switch` and `radio` depending on if the prop `exclusive` is used or not.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
silvalaura: If that's the case, then it's not an `internal` prop. Then, we should make sure we have an example of this in the docs.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.tsx
ccedrone: I can clean that up for legibility then.

website/react-magma-docs/src/pages/api/toggle-button-group.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+```typescript
+import React from 'react';
+import { ToggleButton, ToggleButtonGroup } from 'react-magma-dom';
+import { CheckIcon } from 'react-magma-icons';
+
+export function Example() {
+  return (
+    <ToggleButtonGroup>
+      <ToggleButton aria-label="Check icon" icon={<CheckIcon />} />
```
silvalaura, line 24: All examples on this page will also need to be updated to include `value`

packages/react-magma-dom/src/components/Input/Input.stories.tsx
```diff
@@ -11,11 +11,7 @@ import { ButtonSize, ButtonType, ButtonVariant } from '../Button';
 
 const Template: Story<InputProps> = args => (
   <>
-    <Input {...args} labelText="Input label" />
-    <br />
-    <Input {...args} labelText="Second Input label" />
-    <br />
-    <Input {...args} labelText="Third Input label with a long title" />
+    <Input {...args} labelText="Input label" width="100%"/>
```
silvalaura, line 14: This was an accidental push from me, we should revert it.

packages/react-magma-dom/src/components/ToggleButton/ToggleButton.stories.tsx
katieweinstein: + 1 what Laura said!

katieweinstein (approved):
As a note, Storybook has an "Actions" feature that you can use to hook up functions instead of using `console.log`. Not necessary, just wanted to point that out if you ever run into a messy console!

orion-cengage (approved):

hkrishna29:
**AC:**

01. The component uses the "Subtle" button styles, which can't be changed.
02. The component can consist of a single button on its own, or as a group of buttons. Styling of the buttons (space between, 
      border-radius) needs to update automatically.
03. Buttons may contain text only, icon only, or text + icon
04. Each button within the component needs to turn on and off via click/tap
05. Button states
    5.1 When the button is "on", it uses the styling of the "pressed/active" Secondary button.
    5.2 Regardless of whether the button is currently on or off, it gets the hover and pressed styling when triggered
    5.3 A button can be disabled, but when it is in this component, the border styling needs to be consistent with the rest of the 
         buttons.
06. With exclusive selection set, selecting one option deselects any other.
07. With multiple selection set, multiple buttons can be on at the same time.
08. You can set any button to be turned on by default.
09. You can enforce that at least one button in the group must be on.
10.  You can set the size small, medium (default), and large.
11  Accessibility
     11.1 Some good accessibility rules can be found here: https://mui.com/material-ui/react-toggle-button/#accessibilit
     11.2 Keyboard - Toggle buttons are in DOM order. Navigate between them with the tab key. The button behavior follows 
             standard keyboard semantics.

hkrishna29:
**AC:**

01. The component uses the "Subtle" button styles, which can't be changed.
02. The component can consist of a single button on its own, or as a group of buttons. Styling of the buttons (space between, 
      border-radius) needs to update automatically.
03. Buttons may contain text only, icon only, or text + icon
04. Each button within the component needs to turn on and off via click/tap
05. Button states
    5.1 When the button is "on", it uses the styling of the "pressed/active" Secondary button.
    5.2 Regardless of whether the button is currently on or off, it gets the hover and pressed styling when triggered
    5.3 A button can be disabled, but when it is in this component, the border styling needs to be consistent with the rest of the 
         buttons.
06. With exclusive selection set, selecting one option deselects any other.
07. With multiple selection set, multiple buttons can be on at the same time.
08. You can set any button to be turned on by default.
09. You can enforce that at least one button in the group must be on.
10.  You can set the size small, medium (default), and large.
11  Accessibility
     11.1 Some good accessibility rules can be found here: https://mui.com/material-ui/react-toggle-button/#accessibilit
     11.2 Keyboard - Toggle buttons are in DOM order. Navigate between them with the tab key. The button behavior follows 
             standard keyboard semantics.

PR #1002, opened 2023-02-01
fix(inputbase): update width prop

katieweinstein:
Issue: #

## What I did
Added width prop to InputBase so that width is passed through to the input element.

## Checklist 

### DIFF
--- a/packages/react-magma-dom/src/components/InputBase/index.tsx
+++ b/packages/react-magma-dom/src/components/InputBase/index.tsx
@@ -114,6 +114,11 @@ export interface InputBaseProps
    * Boolean for whether this is a Password Input or not
    */
   isPasswordInput?: boolean;
+  /**
+   * String to determine width of input, must be suffixed with "px", "rem", or "%""
+   * @default "auto"
+   */
+  width?: string;
 }
 
 export interface InputWrapperStylesProps {
@@ -574,6 +579,7 @@ export const InputBase = React.forwardRef<HTMLInputElement, InputBaseProps>(
           style={containerStyle}
           hasError={hasError}
           isClearable={isClearable}
+          width={props.width}
         >
           <StyledInput
             {...other}
--- a/packages/react-magma-dom/src/components/PasswordInput/index.tsx
+++ b/packages/react-magma-dom/src/components/PasswordInput/index.tsx
@@ -48,6 +48,11 @@ export interface PasswordInputProps
    * @internal
    */
   testId?: string;
+  /**
+   * String to determine width of input, must be suffixed with "px", "rem", or "%""
+   * @default "auto"
+   */
+  width?: string;
 }
 
 export const PasswordInput = React.forwardRef<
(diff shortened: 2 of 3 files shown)


packages/react-magma-dom/src/components/InputBase/index.tsx
```diff
@@ -114,6 +114,10 @@ export interface InputBaseProps
    * Boolean for whether this is a Password Input or not
    */
   isPasswordInput?: boolean;
+  /**
+   * String to determine width of input
+   */
+  width: string;
```
silvalaura, line 120: I think we want to make this prop optional:
```suggestion
  width?: string;
```

.changeset/fix-inputWrapper.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-magma-dom": patch
+---
+
+fix(inputbase): update width prop
```
silvalaura, line 5: The changeset is what gets used to populate our [changelog](https://react-magma.cengage.com/version/3.2.1/api-intro/changelog/). The more descriptive we are here, the easier it is for adopters to understand what changes we introduce in each version.

silvalaura (commented):
Looks like the build is breaking because of Passport Input. We may need to add the `width` prop to all inputs that use InputBase.

packages/react-magma-dom/src/components/InputBase/index.tsx
```diff
@@ -114,6 +114,10 @@ export interface InputBaseProps
    * Boolean for whether this is a Password Input or not
    */
   isPasswordInput?: boolean;
+  /**
+   * String to determine width of input
```
silvalaura, line 118:
```suggestion
   * String to determine width of input, must be suffixed with "px", "rem", or "%""
```

packages/react-magma-dom/src/components/PasswordInput/index.tsx
```diff
@@ -112,6 +117,8 @@ export const PasswordInput = React.forwardRef<
 
   const isInverse = useIsInverse(props.isInverse);
 
+  console.log(props.width);
+
```
silvalaura, line 121: We should remove this

.changeset/fix-inputWrapper.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(inputbase): Add and update width prop for InputBase
```
silvalaura, line 5: InputBase is used under the hood for many components, but adopters don't know this. We should mention the specific components it impacts: Input, PasswordInput.

Does this need to also be applied to Search?

silvalaura (commented):
Ideally, we shouldn't push the package-lock. I wonder if that caused the build to fail?

silvalaura (approved):

ccedrone (approved):

hkrishna29:
The preview of the components is looking fine at 400 %

Issue #992, opened 2023-01-17
Support adding children on the right side of Alerts


silvalaura:
We want to support the ability to add different components such as buttons or badges on the right side on an alert. It should replace the space that the dismissible buttons takes up when alerts are not dismissible, or be to the left of the X when the alert is dismissible.

![image](https://user-images.githubusercontent.com/91160746/212998185-c5fe4d8c-5009-4132-96c1-f71f1f533c69.png)

**Use cases:** 
- Ability to add a button/badge when an alert is dismissible
- Ability to add a button/badge when an alert is NOT dismissible
- When the alert text is long, it should wrap and should not interfere with the newly added space on the right

**Acceptance Criteria:**
- Documentation updates to Alert and Banner to illustrate example on how to do this using flexbox. This example should show with and without the close button.

hkrishna29:
Badges and buttons are displayed right hand side of the alert , where the buttons are clickable and badges are not.  
http://somup.com/c0nrYoyhcx

PR #1004, opened 2023-02-02
feat - Support adding children on the right side of Alerts

ccedrone:
Issue: # [992](https://github.com/cengage/react-magma/issues/992)

## What I did
Added `additionalContent` prop to Alerts and Banners for right aligned children elements such as buttons and badges.

example

## Checklist 

### DIFF
--- a/packages/react-magma-dom/src/components/AlertBase/index.tsx
+++ b/packages/react-magma-dom/src/components/AlertBase/index.tsx
@@ -295,9 +295,10 @@ const AlertContents = styled.div<{
 `;
 
 export const AdditionalContentWrapper = styled.div`
-  flex: 1;
+  flex: 1 0 auto;
   justify-content: flex-end;
   display: flex;
+  margin-left: ${props => props.theme.spaceScale.spacing04};
 `;
 
 const IconWrapperStyles = css`
--- a/packages/react-magma-dom/src/components/AlertBase/index.tsx
+++ b/packages/react-magma-dom/src/components/AlertBase/index.tsx
@@ -289,10 +289,6 @@ const AlertContents = styled.div<{
     props.additionalContent && !props.isDismissible
       ? props.theme.spaceScale.spacing03
       : ''};
-  * {
-    margin-left: ${props => (props.additionalContent ? '7px' : '')};
-  }
-
   @media (max-width: ${props => props.theme.breakpoints.small}px) {
     padding-left: 0;
   }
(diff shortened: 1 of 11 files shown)


packages/react-magma-dom/src/components/AlertBase/index.tsx
```diff
@@ -275,16 +276,39 @@ const StyledAlertInner = styled.div<AlertBaseProps>`
...
 
+export const AdditionalContentWrapper = styled.div`
+  flex: 1;
+  justify-content: flex-end;
+  display: flex;
+  * {
+    display: table-cell;
+    margin: 0;
+    margin-left: ${props => props.theme.spaceScale.spacing03};
+  }
```
silvalaura, line 309: Why is this needed? It seems like I can remove these styles and it still works the same

packages/react-magma-dom/src/components/AlertBase/index.tsx
```diff
@@ -444,6 +471,7 @@ export const AlertBase = React.forwardRef<HTMLDivElement, AlertBaseProps>(
         theme={theme}
         variant={variant}
       >
+        {console.log(additionalContent)}
```
silvalaura, line 474: We should remove this

packages/react-magma-dom/src/components/Banner/index.tsx
```diff
@@ -187,6 +194,7 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
     const {
       actionButtonText,
       actionButtonOnClick,
+      additionalContent,
```
silvalaura, line 197: This needs to be added to the BannerProps

silvalaura (commented):
We should also add some unit tests and examples in the docs site

packages/react-magma-dom/src/components/AlertBase/index.tsx
ccedrone: If you look at the Storybook example with two badges, and also the button, it cleans up the left spacing and fixes an issue with left button padding.

ccedrone: Fair enough, I'll remove that then!

silvalaura (approved):

orion-cengage (changes requested):
Couple things I noticed.

- Can we make sure there is 16px of space/margin/whatever to the left of the container with the badge or button, so the text never runs into the item?
- For some reason, the badges are wrapping and that's not supposed to happen. They should stay a single line like the button. Is this a problem with the alert, or the badge component?

image

ccedrone:
Flexbox needed a tweak, added a left margin of 12px to the additional children container as well.

orion-cengage (approved):

katieweinstein (approved):

Issue #991, opened 2023-01-17
New component: Pagination variant


orion-cengage:
Purpose: We need a new variant on the existing pagination component. It will be a better option for all-sized screens when you can expect more than 5 or 6 pages of results.

Figma: https://www.figma.com/file/cPCfDVWC38T7Qt3dIL95sR/Components---Navigation?node-id=1%3A2&t=5s8dkFy7tdXHpYLp-1

AC

1.0  Component has two buttons that navigate to the Previous Page and Next Page.
-- 1.1  The Previous Page button is disabled when you're on the first page.
-- 1.2  The Next Page button is disabled when you're on the last page.
-- 1.3  Put tooltips on both buttons with labels "Previous page" and "Next page".
2.0  Display the page you're currently on, as well as choose a specific page, using the Native Select component.
3.0  Display the number of pages using a text label.
4.0  The width of the component is dependent on the container it is inside of. But if the component gets too wide, the integrity of the design deteriorates.
-- 4.1  Add max-width on the component of 360px
5.0  Accessibility
--  Keyboard tab order is pretty obvious
-- Visually hidden label on the component: “Page number ##, of 6 pages”.
-- Use “aria-hidden” to hide the text label from the screen reader.

Implementation Notes:
- We will add a new prop to pagination, so users can choose between the two options (@orion-cengage to name these). This will allow us to have a single docs page for this component, and we can display examples for both.

silvalaura:
**Testing notes**
re: @ccedrone 

1.0 Component has two buttons that navigate to the Previous Page and Next Page.
-- 1.1 The Previous Page button is disabled when you're on the first page. ✅ 
-- 1.2 The Next Page button is disabled when you're on the last page.✅
-- 1.3 Put tooltips on both buttons with labels "Previous page" and "Next page".✅
2.0 Display the page you're currently on, as well as choose a specific page, using the Native Select component.✅
3.0 Display the number of pages using a text label.✅
4.0 The width of the component is dependent on the container it is inside of. But if the component gets too wide, the integrity of the design deteriorates. 🚫 
-- 4.1 Add max-width on the component of 360px ✅ 
5.0 Accessibility
-- Keyboard tab order is pretty obvious✅
-- Visually hidden label on the component: “Page number ##, of 6 pages”. 🚫 
-- Use “aria-hidden” to hide the text label from the screen reader. ✅

Other areas tested:
- Docs look good ✅
- Code sandbox example 🚫 
- Storybook looks good ✅

[...]

PR #1031, opened 2023-03-10
feat: Pagination variant

ccedrone:
Issue: # [991](https://github.com/cengage/react-magma/issues/991)

## What I did
- New pagination variant adapting pagination into a dropdown selector.

hooray

## Checklist 

## How to test

- Make sure the type `PaginationType.simple` works as a dropdown with two navigation buttons.
- Ensure the doc site has the appropriate changes.


packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,175 @@
...
+        fieldId={id}
+        isInverse={isInverse}
+        onChange={handleChange}
+        theme={theme}
+        value={selectedValue}
+      >
+        {Array.from({ length: count }, (_, i) => (
+          <option
+            data-testid={`${testId}-option`}
+            key={i + count}
```
silvalaura, line 142: Why `i + count`? This can just be `i`

packages/react-magma-dom/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -10,13 +9,35 @@ const Template: Story<PaginationProps> = args => <Pagination {...args} />;
...
+    hidePreviousButton: {
+      control: {
+        type: 'boolean',
+      },
+    },
+    hideNextButton: {
+      control: {
+        type: 'boolean',
+      },
+    },
```
silvalaura, line 40: Let's add the simple prop here

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,175 @@
...
+            onChange={handleChange}
+            value={i + 1}
+          >
+            {i + 1}
+          </option>
+        ))}
+      </StyledNativeSelect>
+      <Spacer size={8} />
+      <label aria-hidden="true">of {count} pages</label>
```
silvalaura, line 151: We should i18n these. Also, we need to account for when `count` is 1, it should say `page`

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+`;
+
+const StyledNativeSelect = styled(NativeSelect)<{
+  disabled?: boolean;
+}>`
+  select {
+    color: ${props => (props.disabled ? props.theme.colors.neutral500 : '')};
+    cursor: ${props => (props.disabled ? 'not-allowed' : '')};
+  }
```
silvalaura, line 48: Just curious why this doesn't work by default with our NativeSelects?

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
```diff
@@ -0,0 +1,178 @@
...
+      <Pagination simple count={4} defaultPage={3} testId={testId} />
+    );
+    let nextButton = getByLabelText('Next Page');
+
+    fireEvent.click(nextButton);
+
+    expect(nextButton).toHaveStyleRule(
+      'color',
+      transparentize(0.4, magma.colors.neutral500)
+    );
```
silvalaura, line 117: In addition to checking the color changes, can we check that the element is disabled?

packages/react-magma-dom/src/components/Pagination/Pagination.tsx
```diff
@@ -118,7 +124,7 @@ function BuildButtonSize(props) {
   }
 }
 
-const NavButton = styled(IconButton)`
+export const NavButton = styled(IconButton)`
```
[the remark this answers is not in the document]
ccedrone: Indeed, it's used in 'SimplePagination'.

silvalaura:
@orion-cengage Do we want to support both size medium and large for Simple? 
The large currently looks odd, maybe because the text is too small, maybe because it's odd for the arrows to be that large?
image

packages/react-magma-dom/src/components/Pagination/Pagination.tsx
```diff
@@ -65,6 +66,11 @@ export interface BasePaginationProps
    * @default false
    */
   showLastButton?: boolean;
+  /**
+   * If true, shows the Pagination pages in a dropdown style.
+   * @default false
+   */
+  simple?: boolean;
```
silvalaura, line 73: **Question for consideration:**
In other components when we have different types/versions, we use a `type` prop, and we set a default (example: Button, LoadingIndicator, List, Input). Should we match that for consistency, or is it better to have a boolean prop?

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
```diff
@@ -0,0 +1,178 @@
...
+
+  it('Should change the active page when an option is selected', () => {
+    const { getByTestId, getAllByTestId } = render(
+      <Pagination simple count={4} testId={testId} />
+    );
+    fireEvent.change(getByTestId(`${testId}-select`), { target: { value: 2 } });
+    let options = getAllByTestId(`${testId}-option`);
+    expect(options[0].selected).toBeFalsy();
+    expect(options[1].selected).toBeTruthy();
+    expect(options[2].selected).toBeFalsy();
```
silvalaura, line 140:
```suggestion
    expect(options[0].selected).toBe(false);
    expect(options[1].selected).toBe(true);
    expect(options[2].selected).toBe(true);
```

_[source](**url**)_

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
```diff
@@ -0,0 +1,178 @@
...
+
+    expect(previousButton).toHaveStyleRule(
+      'color',
+      transparentize(0.7, magma.colors.neutral100)
+    );
+    expect(nextButton).toHaveStyleRule(
+      'color',
+      transparentize(0.7, magma.colors.neutral100)
+    );
+  });
```
silvalaura, line 90: Just a style recommendation, but I like using `describe`s to group together similar tests, or tests related to the same prop. So these could be wrapped in
```
describe('disabled', () => { 
   it ('', () => {});
   it ('', () => {});
})
```
I think it helps organize them and it's easier to see props have what coverage. I did this in ButtonGroup tests, for example.

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
```diff
@@ -0,0 +1,178 @@
...
+
+  it('Does not violate accessibility standards', () => {
+    const { container } = render(<Pagination simple count={4} />);
+
+    return axe(container.innerHTML).then(result => {
+      return expect(result).toHaveNoViolations();
+    });
+  });
+});
```
silvalaura, line 178: Lets add tests for edge cases:
- what if defaultPage or count is 0? -> what do we expect to happen?
- what if defaultPage or count is 1? 

What about coverage for the other Pagination props: hidePreviousButton, numberOfAdjacentPages, etc etc.

ccedrone:
> @orion-cengage Do we want to support both size medium and large for Simple? The large currently looks odd, maybe because the text is too small, maybe because it's odd for the arrows to be that large? image

Removed per Orion, just the medium size will remain.

packages/react-magma-dom/src/components/Pagination/Pagination.tsx
ccedrone: Converted to an enum for consistency!

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
ccedrone: All set, removed those repetitive attributes.

packages/react-magma-dom/src/components/Pagination/Pagination.tsx
```diff
@@ -65,6 +66,16 @@ export interface BasePaginationProps
...
   showLastButton?: boolean;
+  /**
+   * If true, shows the Pagination pages in a dropdown style.
+   * @default button
+   */
+  type?: PaginationType;
+}
+
+export enum PaginationType {
+  default = 'button',
```
silvalaura, line 77: I think we should call the types `simple` and `classic` (at least that's what Orion used in the designs)
```suggestion
  classic = 'classic',
```

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
```diff
@@ -0,0 +1,290 @@
...
+      const nextButton = getByLabelText('Next Page');
+
+      expect(nextButton).toHaveStyleRule(
+        'color',
+        transparentize(0.4, magma.colors.neutral500)
+      );
+    });
+  });
+
+  describe('Active', () => {
```
silvalaura, line 149: These look great 👏

website/react-magma-docs/src/pages/api/pagination.mdx
```diff
@@ -28,6 +28,19 @@ export function Example() {
 }
 ```
 
+## Simple Pagination
+
+An alternative presentation of `Pagination` is availble with the `simple` prop. This converts the existing structure into a dropdown format, while still utilizing previous and next navigation buttons.
```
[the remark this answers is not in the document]
silvalaura: Also, this would be a good place to point out what props can't be used with this type

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
```diff
@@ -0,0 +1,290 @@
...
+    it('Should disable the previous button on the first page', () => {
+      const { getByLabelText } = render(
+        <Pagination type={PaginationType.simple} count={4} defaultPage={1} />
+      );
+      const previousButton = getByLabelText('Previous Page');
+
+      expect(previousButton).toHaveStyleRule(
+        'color',
+        transparentize(0.4, magma.colors.neutral500)
+      );
```
silvalaura, line 133: Instead of checking the colors, we could expect the button `.toBeDisabled()`. The color checking should happen in the Button tests because that's part of the button component. The logic in this component just cares that the element became disabled. 

_(same feedback below for next button)_

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
silvalaura: > Lets add tests for edge cases:
> 
> * what if defaultPage or count is 0? -> what do we expect to happen?
> * what if defaultPage or count is 1?
> 
> What about coverage for the other Pagination props: hidePreviousButton, numberOfAdjacentPages, etc etc.

Looks like the tests for the extra props have been addressed, but would still love to see more unit tests from the first two points, especially since you are now handling the singular/plural "pages" label change

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,190 @@
...
+  display: flex;
+  align-items: center;
+  max-width: 360px;
+  select {
+    min-width: 87px;
+  }
+  label {
+    color: ${buildLabelColor};
+    font-size: ${props => props.theme.typeScale.size02.fontSize};
+    font-weight: 500;
```
silvalaura, line 41: I recently went through and added font-family to every component so for consistency, do you mind adding it here too?

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
silvalaura: Oooh, and tests that when a page gets changed (1. from previous button 2. from next button 3. from dropdown), the `onPageChange` function gets called

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,190 @@
...
+  function paginationLabel() {
+    if (count <= 1) {
+      return `${i18n.simplePagination.ofLabel}
+        ${count}
+        ${i18n.simplePagination.pageLabel}`;
+    } else {
+      return `${i18n.simplePagination.ofLabel}
+        ${count}
+        ${i18n.simplePagination.pagesLabel}`;
+    }
```
silvalaura, line 101: This works great and you can just leave it. But **if** you want an extra little challenge, can you think of another way to write this with less code repeat? hint: it can be done in a single line :)

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,190 @@
...
+      isInverse={isInverse}
+      ref={ref}
+    >
+      {!hidePreviousButton && (
+        <StyledPrevTooltip content={i18n.pagination.previousButtonLabel}>
+          <NavButton
+            aria-label={i18n.pagination.previousButtonLabel}
+            variant={ButtonVariant.link}
+            color={ButtonColor.secondary}
+            disabled={selectedPage < 2 ? true : disabled}
```
silvalaura, line 134: This looks like the same logic in the `disabledPrevTooltip` variable. Can that be used here?

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,190 @@
...
+      </VisuallyHidden>
+
+      <Spacer size={14} />
+      {!hideNextButton && (
+        <StyledNextTooltip content={i18n.pagination.nextButtonLabel}>
+          <NavButton
+            aria-label={i18n.pagination.nextButtonLabel}
+            variant={ButtonVariant.link}
+            color={ButtonColor.secondary}
+            disabled={selectedPage >= count ? true : disabled}
```
silvalaura, line 179: Same note as above but `disabledNextTooltip`

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,190 @@
...
+            disabled={selectedPage < 2 ? true : disabled}
+            icon={<ArrowBackIcon />}
+            isInverse={isInverse}
+            theme={theme}
+            onClick={handlePrev}
+            shape={ButtonShape.fill}
+          />
+        </StyledPrevTooltip>
+      )}
+      <Spacer size={14} />
```
silvalaura, line 143: If `hidePreviousButton` is true, should we have this Spacer? or should we wrap this as part of that logic?

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -0,0 +1,190 @@
...
+        data-testid={`${testId}-select`}
+        disabled={disabled}
+        fieldId={id}
+        isInverse={isInverse}
+        onChange={handleChange}
+        value={selectedPage}
+      >
+        {Array.from({ length: count }, (_, i) => (
+          <option
+            data-testid={`${testId}-option`}
```
silvalaura, line 156: This needs to be unique:
```suggestion
            data-testid={`${testId}-option-${i}`}
```

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
silvalaura, line 1: I'm not sure how common of a use case this is, but I wonder if we need to account for the possibility of the `count` changing. I am doing it on Storybook and that's how I ran into this, but I'm thinking that perhaps API data will load and may need to update the count of the component. If so, I think we need to reset `selectedPage` anytime the value of count changes 🤔  we can talk through this one, I'm just seeing the `selectedPage` get out of sync in some edge cases.

packages/react-magma-dom/src/components/Pagination/Pagination.tsx
ccedrone: Should this be renamed in the files as well?

website/react-magma-docs/src/pages/api/pagination.mdx
```diff
@@ -28,6 +28,19 @@ export function Example() {
...
+
+```typescript
+import React from 'react';
+import { Pagination } from 'react-magma-dom';
+
+export function Example() {
+  return <Pagination count={4} type="simple" />;
```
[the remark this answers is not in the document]
ccedrone: Don't we just want `type="simple"` instead?

packages/react-magma-dom/src/components/Pagination/SimplePagination.test.js
ccedrone: 🕺

website/react-magma-docs/src/pages/api/pagination.mdx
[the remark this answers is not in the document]
silvalaura: It "works", but really the `type` prop expects an Enum, so `PaginationType.simple` is preferred. Here's a similar example where you can see CodeSandbox complain: 
image

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
ccedrone: That's a good point, moved it inside the conditional!

packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
ccedrone: Accepted!

silvalaura (commented):
A couple more notes:
- MUI supports `count={0}`, so I do think it's important we do that (0 vs 1 vs 10 below) because our classic type does.
image

- In the [Pagination Ranges](https://mui.com/material-ui/react-pagination/#pagination-ranges) example, it shows that behavior I was mentioning about "resetting" the component when the count changes: the defaultPage is 2, I clicked through to page 6, then changed the count and that reset the selectedPage to 2.
chrome-capture-2023-2-21 (1)

silvalaura (commented):
Looks great! 

One final note, I noticed this in the docs: 
image

This is because `PaginationVariant` is not exported in `react-magma/packages/react-magma-dom/src/index.ts`.


_Bonus but not required:_
I really like this example because it "proves" how it works/how the component could be hooked up in a real application. Maybe it's worth adding that functionality to the simple example?
image

silvalaura:
This PR is more than a draft, I think it's ready to be set for review ;) Also, don't forget to add the description, screenshots and testing notes.

katieweinstein (approved):

silvalaura (approved):

orion-cengage (approved):

PR #1046, opened 2023-04-18
Fix/simple pagination

ccedrone:
Issue: # [991](https://github.com/cengage/react-magma/issues/991)

## What I did
- Fixed spacing between buttons.
- Fixed VoiceOver to properly read Page X of X selected.
- Fixed Tooltip to display inverse colors when isInverse is enabled on Pagination.

stuff

## Checklist 
- [x] NA
- [x] NA

## How to test

- Confirm proper spacing of 16 pixels between navigations buttons and surrounding elements on either side.
- Confirm Tooltip displays appropriate inverse color when isInverse is enabled on Pagination
- Confirm VoiceOver properly reads Page X of X selected when keyboard navigation tabs through the select and then also when a selection is made.


packages/react-magma-dom/src/components/Pagination/SimplePagination.tsx
```diff
@@ -119,6 +118,8 @@ export const SimplePagination = React.forwardRef<
         }`;
   }
 
+  const pageAriaLabel = `Page number ${selectedPage} ${paginationLabel()} selected`;
```
silvalaura, line 121: Just tested and this is read by VoiceOver **beautifully**! The only thing is that we should i18n `Page number` and `selected` in case this is ever used in another language.

silvalaura (approved):
🕺

PR #1050, opened 2023-04-21
fix(CharCounter): Update calculation for character length on rerender

silvalaura:
_Issue: none_

## What I did

Update Input and Textarea to update the character length when the default value changes. This way, when the component with a character counter gets re-rendered with a different default value, the count will be up to date.

## Checklist 
- [n/a] I have made corresponding changes to the documentation

## How to test

Test the following for both the Input and Textarea:
- Confirm that on Storybook, if the default value is changed, the character counter is updating consistently
- Try to recreate the reported use case (https://screenpal.com/watch/c0fqouVzjFj) on CodeSandbox and confirm this would no longer be an issue:
Test changing `hi` to `hiiiiii`
```
export function Example() {
  return (
    <Input
      maxLength={4}
      value={"hi"}
      labelText="Initial Value Character Counter"
    />
  );
}
```

### DIFF
--- a/.changeset/charcounter-initialvalue.md
+++ b/.changeset/charcounter-initialvalue.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(CharCounter): Update calculation for character length on rerender.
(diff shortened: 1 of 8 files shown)


ccedrone (approved):

ccedrone:
Verified that `value` properly updates the count when updated: Working on https://next--upbeat-sinoussi-f675aa.netlify.app/

Issue #1033, opened 2023-03-20
Spike: Investigate Charts


silvalaura:
Let's do a deep dive into our current charts offering.
- Why is it built on top of Victory?
- Is it a viable option to build future charts using [Carbon visualization](https://carbondesignsystem.com/data-visualization/simple-charts#bar-(simple))? Can we reskin them? Any legal reasons why we couldn't? 
- What other options are there for accessible data visualization libraries?

I happened to see that WebAssign release-planning-tool [repo](https://stash.cengage.com/projects/WA/repos/release-planning-tool/browse/original/WebAssignReleasePlanningTool/src/main/resources/app/package.json#11) has a reference to charts. We can reach out to Alex T from WebAssign if we have questions.

silvalaura:
The big plus to Carbon is that they turn the data into a table by default. None of the [other popular React chart libraries](https://themeselection.com/react-chart-library/) do this. I think that may be the easiest way to guarantee the information is fully accessible, and for that reason, I would lean towards continuing down this path. In addition, Carbon is an open source library, so I don't see any legal reason why we couldn't use it.

As far as the ability to reskin Carbon, it's possible.
- They offer 4 themes: `white`, `g9`, `g10`, `g100`. From there, you can also customize the colors of the data in each chart. 
- We would have to add some custom CSS styling to override the font

[CodeSandbox example](https://codesandbox.io/s/magma-carbon-charts-7m11z8?file=/src/index.js:2780-2781)

We have to decide whether we want to wrap Carbon inside `@react-magma/charts` OR document our recommendation for how to use them in conjunction with Magma
Pros:
- Easier out of the box for other developers (no customization needed)

Cons:
- We'd have to release a new version every time Carbon does to keep up with the latest. AKA more work for us.
- We'd have to deprecate the existing Magma Charts and replace it with this. This shouldn't be a huge deal as _maybe_ one team is utilizing charts.

silvalaura:
- Completed investigation on 4/26/2023. Team has decided to move forward with this approach. 
- Draft pull request with work so far: https://github.com/cengage/react-magma/pull/1051
- Opened the following follow up stories where design requirements will be outlined and the charts will be fully documented:
1. #1052 
2. #1053 
3. #1054 
4. #10550
5. #1056

Some screenshots:


![Image](https://user-images.githubusercontent.com/91160746/234634576-81d6ca0e-0d43-4d0f-a400-5adf265f498f.png)


[...]

PR #1051, opened 2023-04-25
feat(CarbonCharts): Support Carbon Charts

silvalaura:
Issue: #1052 #1033 #1278 #1053 

## What I did

Bringing in _[Carbon Charts](https://charts.carbondesignsystem.com/react/?path=/story/simple-charts-area-simple--area-time-series-natural-curve)_ into @react-magma/charts. 

image
image
image

## Checklist 

## How to test

Navigate to Carbon Charts in the docs site or storybook


packages/charts/src/components/CarbonChart/CarbonChart.stories.tsx
silvalaura, line 1: Can we create a story for each of the graph charts the ticket calls out? Simple, Grouped, Floating, Lollipop

packages/charts/src/components/CarbonChart/CarbonChart.test.js
```diff
@@ -0,0 +1,44 @@
...
+    value: 0,
+  },
+  {
+    group: 'Dataset 2',
+    date: '2019-01-01T05:00:00.000Z',
+    value: 47263,
+  },
+];
+
+describe.skip('CarbonChart', () => {
```
silvalaura, line 19: What happens if we remove this skip?

packages/charts/src/components/CarbonChart/CarbonChart.tsx
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  .cds--modal-container{
+    border-radius:${props => props.theme.borderRadius};
+  }
+
+  p, div, text, .cds--cc--axes g.axis .axis-title,
+  .cds--cc--title p.title,
+  .cds--cc--axes g.axis g.tick text {
+    font-family: 'Work Sans', Helvetica, sans-serif !important;
```
silvalaura, line 55: Could we update all these values to use our theme variables? Font size/family, margin, etc etc

packages/react-magma-dom/src/theme/magma.ts
```diff
@@ -756,6 +757,36 @@ export const magma = {
...
+    '#00A393',
+    '#005F96',
+    '#8F0033',
+    '#136A2D',
+    '#B84900',
+    '#711E6E',
+    '#005249',
+  ],
+
+  charColorsInverse: [
```
silvalaura, line 775:
```suggestion
  chartColorsInverse: [
```

packages/charts/src/components/CarbonChart/CarbonChart.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+      const allGroups = dataSet.map(item => {
+        return 'group' in item ? item['group'] : null;
+      });
+      const uniqueGroups = allGroups.filter(
+        (g, index) => allGroups.indexOf(g) === index
+      );
+
+      uniqueGroups.forEach((group, i) => {
+        if (uniqueGroups.length <= theme.iterableColors.length) {
```
silvalaura, line 185: This should be using `chartColors` and `chartColorsInverse` instead of `iterableColors`

packages/react-magma-dom/src/theme/magma.ts
silvalaura, line 1: The two new values added need to be added to the ThemeInterface

packages/charts/src/components/CarbonChart/CarbonChart.test.js
ccedrone: 'Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.'

silvalaura: Sounds like a legitimate thing we should address

packages/charts/src/components/CarbonChart/CarbonChart.tsx
```diff
@@ -0,0 +1,215 @@
...
+  ThemeContext,
+  useIsInverse,
+  styled,
+} from 'react-magma-dom';
+
+import {
+  SimpleBarChart,
+  LollipopChart,
+  GroupedBarChart,
+} from '@carbon/charts-react';
```
silvalaura, line 14: Missing FloatingBar

packages/charts/src/components/CarbonChart/CarbonChart.tsx
```diff
@@ -0,0 +1,215 @@
...
+      });
+      const uniqueGroups = allGroups.filter(
+        (g, index) => allGroups.indexOf(g) === index
+      );
+
+      uniqueGroups.forEach((group, i) => {
+        if (isInverse) {
+          return (scaleColorsObj[group || 'null'] =
+            theme.chartColorsInverse[i]);
+        } else if (uniqueGroups.length <= theme.chartColors.length) {
```
silvalaura, line 184: We need to check if `isInverse` is true, use `chartColorsInverse` else `chartColors`

packages/react-magma-dom/src/theme/magma.ts
```diff
@@ -291,6 +291,8 @@ export interface ThemeInterface {
   headingFont: string;
   iconSizes: IconSizes;
   iterableColors: string[];
+  chartColors?: string[];
+  chartInverseColors?: string[];
```
silvalaura, line 295: Line 777 calls it `chartColorsInverse`. Don't care which name we use, but it has to be the same :)

packages/charts/src/components/CarbonChart/CarbonChart.tsx
ccedrone: Floating uses `SimpleBarChart`

orion-cengage:
Design feedback:

- Focus color when tabbing through legend is still incorrect. Needs to match the light blue we use on other elements.
image

- Looks like we still need a border on the checkboxes in the legend when they're turned off. Otherwise they disappear from view.  Hopefully we can have the border only when they're turned off?
image

Please use #454545 in light mode.
image

Please use #FFFFFF in dark mode.
image

orion-cengage:
Due to the update to the checkboxes, it looks like it introduced some additional styling issues.

- In light mode, the active checkboxes are now visually smaller and missing the 2px border radius, but they're fine in dark mode.
- This becomes easier to see when some checkboxes are active and others aren't.
- The checkboxes in dark mode are still missing the white outline asked for in the comment above.

image
image
image
image

orion-cengage:
Everything is looking good at this point. Can't promise there won't be tweaks down the road. But consider this "Approved" without actually approving the PR.

orion-cengage (changes requested):
Tasks after a11y review:

Full doc: 
https://cengageo365-my.sharepoint.com/:w:/g/personal/orion_johnson_cengage_com/ETrRXWRBLKRJoJIZEtFq3yoBujnPQ5vq4WINa71Z_7vT_g?e=fYnCos

- [ ] On charts where the bars touch, we need to make sure there is a pixel of separation between them. This somehow got lost between Carbon and our styles. 
image
image

- [ ] I want to know if the data shown in the table is being sorted or if it simply reflects how the data was entered. Use "Vertical grouped bar (time series)" as an example.
image

- [ ] Increase size of circles on the Lollipop chart from 8px to 24px. We need to keep in mind for later charts that interactive points need to be at least 24px.
- [ ] Table title - Stop truncation and let it wrap
- [ ] Can we potentially add an aria label to the main top level svg? Like “aria-label=”Interactive chart”

Data Table

- [ ] In the table modal, change title to font-weight: 600
- [ ] Give header white background, bottom border of #D4D4D4, and remove bottom margin
- [ ] Give footer white background, and top border of #D4D4D4
- [ ] Reduce the height of table headers and rows from 3rem to 2.5rem 
- [ ] Can we stop the table header from being sticky?
- [ ] Can we stop the modal header and footer from being sticky? Or maybe just at a certain breakpoint? 
image

orion-cengage (approved):

ccedrone (approved):

Issue #1037, opened 2023-03-31
Input/Textarea: maxLength prop overrides native HTML


silvalaura:
**Describe the bug**
Between versions 3.1 and 3.2, the behavior of maxLength was changed for Input and TextArea.  This added some feedback below the element about characters remaining, but overwrote the existing functionality of the maxLength property of the native html elements.  Also, it makes error handling somewhat complex for elements like these.  The consuming platform must provide error handling for everything other than “Character limit exceeded”.   I like the character limit feedback, but it comes at the cost of modifying native html element behavior and nuanced error handling.
Is it crazy to propose changing the prop name to charLimit, and retain the native behavior associated with maxLength ?  The error for character limit error is something easily lived with.

**Expected behavior**
- We should support both the native `maxLength` property and the character counter. 
- This change of changing the prop should be backwards compatible. The old prop name needs to be marked as deprecated, with the explanation that moving forward it will work as the native attribute
- Users should be able to use `maxLength` without the character counter
- Users should be able to use both `maxLength` AND the character counter

**Other comments**
We need to figure out the best approach and consider that this would be a breaking change.

silvalaura:
**Testing Notes:**

CodeSandbox with all test cases tested: https://codesandbox.io/s/testing-1037-maxlength-prop-qmmy0l

1. Input: maxCount ✅ 
2. Input: maxCount and hasCharacterCounter={true} ✅ 
3. Input: maxCount and hasCharacterCounter={false} ⚠️ **when the maxcount is reached, the input gets the red error border. maybe this is okay because they specifically set a maxCount, but then maybe we should be showing the # of characters left??** 

4. Input: maxLength ⚠️ **it starts off with the number of characters allowed bolded, when it should not be**
5. Input: maxLength and hasCharacterCounter={true} ✅ 
6. Input: maxLength and hasCharacterCounter={false} ✅ 

7. Input: maxCount + maxLength ⚠️ **it starts off with the number of characters allowed bolded, when it should not be**
8. Input: maxCount + maxLength and hasCharacterCounter={true} ✅ 
9. Input: maxCount + maxLength and hasCharacterCounter={false} ✅ 

10. Textarea: maxCount ✅ 
11. Textarea: maxCount and hasCharacterCounter={true} ✅ 
[...]

PR #1061, opened 2023-05-09
fix(CharacterCounter): Replace `maxLength` prop with `maxCount` and add `hasCharacterCounter` prop

ccedrone:
Issue: # [1037](https://github.com/cengage/react-magma/issues/1037)

## What I did
- Added two new props of 'maxCount' replacing 'maxLength' for enabling the Character Counter and 'hasCharacterCounter'. If 'hasCharacterCounter' is set to false, it hides the Character Counter.

## Checklist 
- [x] NA

## How to test

Test both the regular **Input** and **Textarea** components:
- Confirm the Character Counter appears when `maxCount` is set.
- Confirm `maxLength` still works as a Character Counter.
- Confirm `maxLength` works as a native maxlength when `hasCharacterCounter` is set to false.

### DIFF
--- a/.changeset/fix-characterCounterProp.md
+++ b/.changeset/fix-characterCounterProp.md
@@ -2,6 +2,8 @@
 'react-magma-dom': patch
 ---
 
-fix(Character Counter): Added two new props 'maxCount' and 'hasCharacterCounter'. The prop of 'maxCount' replaces 'maxLength'. Now 'maxLength' works as the native HTML attribute.
+fix(Character Counter): Two new props have been added. The `hasCharacterCounter` prop which defaults to `true` and the `maxCount` prop which replaces `maxLength` and enables the Character Counter.
 
-Note: This update is a breaking change and all uses of the Character Counter need to update to the 'maxCount' prop name.
+In the interim if an input needs a native `maxlength` and not a Character Counter, set `hasCharacterCounter={false}` and then use `maxLength`.
+
+Please note that in the meantime, `maxLength` is still supported but will need to be changed to `maxCount` as future releases will remove `maxLength` and `hasCharacterCounter`.
--- a/.changeset/fix-characterCounterProp.md
+++ b/.changeset/fix-characterCounterProp.md
@@ -1,0 +1,7 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Character Counter): Added two new props 'maxCount' and 'hasCharacterCounter'. The prop of 'maxCount' replaces 'maxLength'. Now 'maxLength' works as the native HTML attribute.
+
+Note: This update is a breaking change and all uses of the Character Counter need to update to the 'maxCount' prop name.
--- a/packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
+++ b/packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
@@ -34,6 +34,11 @@ export default {
         type: 'text',
       },
     },
+    hasCharacterCounter: {
+      control: {
+        type: 'boolean',
+      },
+    },
     helperMessage: {
       control: {
         type: 'text',
@@ -55,6 +60,11 @@ export default {
         options: LabelPosition,
       },
     },
+    maxLength: {
+      control: {
+        type: 'number',
+      },
+    },
     maxCount: {
       control: {
         type: 'number',
@@ -70,8 +80,8 @@ export default {
 
 export const Default = Template.bind({});
 Default.args = {
-  maxCount: 4,
   isInverse: false,
+  maxLength: 4,
 };
 
 export const WithChildren = args => {
--- a/packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
+++ b/packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
@@ -55,12 +55,7 @@ export default {
         options: LabelPosition,
       },
     },
-    maxlength: {
-      control: {
-        type: 'text',
-      },
-    },
-    maxLength: {
+    maxCount: {
       control: {
         type: 'number',
       },
@@ -75,7 +70,7 @@ export default {
 
 export const Default = Template.bind({});
 Default.args = {
-  maxLength: 4,
+  maxCount: 4,
   isInverse: false,
 };
 
--- a/packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
+++ b/packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
@@ -81,7 +81,7 @@ export default {
 export const Default = Template.bind({});
 Default.args = {
   isInverse: false,
-  maxLength: 4,
+  maxCount: 4,
 };
 
 export const WithChildren = args => {
--- a/packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
+++ b/packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
@@ -14,11 +14,7 @@ const labelText = 'Character Counter';
 
 const Template: Story<CharacterCounterProps> = args => (
   <>
-    <Input
-      {...args}
-      testId="test-this-id"
-      labelText={labelText}
-    />
+    <Input {...args} testId="test-this-id" labelText={labelText} />
   </>
 );
 
@@ -59,6 +55,11 @@ export default {
         options: LabelPosition,
       },
     },
+    maxlength: {
+      control: {
+        type: 'text',
+      },
+    },
     maxLength: {
       control: {
         type: 'number',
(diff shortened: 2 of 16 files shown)


silvalaura (commented):
Can you add testing steps? I can't seem to get native`maxLength` to work. Unit tests would also be great.

silvalaura (commented):
This is working, but lets make sure we document this very well:
- The docs site should mention the new prop `hasCharacterCounter` and explain it's temporary and how it should be used
- We should also document maxLength vs maxCount to avoid future confusion
- We are missing unit tests that test the `maxLength` (native) property. We should test that it works as expected alone, with/without `hasCharacterCounter`, with/without `maxCount`
- Can we write down all the cases to be tested in the PR description? what are all the prop combinations and what not that should be verified. Also, lets fill out all the sections in the description-- this helps if we ever need to go back to this PR and remember what exactly it affects


Question / Discussion / Not a request, just thinking out loud:
- if maxCount is present, character counter should show up so hasCharacterCounter should automatically be true / be ignored if false

.changeset/fix-characterCounterProp.md
```diff
@@ -0,0 +1,7 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Character Counter): Added two new props 'maxCount' and 'hasCharacterCounter'. The prop of 'maxCount' replaces 'maxLength'. Now 'maxLength' works as the native HTML attribute.
+
+Note: This update is a breaking change and all uses of the Character Counter need to update to the 'maxCount' prop name.
```
silvalaura, line 7: The way you've built this, it is _not_ a breaking change (on purpose). I think it would be clearer to say something along these lines:

fix(CharacterCounter): 'maxLength' prop now works as the native HTML attribute if new `hasCharacterCounter` is set to `false`. Character Counter functionality is now tied to new `maxCount` prop. `hasCharacterCounter` is a temporary prop until our next major version where the use of `maxLength` for character counter will be deprecated.

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.stories.tsx
```diff
@@ -64,18 +65,23 @@ export default {
...
+      },
+    },
   },
 } as Meta;
 
 export const Default = Template.bind({});
 Default.args = {
-  maxLength: 4,
   isInverse: false,
+  maxLength: 4,
```
silvalaura, line 84: Should this be `maxCount`?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.test.js
silvalaura, line 1: Need new tests for `maxLength` as the native prop and `hasCharacterCounter`. What are all the possible combinations of these props, and how can we ensure they are behaving as expected through unit tests?

packages/react-magma-dom/src/components/CharacterCounter/CharacterCounter.tsx
```diff
@@ -6,6 +6,12 @@ import { I18nContext } from '../../i18n';
 
 export interface CharacterCounterProps
   extends React.HTMLAttributes<HTMLDivElement> {
+  /** 
+   * Enables Character Counter by default. When set to false, the default HTML attribute of 'maxlength' will work.
```
silvalaura, line 10:
```suggestion
   * Enables Character Counter by default.
   * When set to false in combination with 'maxLength', the default behavior of the HTML attribute will appear.
```

packages/react-magma-dom/src/components/FormFieldContainer/FormFieldContainer.tsx
```diff
@@ -33,6 +33,12 @@ export interface FormFieldContainerBaseProps {
    * ID of the form field.  Also used in the descrption ID.
    */
   fieldId: string;
+  /** 
+   * Enables Character Counter by default. When set to false, the default HTML attribute of 'maxlength' will work. 
+   * Note: This is a temporary prop and will be removed in future releases.
```
silvalaura, line 38: Update with same message from CharacterCounter.tsx

packages/react-magma-dom/src/components/Textarea/Textarea.test.js
silvalaura, line 1: Same here, lets add unit tests to Textarea around `maxLength` vs `maxCount`

packages/react-magma-dom/src/components/InputBase/index.tsx
```diff
@@ -37,6 +37,12 @@ export interface InputBaseProps
    * Style properties for the component container element
    */
   containerStyle?: React.CSSProperties;
+  /** 
+   * Enables Character Counter by default. When set to false, the default HTML attribute of 'maxlength' will work.
```
silvalaura, line 41: Update with same message from CharacterCounter.tsx

website/react-magma-docs/src/pages/api/character-counter.mdx
```diff
@@ -40,15 +40,28 @@ import { Input } from 'react-magma-dom';
...
 }
 ```
 
+## New Character Counter props
+
+Two new props have been added. The `hasCharacterCounter` prop which defaults to `true` and the `maxCount` prop which replaces `maxLength` and enables the Character Counter.
+
+In the interim if an input needs a native `maxlength` and not a Character Counter, set `hasCharacterCounter={false}` and then use `maxLength`.
+
+Please note that in the meantime, `maxLength` is still supported but will need to be changed to `maxCount` as future releases will remove `maxLength` and `hasCharacterCounter`.
```
silvalaura, line 54: The changeset is the place to use this kind of voice, where we can explain things to developers in the form of what changed/what's been added (that's where they'll look for it). In our docs site, I think we need to be a bit more direct.

```suggestion
## maxLength vs maxCount

The `maxLength` prop temporarily controlled the Character Counter, but moving forward should exclusively be used for the HTML native behavior of preventing users of typing beyond the set maximum length.

To enable the Character Counter, use `maxCount` (preferred) or **both** `maxLength` and `hasCharacterCounter` (to be deprecated in future versions).
```

silvalaura (commented):
This change affects CharacterCounter, Input, and Textarea. Let's make sure this is outlined in the testing notes; the more clear the testing instructions are in the ticket, the more confident we'll feel that we thoroughly tested both manually and programmatically.

packages/react-magma-dom/src/components/FormFieldContainer/FormFieldContainer.tsx
```diff
@@ -188,12 +204,14 @@ export const FormFieldContainer = React.forwardRef<
...
-          {typeof maxLength === 'number' && (
+          {typeof countProps === 'number' && hasCharacterCounter && (
             <CharacterCounter
+              hasCharacterCounter={hasCharacterCounter}
               id={descriptionId}
               inputLength={inputLength}
               isInverse={isInverse}
-              maxLength={maxLength}
+              maxCount={countProps}
+              maxLength={countProps}
```
silvalaura, line 214: Just want to triple check, should this be as follows?
```suggestion
              maxCount={maxCount}
              maxLength={maxLength}
```

ccedrone: Either seems to work, shall I make the amendment?

silvalaura: If both work, I think we should update it. It'll be safer and easier to get rid of the props that are temporary when it's time.

ccedrone: Actually `countProps` should remain, it breaks and is undefined in the tests otherwise.

silvalaura: Here's what I think:

1. In FormFieldContainer, we should pass maxCount to maxCount and maxLength to maxLength. This will ensure that the CharacterCounter component receives the appropriate values.
2. Then, in the CharacterCounter component, we do the extra handling to figure out which value to use. This means using maxCount when that has a number value, or maxLength as the fallback. Something like
```
  // Temporary while we need to support both maxCount and maxLength.
  const maxCharacters = typeof maxCount === 'number' ? maxCount : maxLength;
```
3. Then, in the rest of the CharacterCounter component we use the `maxCharacters` variable to check all the logic. This of course can be called whatever you want.

I think the benefit of doing this is that there's less risk of the values being accidentally interchanged. How it's working right now is that FormFieldContainer is sending both maxCount and maxLength the same value. Technically it's working, but could potentially cause problems. 


There are multiple ways of doing this, this is just a suggestion. What do you think?

silvalaura (approved):
Code looks good!

Can we just update the title of the PR to include the changes we did, something that mentions the exact changes made? I find this helpful in case in the future, we need to track down why we did something, it makes it easier to look through the branch history and find pull requests. Example:
```
fix(CharacterCounter): Replace `maxLength` prop with `maxCount` and add `hasCharacterCounter` prop
```

Also, can we add to the testing notes that we should test this with both the Input AND Textarea components? I know it's obvious but I don't want to forget when I get to QA it.

PR #1067, opened 2023-05-18
Fix border radius on table pagination

noahwebs:
Issue: #

## What I did
The TablePagination component does not use rounded corners by default even though the Table component does. This can cause a visual problem when the Table and TablePagination components are placed inside a Card with the intention of giving the entire display a border with rounded corners since the TablePagination component will spill over the Card boundary. 

Before:
Screenshot 2023-05-18 at 12 23 48 PM

After: 
Screenshot 2023-05-18 at 12 24 32 PM

## Checklist 

### DIFF
--- a/packages/react-magma-dom/src/components/Table/Table.stories.tsx
+++ b/packages/react-magma-dom/src/components/Table/Table.stories.tsx
@@ -300,7 +300,7 @@ export const ControlledPagination = args => {
   );
 
   return (
-    <Card isInverse={args.isInverse}>
+    <Card>
       <Table {...args}>
         <TableHead>
           <TableRow>
--- a/packages/react-magma-dom/src/components/Table/TablePagination.tsx
+++ b/packages/react-magma-dom/src/components/Table/TablePagination.tsx
@@ -108,6 +108,7 @@ const StyledContainer = styled.div<{
   display: flex;
   justify-content: flex-end;
   padding: ${props => props.theme.spaceScale.spacing02};
+  border-radius: 0 0 ${props => props.theme.borderRadius} ${props => props.theme.borderRadius};
 `;
 
 const PageCount = styled(Label) <{ theme: ThemeInterface }>`


packages/react-magma-dom/src/components/Table/Table.stories.tsx
```diff
@@ -300,7 +300,7 @@ export const ControlledPagination = args => {
   );
 
   return (
-    <Card isInverse={args.isInverse}>
```
noahwebs, line 303: I removed this because it seemed like cruft that us unintentionally redundant.

packages/react-magma-dom/src/components/Table/TablePagination.tsx
```diff
@@ -108,6 +108,7 @@ const StyledContainer = styled.div<{
   display: flex;
   justify-content: flex-end;
   padding: ${props => props.theme.spaceScale.spacing02};
+  border-radius: 0 0 ${props => props.theme.borderRadius} ${props => props.theme.borderRadius};
```
noahwebs, line 111: Do we need to incorporate a similar "hasSquareCorners" property that toggles between this and the previous behavior? I am a bit wary of bluntly making a style change that can't be reverted.

silvalaura: I spoke to our designer @orion-cengage and we agreed that we _should_ add a `hasSquareCorners` to match the prop for tables

packages/react-magma-dom/src/components/Table/Table.stories.tsx
silvalaura: We have this for testing purposes, so let's just keep it as is.

silvalaura (changes requested):

.changeset/fix-tablePaginationBorderRadius.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(TablePagination): TablePagination component now respects theme border radius on bottom corners
```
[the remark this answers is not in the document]
silvalaura: I think we'd prefer to not have a breaking change, so I say let's keep this as a patch change and make `hasSquareCorners` have a default of `true` (despite it not being consistent with the variable by the same name that `Table` has).

packages/react-magma-dom/src/components/Table/TablePagination.tsx
```diff
@@ -108,6 +114,10 @@ const StyledContainer = styled.div<{
   display: flex;
   justify-content: flex-end;
   padding: ${props => props.theme.spaceScale.spacing02};
+  border-radius: ${props => 
+    props.hasSquareCorners 
+    ? "0;"
+    : `0 0 ${props.theme.borderRadius} ${props.theme.borderRadius};`}
```
silvalaura, line 120: nitpick but for consistency:

```suggestion
  border-radius: ${props => 
    props.hasSquareCorners 
    ? "0"
    : `0 0 ${props.theme.borderRadius} ${props.theme.borderRadius}`};
```

Issue #1072, opened 2023-05-24
TablePagination Hide "Rows Per Page" Selector


noahwebs:
**Is your feature request related to a problem? Please describe.**
Sometimes an application will need a table whose rows are backed by very data rich objects. In this case, it may be a reasonable approach to limit the number of rows loaded at once to a very low threshold. Since the "Rows Per Page" selector on the TablePagination component cannot be hidden, those who want to use it are forced to support loading 100 rows at a time. 

There are probably other reasons for wanting to limit the number of rows displayed at once as well (e.g. page real estate, less code complexity, etc.).

**Describe the solution you'd like**
The TablePagination component should take in a "hideRowsPerPageSelector" argument to allow hiding the selector component. Alternatively, this could be implicitly determined by checking if the onRowsPerPageChange function is null. 

**Describe alternatives you've considered**
Without this feature, Magma users will have to "clone and own" an alternate version of the TablePagination component, losing out on any future enhancements to the component. 

**Additional context**
N/A

PR #1076, opened 2023-05-31
feat: Allow hiding RowsPerPageController when no onRowsPerPageChange …

noahwebs:
…function passed

Issue: #1072

## What I did
I extracted the "Rows Per Page" label and selector of the TablePagination component into a local React functional component so that it could be conditionally rendered depending on whether the caller passed an `onRowsPerChange` function. If no `onRowsPerChange` function is passed, the controller should not display. 

In the course of refactoring the code I deleted the branch of code that rendered the controller differently when a `dropdownDropDirection` value was passed to the TablePagination code. Since the event handler function for the two branches was different, the alternate implementation was non-functional. To be more explicit, passing a `dropdownDropDirection` value would cause the controller to break so that you could select new "Rows Per Page" values, but the state would not propagate to the rest of the paginator component. Since the `dropdownDropDirection` property is marked as deprecated, I figured this was a safe decision and better than continuing to bring cruft forward. 

Before:
Screenshot 2023-05-31 at 12 07 43 PM

After:
Screenshot 2023-05-31 at 12 08 34 PM

## Checklist 

## How to test
- Verify that when you don't pass any function to the `onRowsPerPageChange` property of TablePagination, the label and selector are hidden.
- Verify that when you do pass a function, the label and selector are rendered.
- Verify that `rowsPerPage` is respected, even for non-standard values (i.e. 15, 25, etc.)

### DIFF
--- a/packages/react-magma-dom/src/components/Table/TablePagination.tsx
+++ b/packages/react-magma-dom/src/components/Table/TablePagination.tsx
@@ -14,7 +14,7 @@ import { useControlled } from '../../hooks/useControlled';
 import { transparentize } from 'polished';
 import { ButtonGroup, ButtonGroupAlignment } from '../ButtonGroup';
 import { NativeSelect } from '../NativeSelect';
-import { Dropdown, DropdownAlignment, DropdownButton, DropdownContent, DropdownDropDirection, DropdownMenuItem } from '../Dropdown';
+import { DropdownDropDirection } from '../Dropdown';
 
 export interface BaseTablePaginationProps
   extends React.HTMLAttributes<HTMLDivElement> {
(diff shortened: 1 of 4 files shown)


packages/react-magma-dom/src/components/Table/TablePagination.tsx
```diff
@@ -136,6 +136,55 @@ const RowsPerPageLabel = styled.span<{
   color: ${props => (props.isInverse ? props.theme.colors.neutral100 : '')};
 `;
 
+interface RowsPerPageControllerProps {
+  /**
+   * Event that fires when the number of rows per page changes
+   */
+  handleRowsPerPageChange?: (value: any) => void;
+  rowsPerPageValues: number[];
```
silvalaura, line 144: Lets keep the prop description here:
```
  /**
   * Values added to the rows per page select
   */
```

packages/react-magma-dom/src/components/Table/TablePagination.tsx
```diff
@@ -136,6 +136,55 @@ const RowsPerPageLabel = styled.span<{
...
+      >
+        {rowsPerPageItems.map((row, index) => (
+          <option
+            key={index}
+            value={row.value}
+          >
+            {row.label}
+          </option>
+        ))}
+      </NativeSelect>
```
silvalaura, line 183: >  Since the event handler function for the two branches was different, the alternate implementation was non-functional.

Both controllers use the same `handleRowsPerPageChange` function, so I'd love to hear more about the issues you were seeing around this:
- https://github.com/cengage/react-magma/pull/1076/files#diff-72059da60baba78fdc1cdf4cd0725c54fb5c508fa988047877ad22f3e5c84eb8L252
- https://github.com/cengage/react-magma/pull/1076/files#diff-72059da60baba78fdc1cdf4cd0725c54fb5c508fa988047877ad22f3e5c84eb8L243

I would really prefer to keep both otherwise this is a breaking change, and we like to limit those as much as possible to make it easier for our adopters to stay up to date with the latest version.

noahwebs: Perhaps I'm not understanding how the alternate codeflow is supposed to work, but if you go to the Table.stories.tsx file and make this change (on a branch without my changes) in either the UncontrolledPagination or ControlledPagination stories:
Screenshot 2023-07-06 at 10 31 30 AM

Then you get a selector where you can make the selection, but it doesn't update at all:
Screenshot 2023-07-06 at 10 31 12 AM


Effectively, passing a value for dropdownDropDirection completely breaks rows per page selection. 
DropdownMenuItem's onClick function has this signature: `onClick?: (value?: string | number) => void;`
NativeSelect's onChange function has this signature:` onChange?: (event: ChangeEvent) => void;`


handleRowsPerPageChange is written to accept an `event` parameter, not a `value` parameter, so when DropdownMenuItem calls it, it breaks and fails to call setRowsPerPageState or onRowsPerPageChange appropriately: 

Screenshot 2023-07-06 at 10 56 54 AM

Screenshot 2023-07-06 at 10 55 35 AM

So this is a bug that no one has complained about because on one is passing the dropdownDropDirection parameter.

silvalaura: I see what you're saying. In that case, I agree with you. Lets get rid of `dropdownDropDirection` and we'll call this a `minor` change-- it slightly breaks the rules, but like you said, no one has complained in months about this and using NativeSelect is ultimately the best solution. Thanks for digging into this!

silvalaura (approved):
Looks good. I will merge this shortly.

orion-cengage:
I verified this change looks good on the Table component. Should we also do this to the Datagrid component at the same time, or do we need to do that separately?

PR #1128, opened 2023-09-12
fix(input): Update input height to 40px/56px

silvalaura:
Issue: none

## What I did

Update Input height to match button sizes: 40px (instead of 42px) and 56px (instead of 58px)

## Screenshots:

Before:
image
image

After:
image

image

## Checklist 

## How to test

- Test all components that use inputs and ensure nothing is broken. 
- Test the 2 input sizes

### DIFF
--- a/.changeset/inputHeight.md
+++ b/.changeset/inputHeight.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Input): Update Input height to match button sizes: 40px (instead of 42px) and 56px (instead of 58px)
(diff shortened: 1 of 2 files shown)


ccedrone (approved):

orion-cengage:
Found some issues:

- Input with children, two icons, they look overlapped on the right side I think
- image

- Help link top, I think the help icon should be showing up above the input?
- image

- The border for Native Select looks partially missing?
- image

silvalaura:
@orion-cengage Thanks for finding that! I have opened a PR to address that: https://github.com/cengage/react-magma/pull/1132

**Notes:**
1. Updated NativeSelect so that the border is fully visible (had to make the component 38px + 2px of borders for a total of 40px
2. The Input story `with children` was confusing but nothing was broken:
- For example, we support icons for top, right and left (see  Icon Positions story). However, if the user wants to make the button interactive (like to use a tooltip like in the examples), we only support that for top and right. 
- If the user wants to use 2 icons, it can only be done with one to the left and one to the right.
- I created a new story for `with two icons` and made some modifications to the `with children` story to clear this up.

orion-cengage:
As far as I can tell, this looks good to go.

PR #1132, opened 2023-09-18
fix(NativeSelect): Update NativeSelect height to match button sizes

silvalaura:
Issue: none

## What I did

1. Updated NativeSelect so that the border is fully visible (had to make the component 38px + 2px of borders for a total of 40px
2. The Input story `with children` was confusing but nothing was broken:
- For example, we support icons for top, right and left (see  Icon Positions story). However, if the user wants to make the button interactive (like to use a tooltip like in the examples), we only support that for top and right. 
- If the user wants to use 2 icons, it can only be done with one to the left and one to the right.
- I created a new story for `with two icons` and made some modifications to the `with children` story to clear this up.

## Screenshots:
Before:
image

After:
image
image

## Checklist 
- [N/A] I have added tests that prove my fix is effective or that my feature works

## How to test
Test Native Select and other components that use input

### DIFF
--- a/.changeset/nativeSelectHeight.md
+++ b/.changeset/nativeSelectHeight.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(NativeSelect): Update NativeSelect height to match button sizes: 40px (instead of 42px)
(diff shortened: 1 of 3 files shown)


ccedrone (approved):

orion-cengage:
As far as I can tell, this looks good to go.

Issue #1088, opened 2023-06-22
PasswordInput: Show button in size large is off center


silvalaura:
**Describe the bug**
The show button in the large PasswordInput is in the wrong place

**To Reproduce**
1. Go to https://next--upbeat-sinoussi-f675aa.netlify.app/api/password-input/

**Screenshots**
![image](https://github.com/cengage/react-magma/assets/91160746/6fe57418-0840-463a-9bc5-b2575bcffa38)

**Notes**
In addition to fixing the above for large inputs, lets also polish up the show/hide button in the regular size

ccedrone:
Looks good here!

PR #1136, opened 2023-09-20
fix(PasswordInput): Update placement of Show/Hide button on Large input size

silvalaura:
Issue: #1088

## What I did

Fix placement of show / hide button in large password input. Also added input size to storybook and to the list of the component's props.

## Screenshots:

Before:
image

After:
image

## Checklist 
- [n/a] I have added tests that prove my fix is effective or that my feature works

## How to test

Test PasswordInput component in both the small and large size

### DIFF
--- a/.changeset/fix-passInputLarge.md
+++ b/.changeset/fix-passInputLarge.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(PasswordInput): Update placement of Show/Hide button on Large input size
--- a/packages/react-magma-dom/src/components/InputBase/index.tsx
+++ b/packages/react-magma-dom/src/components/InputBase/index.tsx
@@ -459,9 +459,8 @@ const PasswordButtonContainer = styled.span<{
   background-color: transparent;
   width: 0;
   transform: translate(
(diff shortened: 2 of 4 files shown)


orion-cengage (changes requested):
The original issue is fixed, but there needs to be some tweaks to polish it up.

- The size seems inconsistent with other buttons on large inputs. The perfect example of sizing, spacing, etc is the Clear button on large text inputs. 
- image
- It's 40px tall, equal distances from top, bottom and right side, and the focus ring doesn't run into the border of the input.

- One of the problems specific to the password input, is that the show/hide button changes width depending on which state it is in. We need to pick a fixed width that accommodates both text labels so it doesn't change.
- image
- image

- It appears when you change from medium to large input, all of the text in the input changes appropriately in size, but we're not changing the button. We need to either change it to the medium button variant, or just change the text style within the button.

orion-cengage:
Almost there. Just a couple little tweaks.

1. Can we change the width/max-width of the button from 56px to 64px?
2. Change width calculation on input element to "calc(100% - **64px**)"
3. The space to the right of the button isn't equal to the space above and below it. I think if we change the transform applied to the span wrapped around the button from (-60px, 8px) to (**-72px**, 8px) we'll be good.

In the end, we should end up with this:
image

orion-cengage (approved):

Issue #1118, opened 2023-08-21
Modals: Updating padding


orion-cengage:
Please update the padding on modals so it is equal on all four sides. Currently, we have a smaller amount of padding on top to account for the line-height of the title.

Viewport = 600px and above: 24px padding

Viewport = 599px and below: 16px padding

PR #1138, opened 2023-09-20
fix(Modal): Update padding so it's equal on all sides

silvalaura:
Issue: #1118

## What I did

Update the padding on modals so it is equal on all four sides (previously we had a smaller amount of padding on top to account for the line-height of the title)

## Screenshots:

image
image

## Checklist 
- [-] I have made corresponding changes to the documentation

## How to test

Confirm that for the different viewport sizes, the modal padding is the expected size

### DIFF
--- a/.changeset/update-modalPadding.md
+++ b/.changeset/update-modalPadding.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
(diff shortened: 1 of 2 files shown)


orion-cengage (approved):
Changes look great.

PR #1177, opened 2024-01-04
feat(Paragraph): New `noTopMargin` and `noBottomMargin` props that allow customization of which margin to remove.

ccedrone:
## What I did
- Added two new props `noBottomMargin` and `noTopMargin` for additional styling options.

## Screenshots:
no-b
no-t

## Checklist 

## How to test
- Got to Docs > Paragraph
- Ensure both additional examples of No Bottom Margin and No Top Margin appear appropriately

### DIFF
--- a/.changeset/feat-paragraphMargins.md
+++ b/.changeset/feat-paragraphMargins.md
@@ -2,4 +2,4 @@
 'react-magma-dom': minor
 ---
 
-feat(Paragraph): Adding more configuration for the Paragraph component margins. `noTopMargin` and `noBottomMargin` both allow a removal of the margin from one side.
+feat(Paragraph): New `noTopMargin` and `noBottomMargin` props that allow customization of which margin to remove.
--- a/.changeset/feat-paragraphMargins.md
+++ b/.changeset/feat-paragraphMargins.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': minor
+---
+
+feat(Paragraph): Adding more configuration for the Paragraph component margins. `noTopMargin` and `noBottomMargin` both allow a removal of the margin from one side.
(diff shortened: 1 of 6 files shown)


packages/react-magma-dom/src/components/Typography/index.tsx
```diff
@@ -123,7 +125,13 @@ const baseParagraphStyles = props => css`
 export const paragraphLargeStyles = props => css`
   ${baseParagraphStyles(props)}
 
-  margin: ${props.noMargins ? '0' : `${props.theme.spaceScale.spacing06} 0`};
+  margin: ${props.noMargins
```
silvalaura, line 128: I think for the logic for all these, we should update it to allow adopters to use both `noBottomMargin` and `noTopMargin`. I know it seems counterintuitive because there's already a `noMargins` prop, but otherwise it's unclear to someone using both which one takes precedence.

.changeset/feat-paragraphMargins.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': minor
+---
+
+feat(Paragraph): Adding more configuration for the Paragraph component margins. `noTopMargin` and `noBottomMargin` both allow a removal of the margin from one side.
```
silvalaura, line 5: Just a nitpick, but the changeset descriptions are what adopters will see when upgrading versions. I think we can be more direct. Instead of `configurations` we should call them `props`.

```suggestion
feat(Paragraph): New `noTopMargin` and `noBottomMargin` props that allow customization of which margin to remove.
```

Also for organization purposes, we should use the same line here as the pull request title. This would allow us to find the pull request in the history quickly and match it to that item in the change log.

silvalaura (changes requested):

silvalaura (approved):
Looks good. Please just update the PR title to match the changeset description before merging.

PR #1185, opened 2024-01-19
Patch - Release 3.7.0

ccedrone:


silvalaura:
Merged https://github.com/cengage/react-magma/pull/1189 instead of this one (almost identical, but that one had a few more updates to the changelogs)

Issue #1201, opened 2024-02-27
Bug: Issue with pagination control, rows per page, on Table and Datagrid


orion-cengage:
Issue:

Looks like this issue was introduced in v4.3.0. Everything is working as intended in v4.2.0

**Table pagination:** https://react-magma.cengage.com/version/4.3.0/api/table/#pagination

The table pagination starts at 10 rows per page. If you change that to a different value using the select, the number of rows shown updates correctly, but the select still shows as 10 selected.

https://github.com/cengage/react-magma/assets/24397282/babfa1c0-a366-4015-9dbd-13eb9183f813

**Datagrid:** https://react-magma.cengage.com/version/4.3.0/api/datagrid/

The way this is showing on Datagrid is slightly different. On the very first example in the documentation, the pagination is set to 10 rows per page, but all 25 rows in the example are already showing by default. If you change the value to 20 or some other value, the table itself updates correctly, but the select in the pagination remains set at 10, just like the regular table component.

![image](https://github.com/cengage/react-magma/assets/24397282/e6a31790-88cf-4a7a-b761-a86276181b77)

silvalaura:
PR https://github.com/cengage/react-magma/pull/1159 introduced this bug when we changed NativeSelect.

silvalaura:
This ticket correctly addresses the first issue. However, the weirdness with our DataGrid example noted above has not been addressed.

silvalaura:
Confirmed Datagrid issue has been addressed ✅

PR #1203, opened 2024-03-07
fix(NativeSelect): Fixed related simple pagination issue with selected states not updating after selection.

ccedrone:
Issue: # [1201](https://github.com/cengage/react-magma/issues/1201)

## What I did
- Added a conditional layer to prevent additional wrapping of the `NativeSelect`. This fixes the selection issue documented in the bug ticket.

## Screenshots:
Selection

## Checklist 
- [N/A] I have made corresponding changes to the documentation

## How to test

- Go to Storybook
- Go to Datagrid > default
- Within the table click on the Rows Per Page select and change from 10 to anything else
- Verify the new selection appears in the dropdown and the data in the table updates accordingly

### DIFF
--- a/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
+++ b/packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { axe } from '../../../axe-helper';
 import { magma } from '../../theme/magma';
 import { NativeSelect } from '.';
-import { render } from '@testing-library/react';
+import { render, fireEvent } from '@testing-library/react';
 import { transparentize } from 'polished';
 import { Tooltip } from '../Tooltip';
 import { IconButton } from '../IconButton';
@@ -63,6 +63,20 @@ describe('NativeSelect', () => {
       'border',
       `1px solid ${magma.colors.neutral500}`
     );
+  });
+
+  it('should update the selected option', () => {
+    const { getByTestId, getAllByTestId } = render(
+      <NativeSelect testId={testId}>
+        <option data-testid={`${testId}-option-1`}>1</option>
+        <option data-testid={`${testId}-option-2`}>2</option>
+        <option data-testid={`${testId}-option-3`}>3</option>
+      </NativeSelect>
+    );
+    fireEvent.change(getByTestId(testId), { target: { value: 2 } });
+    expect(getByTestId(`${testId}-option-1`).selected).toBeFalsy();
+    expect(getByTestId(`${testId}-option-2`).selected).toBeTruthy();
+    expect(getByTestId(`${testId}-option-3`).selected).toBeFalsy();
   });
 
   it('should render a default inverse border', () => {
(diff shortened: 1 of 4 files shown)


silvalaura (commented):
Can we add some unit tests for this fix? Maybe we should have some on Table or Datagrid so we can catch any future issues. And/or we should add some to NativeSelect itself.

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
```diff
@@ -159,5 +173,36 @@ describe('NativeSelect', () => {
         'flex'
       );
     });
+
+    it(`Shouldn't display an additional wrapper when additionalContent is unset'`, () => {
```
silvalaura, line 177: I think this would be better as 2 separate tests, no rerenders. 
1. When additionalContent is present, a wrapper is present
2. When there's no additionalContent present, a wrapper is NOT present

packages/react-magma-dom/src/components/NativeSelect/NativeSelect.test.js
```diff
@@ -65,6 +65,20 @@ describe('NativeSelect', () => {
...
+      <NativeSelect testId={testId}>
+        <option data-testid={`${testId}-option-1`}>1</option>
+        <option data-testid={`${testId}-option-2`}>2</option>
+        <option data-testid={`${testId}-option-3`}>3</option>
+      </NativeSelect>
+    );
+    fireEvent.change(getByTestId(testId), { target: { value: 2 } });
+    expect(getByTestId(`${testId}-option-1`).selected).toBeFalsy();
+    expect(getByTestId(`${testId}-option-2`).selected).toBeTruthy();
+    expect(getByTestId(`${testId}-option-3`).selected).toBeFalsy();
```
silvalaura, line 79: Do options have `selected`? Why check for falsy/truthy instead of true/false?

ccedrone: Changed testing method.

packages/react-magma-dom/src/components/Table/TablePagination.test.js
```diff
@@ -77,11 +77,16 @@ describe('Table Pagination', () => {
...
+      const appliedSelection = document.querySelector(
+        'select[data-testid=rowPerPageSelect]'
+      );
+
+      fireEvent.change(rowsSelect, { target: { value: 20 } });
 
       expect(handlePageChange).toHaveBeenCalledWith(expect.any(Object), 1);
       expect(handleRowsPerPageChange).toHaveBeenCalledWith('20');
       expect(getByText(/1-20/i)).toBeInTheDocument();
+      expect(appliedSelection).toHaveDisplayValue('20');
```
silvalaura, line 89: Is there a reason we are using `appliedSelection` here instead of `rowsSelect`? both are searching for the item with test id `rowPerPageSelect`

ccedrone: The `document.querySelector` method is able to see the change, where `getByTestId` does not.

silvalaura (approved):

Issue #1143, opened 2023-10-04
New Component: Navigation > Stepper


orion-cengage:
**Purpose:** Create new Stepper component for multi-step processes

**Figma:** https://www.figma.com/file/cPCfDVWC38T7Qt3dIL95sR/Components---Navigation?type=design&node-id=701%3A1229&mode=design&t=CiJHJjCC5jvgsmwM-1

![image](https://github.com/cengage/react-magma/assets/24397282/399225d5-9060-4656-bf16-da0bf5e2ed60)

### 1.0  Stepper can support any number of steps
-- 1.1  Should we enforce a minimum of two or rely on people not to be ridiculous?
-- 1.2 Stepper orientation can be horizontal or vertical (to be added in #1208 )(two separate components?)

### 2.0  Step anatomy
-- 2.1  Labels on every step
--- 2.1.1  Primary label
--- 2.1.2  Secondary label

![image](https://github.com/cengage/react-magma/assets/24397282/3f04ab80-749c-421a-ab8a-cee506e2e274)

-- 2.2 Option to hide labels

![image](https://github.com/cengage/react-magma/assets/24397282/675a7e0b-af14-47e9-8904-0c00fa41f228)

-- 2.3 Option to show step summary instead of labels on each step
--- 2.3.1 Step Summary shows the primary and secondary labels for the current step.
--- 2.3.2 The string that reads "step # of #" needs to be customizable so it could say things like "question # of #" or "module # of #".
--- 2.3.3 This could perhaps be done by Summary being a separate component and showing how to use it with the Stepper.

![image](https://github.com/cengage/react-magma/assets/24397282/a7c48aa3-b904-447b-bb09-9dcdac9aaff4)

-- 2.4  Status indicator
--- 2.4.1  Statuses include: 
---- 2.4.1.1 Current step

![image](https://github.com/cengage/react-magma/assets/24397282/0a17f650-6d35-46de-8866-3565cc6cdf2b)

---- 2.4.1.2 Completed Step

![image](https://github.com/cengage/react-magma/assets/24397282/88a4b47b-1c91-4a7c-b8d2-c83cf139858a)

---- 2.4.1.3 Incomplete Step

![image](https://github.com/cengage/react-magma/assets/24397282/4b8f3ebc-0f91-4b1d-a5d2-392e593405c1)

---- 2.4.1.4 Error

[...]

PR #1292, opened 2024-05-20
feat(Stepper): Displays step based content for use in multi-step interfaces.

ccedrone:
Issue: # [1143](https://github.com/cengage/react-magma/issues/1143)

## What I did

New Stepper component I built with my mitts, used commonly in wizard interfaces.

## Screenshots:
Steppah

## Checklist 

## How to test

- Verify the Stepper is showing the right values for the current, incomplete, and complete steps.
- Make sure the text doesn't collide between labels.
- Check the responsive states between labels, no labels, and summary.
- Check the regular layout states of labels, no labels, and summary.
- Test the a11y throughout the steps.
- See if 18n works as intended.


packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { ThemeInterface } from '../../theme/magma';
+import { useIsInverse } from '../../inverse';
+import { Step, StepProps, StepStatus } from './Step';
+import { Button } from '../Button';
+// import { I18nContext } from '../../i18n';
+
+/**
+ * @children required
+ */
+export interface StepperProps extends React.HTMLAttributes<HTMLDivElement> {
```
silvalaura, line 14: I think the Stepper props should be:
- active / activeStep: index of the active step
- isLabelVisuallyHidden
- isSummaryVisuallyHidden
- breakpoint: there's a requirement around this, but I have some questions about how to use it / if it's still applicable with only a vertical stepper
- isInverse
- testId

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+import { ThemeContext } from '../../theme/ThemeContext';
+import { ThemeInterface } from '../../theme/magma';
+import { useIsInverse } from '../../inverse';
+// import { I18nContext } from '../../i18n';
+
+/**
+ * @children required
+ */
+export interface StepProps extends React.HTMLAttributes<HTMLDivElement> {
```
silvalaura, line 13: We can do something similar to how Mantine handles it here: https://github.com/mantinedev/mantine/blob/master/packages/@mantine/core/src/components/Stepper/Stepper.tsx#L208-#L224 where some of the props are private, or we can handle it with context. 

Public StepProps should be:
- label
- ariaLabel
- summary/description
- hasError
- testId
- isInverse

I don't think we need to ask for the stepStatus, as we can figure it out with logic:
if index === active index, status = current;
if index < active index, status = complete;
if index > active index, status = incomplete;

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+    const isInverse = useIsInverse(isInverseProp);
+
+    return (
+      <StyledStep
+        currentStep={currentStep}
+        theme={theme}
+        isInverse={isInverse}
+        ref={ref}
+        data-testid={testId}
+        {...rest}
```
silvalaura, line 201: `...rest` should be first

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+      >
+        {stepStatus && (
+          <StyledStepIndicator
+            hasError={hasError}
+            isInverse={isInverse}
+            stepStatus={stepStatus}
+            theme={theme}
+          >
+            {stepStatus === 'complete' && !hasError && <CheckIcon />}
```
silvalaura, line 210: We should add ` aria-hidden="true"` to all the icons

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+            {hasError && <CrossIcon />}
+          </StyledStepIndicator>
+        )}
+        <StyledStepTextWrapper theme={theme} isInverse={isInverse} tabIndex={0}>
```
silvalaura, line 214: Do we need to pass theme and isInverse? It doesn't seem to use it

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+          {!summaryView && !hideLabels ? (
+            <>
+              {label && (
+                <StyledLabel label={label} isInverse={isInverse} theme={theme}>
```
silvalaura, line 218: StyledLabel doesn't need label

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+                  {label}
+                </StyledLabel>
+              )}
+              {secondaryLabel && (
+                <StyledSecondaryLabel
+                  secondaryLabel={secondaryLabel}
```
silvalaura, line 224: Same here, this isn't necessary

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * Error state for each step
+   * @default false
+   */
+  hasError?: boolean;
+  /**
+   * Hides label and secondaryLabel
+   * @default false
+   */
+  hideLabels?: boolean;
```
silvalaura, line 30: hideLabels should be passed to the Stepper, not the Step so it's consistent amongst all Steps.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+   */
+  breakpointStyle?: BreakPointStyle;
+  /*
+   * Sets a custom label for the Step count # of #.
+   */
+  stepDescriptionLabel?: string;
+  /*
+   * Sets a custom label for the completed Step.
+   */
+  completedStepDescription?: string;
```
silvalaura, line 30: The name is throwing me off because it seems like it would appear after a single step is completed. Instead, this is only visible when the whole stepper is done. Let's brainstorm different name options for this, and lets update the description to make this more clear. We should also state here what the default string is (and lets check with @orion-cengage on what that default should be)

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * Hides label and secondaryLabel in place of summary view
+   * @default false
+   */
+  summaryView?: boolean;
```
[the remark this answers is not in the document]
silvalaura: Also, summaryView should be on the Stepper, not the Step. But the comments above are still valid.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+  /*
+   * Current step value
+   */
+  currentStep?: number;
+  /**
+   * @internal
+   */
+  hasBreakpoint?: boolean;
```
silvalaura, line 38: why do we need both `hasBreakpoint` and `breakpoint`?

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+   */
+  breakpoint?: number;
+  /*
+   * Determines the Stepper view based on a minimum breakpoint number.
+   */
+  breakpointStyle?: BreakPointStyle;
+  /*
+   * Sets a custom label for the Step count # of #.
+   */
+  stepDescriptionLabel?: string;
```
silvalaura, line 26: Again, this prop name is throwing me off. What do other systems use? Or how can we make this more obvious? Whatever we decide, we should add the default in the description too.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+   * @internal
+   */
+  stepStatus?: StepStatus;
+  /**
+   * @internal
+   */
+  theme?: ThemeInterface;
+}
+
+export enum BreakPointStyle {
```
silvalaura, line 61: I'm also confused by this name, what does the breakpoint have to do with the labels/summary?

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+  hideLabels = 'hideLabels',
+  summaryView = 'summaryView',
+}
+
+function StepSeparatorBackgroundColors(props: StepperProps) {
```
silvalaura, line 66: Same comments in this file as on Step:
1. Lets use the enums instead of strings for comparisons
2. Functions should not be capitalized
3. Try destructuring prop names when it makes sense
4. Remove props being passed to styling constants when they are not used

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+      ) {
+        setBreakPointSummaryView(false);
+        setBreakPointHideLabels(false);
+      }
+
+      return () => window.removeEventListener('resize', handleResize);
+    }, []);
+
+    // Sets aria-current based on the active step, else false.
+    const ariaCurrent = index => {
```
silvalaura, line 179: This can be done in one line:
```
const ariaCurrent = index => currentStep === index;
```

If you don't use it anywhere else, do this in place instead of creating a constant.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+      if (currentStep === index) {
+        return true;
+      }
+      return false;
+    };
+
+    // Step states
+    const steps = React.Children.toArray(children);
+
+    const Steps = React.Children.map(
```
silvalaura, line 189: These two constants have the same name, the only difference is the capitalization of the first letter. This feels error prone and could cause confusion, can we change it?

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+              : currentStep >= index
+              ? StepStatus.active
+              : StepStatus.incomplete;
+
+          const item = React.cloneElement(child, {
+            'aria-current': ariaCurrent(index),
+            children: currentStep === index ? child : null,
+            key: index,
+            isInverse: isInverse,
+            hideLabels: breakPointHideLabels ? true : hideLabels,
```
silvalaura, line 208: This is equivalent to:
```suggestion
            hideLabels: breakPointHideLabels || hideLabels
```

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+        ) {
+          return child;
+        } else if (
+          // Retains last Step label and description
+          item.type === Step &&
+          currentStep >= +Steps.length &&
+          currentStep <= index + 1
+        ) {
+          return child;
+        }
```
silvalaura, line 254: Since you return the same thing in both case, this can be combined to one statement
```
if (
    item.type === Step &&
    (currentStep >= index && currentStep <= index ||
    currentStep >= +Steps.length && currentStep <= index + 1)
  ) {
    return child;
  }
  ```

File: packages/react-magma-dom/src/components/Stepper/Stepper.tsx [hunk 19]
```diff
@@ -0,0 +1,313 @@
...
+      }
+    );
+
+    // Final step description
+    const completedStepDescription = props.completedStepDescription
+      ? props.completedStepDescription
+      : 'Steps Completed';
```
silvalaura, line 261: We should use i18n instead of hard coding the value here

silvalaura: Actually, this leads to my question for @orion-cengage. Is this completed step description in the requirements? Does it need to be a prop or can we have a default in i18n that users can replace (as they can with all of our text strings)?

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+    // Shows both the step count out of total steps and the step label, allows for customization of the step title.
+    const stepSummary = (
+      <StyledSummary
+        data-testid="stepper summary"
```
silvalaura, line 266: Lets not hard code this test id and instead concatenate `summary` with whatever the user passes to the stepper.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+        isInverse={isInverse}
+        summaryView={summaryView}
+        theme={theme}
+      >
+        {currentStep < steps.length
+          ? stepDescriptionLabel
+            ? `${stepDescriptionLabel} ${currentStep + 1} of ${steps.length}`
+            : `Step
```
silvalaura, line 274: i18n

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,313 @@
...
+export interface StepperProps extends React.HTMLAttributes<HTMLDivElement> {
+  testId?: string;
+  /*
+   * Custom number for specific breakpoint responsive styling
+   */
+  breakpoint?: number;
+  /*
+   * Determines the Stepper view based on a minimum breakpoint number.
+   */
+  breakpointStyle?: BreakPointStyle;
```
silvalaura, line 22: I don't understand this prop and enum. Maybe we can chat about this.

silvalaura (changes requested):

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+Default.args = {};
+
+export const SummaryView = Template.bind({});
+SummaryView.args = {
+  summaryView: true,
+};
+
+export const HideLabels = Template.bind({});
+HideLabels.args = {
+  hideLabels: true,
```
silvalaura, line 168: If we can control these props and the content is the same, I don't think we need individual stories for them.

packages/react-magma-dom/src/components/Stepper/Stepper.test.js
```diff
@@ -0,0 +1,349 @@
...
+
+describe('Stepper', () => {
+  it('should find element by testId', () => {
+    const { getByTestId } = render(<Stepper testId={testId}>{TEXT}</Stepper>);
+
+    expect(getByTestId(testId)).toBeInTheDocument();
+  });
+
+  it('Does not violate accessibility standards', () => {
+    const { container } = render(<Stepper>{TEXT}</Stepper>);
```
silvalaura, line 20: This wouldn't be used like this (text inside the Stepper component), so I'm not sure this is a helpful test.

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+  label?: string;
+  /**
+   * Sub label beneath each step
+   */
+  secondaryLabel?: string;
+  /**
+   * @internal
+   */
+  stepStatus?: StepStatus;
+  isInverse?: boolean;
```
silvalaura, line 48: isInverse isn't used at the Step level, so we should remove it.

website/react-magma-docs/src/pages/api/stepper.mdx
```diff
@@ -0,0 +1,592 @@
...
+        </Button>
+      </Container>
+    </>
+  );
+}
+```
+
+## Completed Step Description
+
+When using `summaryView`, after all of the steps are complete, the `completedStepDescription` prop takes a string and changes the title.
```
silvalaura, line 216: I noticed in this example that Fourth Step is still visible. Should that disappear when all steps are completed?
image

website/react-magma-docs/src/pages/api/stepper.mdx
```diff
@@ -0,0 +1,592 @@
...
+        {currentStep === 4 && <div>Steps Completed</div>}
+      </Container>
+
+      <Container style={{ padding: '20px 0' }}>
+        <Button style={{ marginRight: '4px' }} onClick={handleOnPrevious}>
+          Previous
+        </Button>
+        <Button onClick={currentStep >= 4 ? handleFinish : handleOnNext}>
+          {currentStep >= 4 ? 'Finish' : 'Next'}
+        </Button>
```
silvalaura, line 139: We could be using the ButtonGroup component here to style the buttons

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
silvalaura: In the docs, you use `Custom Step label` as a heading, and that makes more sense to me. What do you think?

website/react-magma-docs/src/pages/api/stepper.mdx
```diff
@@ -0,0 +1,592 @@
...
+    </>
+  );
+}
+```
+
+## Breakpoints
+
+For responsive layouts, the `breakpoint` prop used along with `breakpointStyle` hides certain elements of the `Stepper` at a specified screen width. `breakpoint` takes a number for the screen width and `breakpointStyle` hides either labels or shows just the summary view. By default, step labels will display if no responsive props are used.
+
+#### Breakpoint - Hide Labels
```
silvalaura, line 350: This example falls outside the bounds
image

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+  breakpointStyle?: BreakPointStyle;
+  /*
+   * Index set by Stepper component
+   */
+  currentStep?: number;
+  /**
+   * Error state for each step
+   * @default false
+   */
+  hasError?: boolean;
```
silvalaura, line 25: If only the current step can have an error, we should have the Stepper (parent) have the hasError prop. (@orion-cengage can you confirm if we expect a past step to have an error?)

website/react-magma-docs/src/pages/api/stepper.mdx
```diff
@@ -0,0 +1,592 @@
...
+  );
+}
+```
+
+# Stepper Props
+
+**Any other props supplied will be provided to the wrapping `div` element.**
+
+<StepperProps />
+
```
silvalaura, line 591: We also need StepProps here

packages/react-magma-dom/src/components/Stepper/Stepper.test.js
silvalaura, line 1: Do we need tests for Step?

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+import { useIsInverse } from '../../inverse';
+import { transparentize } from 'polished';
+import { BreakPointStyle } from './Stepper';
+import { HiddenLabelText } from '../Checkbox';
+
+/**
+ * @children required
+ */
+export interface StepProps extends React.HTMLAttributes<HTMLDivElement> {
+  breakpointStyle?: BreakPointStyle;
```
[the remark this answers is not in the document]
ccedrone: Missed that when cleaning up, removed.

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+  /*
+   * Index set by Stepper component
+   */
+  currentStep?: number;
+  /**
+   * Error state for each step
```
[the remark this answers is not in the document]
ccedrone: Because it's assigned individually and doesn't need to be apart of the stepStatus enum logic.

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+        return props.theme.colors.primary500;
+      } else if (props.stepStatus === 'incomplete') {
+        return props.theme.colors.neutral300;
+      }
+    }
+  }
+}
+
+function StepCircleBackgroundColors(props: StepProps) {
+  if (props.isInverse) {
```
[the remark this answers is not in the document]
ccedrone: Fancy!

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+    if (props.hasError) {
+      return props.theme.colors.neutral100;
+    }
+    return props.theme.colors.primary600;
+  } else {
+    return props.theme.colors.neutral100;
+  }
+}
+
+const StyledStep = styled.div<StepProps>`
```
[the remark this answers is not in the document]
ccedrone: This uses `isInverse`, needs the props.

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,242 @@
...
+        : 'none'};
+  background: ${StepCircleBackgroundColors};
+  svg {
+    color: ${StepSvgColors};
+    width: ${props => props.theme.spaceScale.spacing05};
+    height: ${props => props.theme.spaceScale.spacing05};
+  }
+`;
+
+const StyledStepTextWrapper = styled.div<StepProps>`
```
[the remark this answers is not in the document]
ccedrone: It uses `isInverse`, needs props.

packages/react-magma-dom/src/components/Stepper/Step.tsx
ccedrone: Removed.

packages/react-magma-dom/src/components/Stepper/Step.tsx
ccedrone: Needed for styling: 

`function stepLabelColors(props: StepProps) {
  if (props.isInverse) {
    if (props.label) {
      return props.theme.colors.neutral100;
    } else if (props.secondaryLabel) {
      return transparentize(0.3, props.theme.colors.neutral100);
    }
  } else {
    if (props.label) {
      return props.theme.colors.neutral700;
    } else if (props.secondaryLabel) {
      return props.theme.colors.neutral500;
    }
  }
}`

packages/react-magma-dom/src/components/Stepper/Step.tsx
ccedrone: Needed for styling: 

`function stepLabelColors(props: StepProps) {
  if (props.isInverse) {
    if (props.label) {
      return props.theme.colors.neutral100;
    } else if (props.secondaryLabel) {
      return transparentize(0.3, props.theme.colors.neutral100);
    }
  } else {
    if (props.label) {
      return props.theme.colors.neutral700;
    } else if (props.secondaryLabel) {
      return props.theme.colors.neutral500;
    }
  }
}`

silvalaura: You can do something like 
`<StyledSecondaryLabel isInverse={isInverse} theme={theme}>{secondaryLabel}</StyledSecondaryLabel>`
then
```
function stepLabelColors(StepProps, hasLabel, hasSecondaryLabel){}
```
and
```
const StyledSecondaryLabel = styled.span<StepProps>`
  color: ${stepLabelColors(props, false, true)};
  font-size: ${props => props.theme.typeScale.size01.fontSize};
  line-height: normal;
  margin: 5px 0;
`;
const StyledLabel = styled.span<StepProps>`
  color: ${stepLabelColors(props, true, false)};
  font-size: ${props => props.theme.typeScale.size02.fontSize};
  font-weight: 600;
  line-height: normal;
`;
```
since you know it's always going to be true or false in those cases without passing the label string

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
orion-cengage: Discussed solution: We will rename it to "completionLabel" and set the default string to "All steps completed"

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
orion-cengage: Discussed solution: We will rename it to "completionLabel" and set the default string to "All steps completed"

packages/react-magma-dom/src/components/Stepper/Step.tsx
silvalaura: After speaking with Orion, it sounds like it's best we keep hasError at the step level.

orion-cengage: Discussed solution: We are going to allow any step to have an error state, not just the current step. Hopefully this will be an edge case.

orion-cengage:
Please make sure you are using the correct text styles so that you get the desired line-height, letter-spacing, etc. It's more than just the font size and weight. Also, when the line-heights are correct, there should only be a 2px gap between labels.

image

orion-cengage:
Since the stepper isn't interactive, none of the steps should ever get focus.

image

orion-cengage:
There's a typo on these two example pages in Storybook:

image

packages/react-magma-dom/src/components/Stepper/Step.tsx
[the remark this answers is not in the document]
ccedrone: Resolved per our conversation with Orion, keeping it as is.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
ccedrone: Removed `hasBreakpoint` no longer needed.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
ccedrone: `customStepLabel` it is.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
ccedrone: Changed to `StepperLayout`.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
ccedrone: Now named `breakpointLayout` per our discussion.

packages/react-magma-dom/src/components/Stepper/Step.tsx
ccedrone: It's used in the styling of `Step` throughout the component, I can make it `@internal`.

packages/react-magma-dom/src/components/Stepper/Stepper.test.js
ccedrone: They're covered within the Stepper tests ultimately and don't seem to need another separate test file.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+      }
+    );
+
+    // Final step description
+    const completionLabel = props.completionLabel
+      ? props.completionLabel
+      : i18n.stepper.completionLabel;
+
+    // Shows both the step count out of total steps and the step label, allows for customization of the step title.
+    const stepSummary = (
```
silvalaura, line 280: Since this is not reused and you're not doing any logic, I would prefer to not have it in a constant. For legibility purposes, lets throw the JSX in line 315 directly.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+              )}
+            </>
+          );
+        }
+      }
+    );
+
+    // When summaryView is set to true, this shows one step label and description at a time based on the active step below the Stepper component.
+
+    const StepLabels = React.Children.map(
```
silvalaura, line 254: This is getting called every time something changes because it's a constant in the body of the file, and that's not necessary because it's actually only used for summary. Also, if there are lots of steps, it's doing a lot of unnecessary looping.

To avoid this, wrap this in an anonymous function. What this will do is only evaluate it when it gets called:
```
    const getSummaryStepLabels = () => {...};
```

Also consider updating the name as it only gets called for summary, so we can be more specific.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+        if (child.type === Step) {
+          const stepStatusStyles =
+            currentStep >= index + 1
+              ? StepStatus.complete
+              : currentStep >= index
+              ? StepStatus.active
+              : StepStatus.incomplete;
+
+          const item = React.cloneElement(child, {
+            'aria-current': currentStep === index,
```
silvalaura, line 227: From a quick look it's unclear to be whether this should evaluate to true/false or 'step'/false (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) but we should investigate. I have seen examples that do both.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+            <>
+              {item}
+              {index !== step.length - 1 && (
+                <StyledSeparator
+                  isInverse={isInverse}
+                  stepStatus={stepStatusStyles}
+                  theme={theme}
+                />
+              )}
+            </>
```
silvalaura, line 246: I'm debating about this, and would love your thoughts. 

Currently, for the last child it returns the child and then false. 
image

We could use this array notation to prevent the unnecessary false child
```
[
  item,
  index !== step.length - 1 && (
    <StyledSeparator
      key={`separator-${index}`}
      isInverse={isInverse}
      stepStatus={stepStatusStyles}
      theme={theme}
    />
  )
]
```
image

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+          const stepStatusStyles =
+            currentStep >= index + 1
+              ? StepStatus.complete
+              : currentStep >= index
+              ? StepStatus.active
+              : StepStatus.incomplete;
+
+          const item = React.cloneElement(child, {
+            'aria-current': currentStep === index,
+            children: currentStep === index ? child : null,
```
silvalaura, line 228: Is `children` actually needed here?

website/react-magma-docs/src/pages/api/stepper.mdx
```diff
@@ -0,0 +1,599 @@
...
+    </>
+  );
+}
+```
+
+## Inverse
+
+```typescript
+import React from 'react';
+import { Container, Stepper, Step } from 'react-magma-dom';
```
silvalaura, line 529: Button is missing from all the imports

silvalaura (changes requested):

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,188 @@
...
+        </Button>
+      </Container>
+    </>
+  );
+};
+
+export const Default = Template.bind({});
+Default.args = {};
+
+// export const HideLabelsWithBreakpointSummaryView = Template.bind({});
```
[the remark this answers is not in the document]
ccedrone: That was removed.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+   */
+  breakpointLayout?: StepperLayout;
+  /*
+   * Changes the Stepper view.
+   */
+  layout?: StepperLayout;
+  /*
+   * Sets a custom label for the Step count # of #.
+   */
+  customStepLabel?: string;
```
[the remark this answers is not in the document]
ccedrone: Custom help implies the override from default in my opinion.

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+    return theme.colors.primary400;
+  } else {
+    if (stepStatus === StepStatus.complete) {
+      return theme.colors.primary500;
+    }
+    return theme.colors.neutral300;
+  }
+}
+
+const StyledStepper = styled.div<StepperProps>`
```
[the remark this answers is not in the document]
ccedrone: It is using the props:

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+`;
+
+const StyledStepContent = styled.div<StepperProps>`
```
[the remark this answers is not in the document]
ccedrone: That's old code, it's been updated with just one prop.

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,247 @@
...
+export interface StepProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * Error state for each step
+   * @default false
+   */
+  hasError?: boolean;
+  /**
+   * @internal
+   */
+  hasHideLabels?: boolean;
```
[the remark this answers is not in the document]
ccedrone: `hasHiddenLabels` it is.

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,188 @@
...
+    },
+    isInverse: {
+      control: {
+        type: 'boolean',
+      },
+    },
+  },
+} as Meta;
+
+const Template: Story<StepperProps> = args => {
```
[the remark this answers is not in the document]
silvalaura: We can set a useEffect to listen for any changes with the layout like so:
```
    React.useEffect(() => {
      setHideLabelsLayout(layout === StepperLayout.hideLabels);
      setShowLabelsLayout(layout === StepperLayout.showLabels);
      setSummaryViewLayout(layout === StepperLayout.summaryView);
    }, [layout]);
```

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,321 @@
...
+    const completionLabel = props.completionLabel
+      ? props.completionLabel
+      : i18n.stepper.completionLabel;
+
+    // Shows both the step count out of total steps and the step label, allows for customization of the step title.
+    const stepSummary = (
+      <StyledSummary
+        data-testid="stepper summary"
+        isInverse={isInverse}
+        layout={layout}
```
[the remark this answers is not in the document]
ccedrone: Removed!

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
ccedrone: Removed!

packages/react-magma-dom/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,247 @@
...
+                  secondaryLabel={secondaryLabel}
+                  isInverse={isInverse}
+                  theme={theme}
+                >
+                  {secondaryLabel}
+                </StyledSecondaryLabel>
+              )}
+            </>
+          ) : (
+            <HiddenLabelText>
```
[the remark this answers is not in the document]
ccedrone: I have this built into `Step` within the `<HiddenLabelText>` wrapper, this was the pattern I followed:

https://designsystem.digital.gov/components/step-indicator/

orion-cengage (commented):
- The margin between the primary and secondary labels should only be 2px, and there shouldn't be any bottom-margin.
image
image

packages/react-magma-dom/src/components/Stepper/Stepper.test.js
```diff
@@ -0,0 +1,401 @@
...
+
+      const step = getByTestId(testId).querySelector('span');
+
+      expect(step).toHaveStyleRule('background', magma.colors.danger500);
+    });
+
+    it('should have a primary label', () => {
+      const { getByText } = render(
+        <Stepper currentStep={1}>
+          <Step label={TEXT} hasError testId={testId} />
```
silvalaura, line 137: Not sure if you intended to keep `hasError` here and on line 152

silvalaura (commented):
Looking good! Just a few more final touch ups ✨

packages/react-magma-dom/src/components/Stepper/Stepper.test.js
```diff
@@ -0,0 +1,401 @@
...
+
+  it('Does not violate accessibility standards', () => {
+    const { container } = render(<Stepper>{TEXT}</Stepper>);
+
+    return axe(container.innerHTML).then(result => {
+      return expect(result).toHaveNoViolations();
+    });
+  });
+
+  describe('i18n', () => {
```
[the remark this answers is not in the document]
silvalaura: I don't think this was addressed yet

packages/react-magma-dom/src/components/Stepper/Stepper.test.js
silvalaura: 🥲

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
silvalaura, line 1: Notice that multiple items are getting the same testId here, which we do not want. They should be unique. We should have caught this in the unit tests, so perhaps there's a gap there.
image

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const [currentStep, setCurrentStep] = React.useState(0);
+
+  const [numberOfSteps, setNumberOfSteps] = React.useState('4');
+
+  console.log(`number of steps: ${numberOfSteps}`);
```
silvalaura, line 68: Let's remove this console log now that it's working. This looks and works great by the way!

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+        alignItems={FlexAlignItems.center}
+        justify={FlexJustify.spaceBetween}
+        style={{ paddingTop: '20px' }}
+      >
+        <Flex behavior={FlexBehavior.item}>
+          <Input
+            labelPosition={LabelPosition.left}
+            labelText="Step Amount"
+            value={numberOfSteps}
+            onChange={event => setNumberOfSteps(event.target.value)}
```
silvalaura, line 131: Edge case but we should probably prevent this from ever becoming 0 and causing this error
image

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+};
+
+export const Default = Template.bind({});
+Default.args = {
+  // numberOfSteps: 6,
+};
+
+export const WithError = ErrorTemplate.bind({});
+WithError.args = {
+  // numberOfSteps: 4,
```
silvalaura, line 199: This can be removed

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+      setCurrentStep(0);
+    }
+  };
+
+  const steps = Number(numberOfSteps);
+
+  const step = [...Array(steps)].map((_, i) => {
+    ++i;
+    return (
+      <Step
```
silvalaura, line 91: We can add `testId={`step-${i}`}` here for testing the ids

packages/react-magma-dom/src/components/Stepper/Stepper.test.js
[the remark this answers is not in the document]
silvalaura: Bump ^

packages/react-magma-dom/src/components/Stepper/Stepper.test.js
```diff
@@ -0,0 +1,471 @@
...
+
+        const separator = getByTestId(testId).nextElementSibling;
+
+        expect(separator).toHaveStyleRule(
+          'background',
+          magma.colors.primary400
+        );
+      });
+
+      it('should have a inverse completed styled separator', () => {
```
silvalaura, line 317: `a inverse` is used in a few tests 🙈

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,322 @@
...
+        }
+      }
+
+      window.addEventListener('resize', handleResize);
+      handleResize();
+
+      return () => window.removeEventListener('resize', handleResize);
+    }, [windowWidth]);
+
+    // Step states
```
silvalaura, line 218: I don't think this comment is still applicable here? Or maybe it's above the wrong line of code

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,322 @@
...
+  width: ${props => (props.showLabelsLayout ? 'calc(100% - 24px)' : '100%')};
+  height: 2px;
+  top: 11px;
+  left: ${props => (props.showLabelsLayout ? 'calc(50% + 12px)' : '')};
+  position: ${props => (props.showLabelsLayout ? 'absolute' : 'relative')};
+  align-self: baseline;
+  transition: all 0.4s ease;
+`;
+
+const StyledSummary = typedStyled.div<StepperProps>`
```
silvalaura, line 112: Did we lose some space between the stepper and the summary? Feels a little tight
image

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,322 @@
...
+  stepLabel?: string;
+  /**
+   * Sets a custom label for the completed Step.
+   * @default 'All steps completed'
+   */
+  completionLabel?: string;
+  /**
+   * Current step value.
+   */
+  currentStep?: number;
```
silvalaura, line 43: We should probably make `currentStep` required. This whole thing doesn't do anything without it

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
orion-cengage: Agreed. If you put borders on the indicator portion, and the container for the labels, there should be 8px of space in between.
image

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -215,7 +234,8 @@ export const Stepper = React.forwardRef<HTMLDivElement, StepperProps>(
       return () => window.removeEventListener('resize', handleResize);
     }, [windowWidth]);
 
-    // Step states
+    const stepLabels = React.Children.toArray(children);
```
silvalaura, line 237: These aren't the labels, this is an array of the Stepper children

silvalaura: This could be updated to something like below to avoid checking `React.Children.toArray(children)` twice
```
    const stepperChildren = React.Children.toArray(children);
    const numberOfSteps = stepperChildren.length;
```

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -232,9 +252,31 @@ export const Stepper = React.forwardRef<HTMLDivElement, StepperProps>(
...
+            if (React.isValidElement(child)) {
+              return Object.keys(child.props).includes('label');
+            }
+          });
+
+          const secondaryLabel = stepLabels.some(child => {
+            if (React.isValidElement(child)) {
+              return Object.keys(child.props).includes('secondaryLabel');
+            }
+          });
```
silvalaura, line 274: Right now, this is getting evaluated inside the loop of children, so this gets called a lot of times. We actually only need to call this once, so here are a few suggestions on how we can make this more efficient and concise:
- move this outside of the loop
- update variable names to be descriptive-- what we are doing here is checking if all the children have labels, so let's call it that for clarity. don't be afraid to add comments if you think any logic could be confusing in the future
- since label & secondaryLabel are using the same loop, we can combine those checks for efficiency
- we don't need an additional `bothLabels` check if we already have a value for label and secondarylabel, we can just check if both are true
- our ternary was getting a little crazy, so we could simplify that as well

Here's an example: 
image

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -95,24 +97,41 @@ const StyledWrapper = typedStyled.div`
...
+      ? 'calc(100% - 24px)'
+      : '100%'};
   height: 2px;
   top: 11px;
-  left: ${props => (props.showLabelsLayout ? 'calc(50% + 12px)' : '')};
-  position: ${props => (props.showLabelsLayout ? 'absolute' : 'relative')};
+  left: ${props =>
+    (props.showLabelsLayout && props.bothLabels) ||
+    (props.showLabelsLayout && props.label) ||
+    (props.showLabelsLayout && props.secondaryLabel)
```
silvalaura, line 119: Consider making this more concise:  
```${props => props.showLabelsLayout && (props.bothLabels || props.label || props.secondaryLabel)}```

File: packages/react-magma-dom/src/components/Stepper/Step.tsx [hunk 63]
```diff
@@ -0,0 +1,272 @@
...
+                  isInverse={isInverse}
+                  data-testid={testId && `${testId}-secondaryLabel`}
+                  theme={theme}
+                >
+                  {secondaryLabel}
+                </StyledSecondaryLabel>
+              )}
+            </>
+          ) : (
+            <HiddenLabelText>{labels()}</HiddenLabelText>
```
silvalaura, line 266:
```suggestion
            <HiddenLabelText>{label && secondaryLabel ? `${label} ${secondaryLabel}` : label || secondaryLabel}</HiddenLabelText>
```

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const handleOnChange = event => {
+    if (numberOfSteps > 0) {
+      return setNumberOfSteps(Number(event.target.value));
+    } else {
+      return setNumberOfSteps(1);
+    }
+  };
+
+  const steps = numberOfSteps;
```
silvalaura, line 92: This is redundant

packages/react-magma-dom/src/components/Stepper/Stepper.stories.tsx
```diff
@@ -0,0 +1,205 @@
...
+        testId={`Step ${i}`}
+        label={`Step ${i}`}
+        secondaryLabel={`Description area in secondaryLabel component ${i}`}
+      />
+    );
+  });
+
+  return (
+    <>
+      <Stepper currentStep={currentStep} {...args}>
```
silvalaura, line 108: Lets add `ariaLabel` to our storybook example in case Perkins reviews this again from here

packages/react-magma-dom/src/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,366 @@
...
+          };
+
+          if (
+            showLabelsLayout &&
+            (allStepsHaveLabels || allStepsHaveSecondaryLabels)
+          ) {
+            return (
+              <StyledWrapper
+                aria-current={currentStep === index}
+                // data-testid={testId && `${testId}-step-${index}`}
```
silvalaura, line 304: If we don't need it, lets get rid of this line

website/react-magma-docs/src/pages/api/stepper.mdx
silvalaura, line 1: We need to document the new `ariaLabel` prop, and probably use it in our examples below

silvalaura (approved):

Issue #1250, opened 2024-04-15
Dropdown - Hide SVGs from screen readers


orion-cengage:
CENG_3300

Docs: https://react-magma.cengage.com/version/4.3.0/api/dropdown/

Success Criteria: https://www.w3.org/WAI/WCAG21/Understanding/non-text-content

Within the dropdown component, there is an SVG next to the button that opens the dropdown that is not named and is not hidden from assistive technology. This could confuse users who reach these elements without knowing what they are for.

Add `aria-hidden="true"` to the the SVGs within the button that opens the dropdown component.

orion-cengage:
I tested this is Storybook and Docs, in Chrome on Mac. Everything looks good.

PR #1297, opened 2024-05-24
fix(TreeView): Update styles

silvalaura:
## What I did

Address the following bugs:

- [X] Inverse, hover state is too light and barely visible. The background color needs to be (0, 0, 0, .20) instead of ".05)
- [X] Inverse, focus state, the outline needs to be our inverse focus color like on other inputs
- [X] Based on feedback from Perkins on other components, we need to hide the actual svgs for icons from screen readers using aria-hidden=true. See https://github.com/cengage/react-magma/issues/1250 for example.
- [X] When selection is turned off, we still need a hover state on nodes.

## Screenshots:

image

## Checklist 

## How to test

Retest the bugs above on Storybook

### DIFF
--- a/.changeset/treeview-style.md
+++ b/.changeset/treeview-style.md
@@ -1,0 +1,5 @@
+---
+"react-magma-dom": minor
+---
+
+feat(TreeView): TreeView & TreeItem style updates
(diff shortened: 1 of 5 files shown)


orion-cengage:
Retested changes and everything looks good.

PR #1298, opened 2024-05-24
fix(Input): Add `inputWrapperStyle` prop

silvalaura:
Issue: #

## What I did

Add `inputWrapperStyle` prop to allow more flexibility with Input widths

## Screenshots:

Before:
image

After:
image

## Checklist 

### DIFF
--- a/.changeset/input-wrapperStyle.md
+++ b/.changeset/input-wrapperStyle.md
@@ -2,4 +2,4 @@
 "react-magma-dom": patch
 ---
 
-fix(Input): Add `wrapperContainerStyle` prop to allow more flexibility with Input widths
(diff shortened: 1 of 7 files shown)


silvalaura:
TODO: 
- add unit test
- ensure label wraps
- isClearable???

orion-cengage (changes requested):
I don't see an update to the documentation about this new prop. Maybe I misunderstood that it would be called out explicitly.

In the Storybook example, can we have isClearable turned off? I don't want to advertise the issue we have with the arrows.

I don't know how I missed this for so long, but the icon on the error message needs to be top aligned, not center.

image

image

orion-cengage (approved):

ccedrone:
✔️

Issue #1120, opened 2023-08-25
Component Update: Hyperlink - Decoration and Underline


orion-cengage:
Purpose: We need to update our Hyperlink component in the following ways:

- Make it easy to add an icon to the left or right side
- Make the underline an option you can turn off

**Figma:** https://www.figma.com/file/VoWYiiOKKppcYm5QLgPyDC/Components---Inputs?type=design&node-id=1112%3A537&mode=design&t=kdAzH0OJ184MuCgA-1

### 1.0 Hyperlink w/ Icon
-- 1.1 Make it easy to add an icon to the right and/or left side of the link

![image](https://github.com/cengage/react-magma/assets/24397282/9e42ec4b-9eec-4913-8e1b-942438a8e0f3)

-- 1.2 Commonly needed when links are used on their own, away from surrounding text, to help indicate what the link will do.
-- 1.3 Dev should be able to set the icon size to a value that will pair best with the size of the text
-- 1.4 Dev should be able to set the space between the text and the icon using our spacing system.

![Image](https://github.com/cengage/react-magma/assets/24397282/7e192c1b-e517-48a8-bd52-111aee1d0d5b)

-- 1.5 Focus style goes around both the text and icon, not just the text.

![image](https://github.com/cengage/react-magma/assets/24397282/098c0055-f6a4-48ab-8e20-f6d0be9d9674)

-- 1.6 Icon color should match the text.

### 2.0 Hyperlink underline
-- 2.1 This only impacts the hyperlink component. When a hyperlink is used within other elements like paragraphs, the underline is still mandatory

![Image](https://github.com/cengage/react-magma/assets/24397282/992f00f1-5a14-4127-b9dc-25594a645627)

![Image](https://github.com/cengage/react-magma/assets/24397282/4dba38fa-dac8-403b-bbc7-17cfab1bda29)

-- 2.2 Dev should be able to turn off the underline when the component is used. 
-- 2.3 On hover, the underline is used and is not optional.

### 3.0 Hyperlink hover color
-- 3.1 Change hover color of all hyperlink styling to the following:
--- 3.1.1 Default = primary-700

![image](https://github.com/cengage/react-magma/assets/24397282/ec661ecc-ac97-4a08-a93f-accd2b9fdc92)

--- 3.1.2 Inverse = neutral-100

![image](https://github.com/cengage/react-magma/assets/24397282/60ea3d70-fc55-4493-8fd7-1b34c30a0a83)

ccedrone:
[...]

PR #1304, opened 2024-06-03
Hyperlink: Support Icons and toggling underline styling

silvalaura:
Issue: #1120

## What I did

- Added support for an icon on the left, right, or both sides of a Hyperlink
- Added `hasUnderline` prop that defaults to true
- Updated Hyperlink and native anchors hover color

## Screenshots:

image
image

## Checklist 

## How to test

Confirm all use cases for hyperlinks are working. Consider:
- styled as button or link
- inverse on/off
- hasUnderline on/off
- passing 1 icon
- passing 2 icons
- icon positions: left, right or both
- using it inside `<Paragraph>`, `<p>`, and other text wrappers

### DIFF
--- a/.changeset/hyperlink.md
+++ b/.changeset/hyperlink.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Hyperlink): Support Icons and toggling underline styling
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -21,4 +21,5 @@ export const parameters = {
   viewport: {
     viewports: INITIAL_VIEWPORTS,
   },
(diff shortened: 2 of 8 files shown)


orion-cengage (changes requested):
Feedback:

- Please correct the hover state color for the inverse hyperlink to be neutral-100 (#FFFFFF).
- In the default examples, the paragraph specifically, when the icon comes before the text it is making the alignment wonky. The inverse example showing the icon after the text looks correct.
image

- When we use a hyperlink inside a passage of text (paragraph or otherwise), can we make sure the line-height of the link matches that of the surrounding text? I think this will help avoid weird alignment/baseline issues. I think the "display: inline-flex" is the culprit.
image

Perhaps some styles from how Storybook is doing it here could help?
image

orion-cengage (approved):

ccedrone (approved):

Issue #1315, opened 2024-06-12
A11y: Modal aria labels


silvalaura:
We can improve the accessibility in our modals by updating the following:

- `aria-describedBy=“modal”` — This is not required and should be removed
- `aria-label=“modal”` — We hard code this right now and it’s fine that it’s hard coded as long as a user is using the heading. This is because `aria-labeledby` takes precedence over `aria-label`. For modals that don't have a `header` prop, we need to allow them to pass in a custom label. We should add a `ariaLabel` prop and do something like `aria-label={header ? null : props.ariaLabel}`

orion-cengage:
I reviewed the the modal update in Dev Storybook and it looks good to me.

ccedrone:
In Storybook > Drawer > Site Navigation, I don't see an `aria-label` on the Drawer, everything else looks good though.

PR #1318, opened 2024-06-18
fix(modal): Update aria labels

silvalaura:
Issue: #1315 

## What I did

- Removed unnecessary `aria-describedBy=“modal”`
- Added `ariaLabel` prop so that Modals without headers can customize the `aria-label` instead of defaulting to `aria-label="modal"` which is not correct for a11y standards.
- Updated unit tests for Modal and Drawer

## Screenshots:

image

## Checklist 

## How to test

Test the following in storybook examples and docs: 

Modals AND Drawers with headers: 
1. should not have aria-describedBy
2. should have aria-labeledBy
3. should not have aria-label

Modals AND Drawers without headers: 
1. should not have aria-describedBy
2. should not have aria-labeledBy
3. should have aria-label

### DIFF
--- a/.changeset/modal-aria.md
+++ b/.changeset/modal-aria.md
@@ -1,0 +1,5 @@
+---
+"react-magma-dom": patch
+---
+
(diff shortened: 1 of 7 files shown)


ccedrone (approved):

orion-cengage (approved):

ccedrone:
In Storybook > Drawer > Site Navigation, I don't see an `aria-label` on the Drawer, everything else looks good though.

Issue #1323, opened 2024-06-20
Deprecate LineChart


silvalaura:

orion-cengage:
Deprecation alert looks good to me.

PR #1326, opened 2024-06-20
chore: Deprecate LineChart

silvalaura:
Issue: #1323

## What I did

Deprecate LineChart

image


orion-cengage (changes requested):
Since CarbonChart is labeled as Chart in the navigation, maybe we should link to that page inside of the error message? Make it a little easier for people to find the right page.

Issue #1309, opened 2024-06-05
A11y: Treeview feedback from Perkins


orion-cengage:
The follow is feedback from Perkins after a thorough a11y review.

### Keyboard and focus

- Walking back up the tree using the left arrow isn't working. Pressing the left arrow on a nested end node should move up to the parent node. Pressing left again should close the parent node if it is open or move up again if it is closed. Pressing left a third time should move up to the grandparent node (if it exists). To see this behavior in action, open the [APG treeview example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/) and get keyboard focus into a child item of Project 3. Then press the left arrow key several times to see how it moves back up the tree.
- When tabbing into a multi-select treeview with selected items, the focus should land on the first selected item. This seems to happen on the initial load when tabbing forward. Tabbing backward into the treeview doesn't bring focus to the first selected item. After that, the focus seems to not return to the treeview.

### Suggestions

- Animations do not respect a user's [reduced motion preference (MDN)](https://urldefense.com/v3/__https:/developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion__;!!MXVguWEtGgZw!LbMMJFCfCiiaKldD5gJ5rgH7aq23uCweuAcXFvmvrm8H7QPaUR2JLcnPIAHsVTHaMFGENVWCkRv6B9QSsoLx4eSYFcI$). The animations are subtle but do cause significant, visible movement of treeitems. Consider removing animations when users have reduced motion enabled.

_The screenreader part of this review has been moved to https://github.com/cengage/react-magma/issues/1420_

chris-cedrone-cengage:
Looks good here, can't find anything off!

orion-cengage:
I took a look in Storybook and everything looks good to me.

PR #1338, opened 2024-07-01
fix(TreeView): Accessibility updates

silvalaura:
Issue: #1309

## What I did

Address the following Perkins a11y concerns around Tree:

1. Walking back up the tree using the left arrow isn't working. Pressing the left arrow on a nested end node should move up to the parent node. Pressing left again should close the parent node if it is open or move up again if it is closed. Pressing left a third time should move up to the grandparent node (if it exists). To see this behavior in action, open the [APG treeview example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/) and get keyboard focus into a child item of Project 3. Then press the left arrow key several times to see how it moves back up the tree. ✅ 
2. Animations do not respect a user's [reduced motion preference (MDN)](https://urldefense.com/v3/__https:/developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion__;!!MXVguWEtGgZw!LbMMJFCfCiiaKldD5gJ5rgH7aq23uCweuAcXFvmvrm8H7QPaUR2JLcnPIAHsVTHaMFGENVWCkRv6B9QSsoLx4eSYFcI$). The animations are subtle but do cause significant, visible movement of treeitems. Consider removing animations when users have reduced motion enabled. ✅ 
3. When tabbing into a multi-select treeview with selected items, the focus should land on the first selected item. This seems to happen on the initial load when tabbing forward. Tabbing backward into the treeview doesn't bring focus to the first selected item. After that, the focus seems to not return to the treeview. ✖️  _Cannot reproduce, so did not make any changes_

## Checklist 

## How to test

- Notes from Perkins are above, so follow those steps for reproducing.
- To test the reduced motion in macOS: System Preferences > Accessibility > Display > Reduce motion.


chris-cedrone-cengage (approved):

orion-cengage (approved):
I tested in Storybook and everything seems to be working as desired.

Issue #1335, opened 2024-06-26
Styled export no longer works for adopters


silvalaura:
We recently updated our `styled` imports: https://github.com/cengage/react-magma/commit/c1ef7d0cb94b1068d0b5fe237b901a6bba839a66 as a result of [failing builds.](https://laura-silva.notion.site/_styled-is-not-defined-bug-d1eabe4ab171480bb77e37b0dd545e70) This works fine internally within Magma, but our consumers are also expecting to import `styled` from our package and that no longer works. 

This can be tested by reverting this PR: https://github.com/cengage/react-magma/pull/1313 and also confirming on CodeSandbox that `styled` can be imported from `react-magma-dom`. 

Notes:
- See https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/theme/styled.ts

silvalaura:
Confirmed:
- Every component looks as expected on the docs site ✅ 
- When trying to import `styled` from react-magma-dom in CodeSandbox, I can style components as expected ✅

PR #1357, opened 2024-08-01
 fix (Build) Updating emotion to v11

vverein:
Issue: [#1335](https://github.com/cengage/react-magma/issues/1335)

## What I did

- Updated emotion to v11
- Updated unit test snapshots

## Screenshots:
 No visual changes expected

## Checklist 

## How to test

1.  install packages
2. ensure that the project runs locally and all parts build successfully
3. check all the components to see there are no new visual changes


.changeset/emotion-update-new.md
```diff
@@ -0,0 +1,5 @@
+---
+"react-magma": minor
```
silvalaura, line 2: This is technically correct because we are updating all the packages, but I think this will fail the automation as it expects one of the package names. I think we actually will need to make a changeset file for each of the packages so that when we release, all of the packages get updated:
- react-magma-dom
- @react-magma/charts
- @react-magma/dropzone
- @cengage-patterns/header
(we can skip docs)

vverein: Done 👍

silvalaura (approved):

Issue #1146, opened 2023-10-18
Dropzone: File browser does not open starting in v4


silvalaura:
**Describe the bug**
When trying to Browse File in a Dropzone, the file browser does not open in versions 4 and above. Note that dragging a file into the dropzone still works.

**To Reproduce**
1. Go to https://react-magma.cengage.com/version/4.1.1/api/dropzone/
2. Click on Browse File
3. Notice that nothing happens

**Expected behavior**
When clicking Browse File, the system file browser should appear.

silvalaura:
Confirmed Browse Files opens the file browser on Chrome, Firefox and Safari:

![Image](https://github.com/user-attachments/assets/50d0a1b9-2f39-494a-a042-e0af332a399e)

PR #1362, opened 2024-08-07
fix(Dropzone): Open file browse on button click

yatimov:
Issue: #1146 

## What I did
Fix the 'Open File' menu to open on button click.

## Checklist 

## How to test
1.Go to dropzone in storybook
2.Try to open file by click `Browse Files`

### DIFF
--- a/.changeset/dropzone-files-browse.md
+++ b/.changeset/dropzone-files-browse.md
@@ -1,0 +1,5 @@
+---
+"@react-magma/dropzone": patch
+---
+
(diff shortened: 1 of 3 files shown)


silvalaura:
Changes look good. Is it possible to write a unit test so we can catch if this breaks again in the future?

yatimov:
> Changes look good. Is it possible to write a unit test so we can catch if this breaks again in the future?

@silvalaura added a test

silvalaura (approved):

PR #1364, opened 2024-08-08
fix(datePicker): Fix date picker clearing date issue

### LINKED ISSUE #963: Bug: Clearing DatePicker
**Describe the bug**
The date picker input never clears the date after the user has chosen a date through the calendar. This also makes it impossible to programmatically clear that field after it's been selected. We have some code in there to support `isClearable`, but we should expose this functionality.

**To Reproduce**
Go to https://codesandbox.io/p/sandbox/datepicker-doesnt-programmatically-clear-8f3p5d?file=%2Fexample.tsx%3A24%2C11

Path 1:
1. Select a date
2. Click the isClearable X
3. Notice that the Chosen Date is still the old date. 

Path 2:
1. Select a date
2. Click on ClearDate
3. Notice the Chosen Date value is cleared, but the input still displays the old date

**Expected behavior**
- DatePicker can be programmatically cleared through any of the Input events (onDateChange, onChange, onInputChange)
- isClearable button fully clears the selected dates
- Example is added to Docs site
- Unit tests are added for these two use cases

OlenaYashyna:
Issue: #963

## What I did
Fix clearing DatePicker functionality
Add 'Clearing the Date' section to the documentation (docs and storybook)

## Checklist 

### DIFF
--- a/packages/react-magma-dom/src/components/Input/Input.test.js
+++ b/packages/react-magma-dom/src/components/Input/Input.test.js
@@ -296,7 +296,6 @@ describe('Input', () => {
 
   it('should clear the input when the clear input button is clicked', () => {
     const onClear = jest.fn();
-    const onDateChange = jest.fn();
     const labelText = 'Input Label';
     const value = 'Test Value';
     const { getByTestId, getByLabelText } = render(
@@ -304,7 +303,6 @@ describe('Input', () => {
         labelText={labelText}
         value={value}
         onClear={onClear}
-        onDateChange={onDateChange}
         isClearable
       />
     );
@@ -566,9 +564,8 @@ describe('Input', () => {
 
     it('Shows the label "characters allowed" equal to the maxCount if the user clears the input by backspacing', () => {
       const onChange = jest.fn();
-      const onDateChange = jest.fn();
       const { getByText, getByLabelText } = render(
-        <Input labelText={labelText} maxCount={4} onChange={onChange} onDateChange={onDateChange} />
+        <Input labelText={labelText} maxCount={4} onChange={onChange} />
       );
 
       fireEvent.change(getByLabelText(labelText), {
@@ -588,12 +585,10 @@ describe('Input', () => {
 
     it('Shows the label "characters allowed" equal to the maxCount if the user clears the input by clicking the onClear button', () => {
       const onClear = jest.fn();
-      const onDateChange = jest.fn();
       const { getByText, getByLabelText, getByTestId } = render(
         <Input
           labelText={labelText}
           onClear={onClear}
-          onDateChange={onDateChange}
           isClearable
           maxCount={4}
         />
@@ -610,8 +605,6 @@ describe('Input', () => {
 
       expect(getByText('4 ' + charactersAllowed)).toBeInTheDocument();
       expect(onClear).toBeCalled();
-      expect(onDateChange).toHaveBeenCalledWith(null);
-      expect(getByLabelText(labelText)).toHaveAttribute('value', '');
     });
   });
 });
--- a/packages/react-magma-dom/src/components/Input/Input.test.js
+++ b/packages/react-magma-dom/src/components/Input/Input.test.js
@@ -296,6 +296,7 @@ describe('Input', () => {
 
   it('should clear the input when the clear input button is clicked', () => {
     const onClear = jest.fn();
+    const onDateChange = jest.fn();
     const labelText = 'Input Label';
     const value = 'Test Value';
     const { getByTestId, getByLabelText } = render(
@@ -303,6 +304,7 @@ describe('Input', () => {
         labelText={labelText}
         value={value}
         onClear={onClear}
+        onDateChange={onDateChange}
         isClearable
       />
     );
@@ -564,8 +566,9 @@ describe('Input', () => {
 
     it('Shows the label "characters allowed" equal to the maxCount if the user clears the input by backspacing', () => {
       const onChange = jest.fn();
+      const onDateChange = jest.fn();
       const { getByText, getByLabelText } = render(
-        <Input labelText={labelText} maxCount={4} onChange={onChange} />
+        <Input labelText={labelText} maxCount={4} onChange={onChange} onDateChange={onDateChange} />
       );
 
       fireEvent.change(getByLabelText(labelText), {
@@ -585,10 +588,12 @@ describe('Input', () => {
 
     it('Shows the label "characters allowed" equal to the maxCount if the user clears the input by clicking the onClear button', () => {
       const onClear = jest.fn();
+      const onDateChange = jest.fn();
       const { getByText, getByLabelText, getByTestId } = render(
         <Input
           labelText={labelText}
           onClear={onClear}
+          onDateChange={onDateChange}
           isClearable
           maxCount={4}
         />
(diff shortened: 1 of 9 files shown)


packages/react-magma-dom/src/components/Search/Search.test.js
```diff
@@ -106,6 +107,7 @@ describe('Search', () => {
         labelText={labelText}
         onSearch={onSearchSpy}
         onClear={onClear}
+        onDateChange={onDateChange}
```
silvalaura, line 110: Search component shouldn't need `onDateChange` as search has nothing to do with dates

packages/react-magma-dom/src/components/Input/Input.test.js
```diff
@@ -296,13 +296,15 @@ describe('Input', () => {
...
     const onClear = jest.fn();
+    const onDateChange = jest.fn();
     const labelText = 'Input Label';
     const value = 'Test Value';
     const { getByTestId, getByLabelText } = render(
       <Input
         labelText={labelText}
         value={value}
         onClear={onClear}
+        onDateChange={onDateChange}
```
silvalaura, line 307: This should be an optional prop-- most inputs won't have dates, so we shouldn't expect adopters to use this prop

website/react-magma-docs/src/pages/api/date-pickers.mdx
```diff
@@ -95,6 +95,48 @@ export function Example() {
 }
 ```
 
+## Clearing the Date
+
+After a date has been selected the user can programmatically clear the input field and the `Chosen Date:` field via `isClearable` X button or `Clear Date` button.
```
silvalaura, line 100: We can make this more generic since we don't know the exact use case for date clearing. I think it would be enough to say something along the lines of ```Dates can be cleared with the `isClearable` prop, or programmatically through the `onDateChange` function by passing null.```

silvalaura (commented):
I could be convinced either way, but do we think that we should add this clearing logic for `onInputChange` and/or `onChange`? Right now if I manually clear the input using the backspace, we chosen date remains as the last valid date 🤔

OlenaYashyna:
> I could be convinced either way, but do we think that we should add this clearing logic for `onInputChange` and/or `onChange`? Right now if I manually clear the input using the backspace, we chosen date remains as the last valid date 🤔

I've tested this case and it works fine. I made a screen record to check, if I missed any case.

https://github.com/user-attachments/assets/21abd3cc-d62e-415e-ae8c-af476204c366

packages/react-magma-dom/src/components/Search/Search.test.js
OlenaYashyna: fixed optional prop using

packages/react-magma-dom/src/components/Input/Input.test.js
OlenaYashyna: fixed optional prop using

website/react-magma-docs/src/pages/api/date-pickers.mdx
OlenaYashyna: Updated. Thank you for the idea. So far it's difficult for me to come up with a description

silvalaura:
> > I could be convinced either way, but do we think that we should add this clearing logic for `onInputChange` and/or `onChange`? Right now if I manually clear the input using the backspace, we chosen date remains as the last valid date 🤔
> 
> I've tested this case and it works fine. I made a screen record to check, if I missed any case.
> 
>  Date.Picker.clearing.the.date.mov

Yeah this looks great! I was just wondering about the case of partial removal, so backspacing the year only completely clears the input field because the date is no longer valid-- would developers expect that date to come back to them as empty since it visually looks cleared, or is it okay that the date they get back is the last valid one? This may be a question for our designer on what functionality we expect. 

The other part of my question was around the two other functions that adopters hook into: `onInputChange` and `onChange`. I think I'm okay with keeping clearing the date specific to `onDateChange`, so lets leave it as is.

OlenaYashyna:
> > > I could be convinced either way, but do we think that we should add this clearing logic for `onInputChange` and/or `onChange`? Right now if I manually clear the input using the backspace, we chosen date remains as the last valid date 🤔
> > 
> > 
> > I've tested this case and it works fine. I made a screen record to check, if I missed any case.
> > Date.Picker.clearing.the.date.mov
> 
> Yeah this looks great! I was just wondering about the case of partial removal, so backspacing the year only completely clears the input field because the date is no longer valid-- would developers expect that date to come back to them as empty since it visually looks cleared, or is it okay that the date they get back is the last valid one? This may be a question for our designer on what functionality we expect.
> 
> The other part of my question was around the two other functions that adopters hook into: `onInputChange` and `onChange`. I think I'm okay with keeping clearing the date specific to `onDateChange`, so lets leave it as is.

Now I see what case you're talking about. After partial removal the input is cleared but Chosen day is not. As for me, it would be more compatible if the chosen date was updated (cleared) as well. If so, I will update the PR

silvalaura:
> > > > I could be convinced either way, but do we think that we should add this clearing logic for `onInputChange` and/or `onChange`? Right now if I manually clear the input using the backspace, we chosen date remains as the last valid date 🤔
> > > 
> > > 
> > > I've tested this case and it works fine. I made a screen record to check, if I missed any case.
> > > Date.Picker.clearing.the.date.mov
> > 
> > 
> > Yeah this looks great! I was just wondering about the case of partial removal, so backspacing the year only completely clears the input field because the date is no longer valid-- would developers expect that date to come back to them as empty since it visually looks cleared, or is it okay that the date they get back is the last valid one? This may be a question for our designer on what functionality we expect.
> > The other part of my question was around the two other functions that adopters hook into: `onInputChange` and `onChange`. I think I'm okay with keeping clearing the date specific to `onDateChange`, so lets leave it as is.
> 
> Now I see what case you're talking about. After partial removal the input is cleared but Chosen day is not. As for me, it would be more compatible if the chosen date was updated (cleared) as well. If so, I will update the PR

I want to triple check with our designer who comes back from vacation on Monday, and I'll let you know soon. Thank you! everything else looks great ⭐

OlenaYashyna:
> > > > > I could be convinced either way, but do we think that we should add this clearing logic for `onInputChange` and/or `onChange`? Right now if I manually clear the input using the backspace, we chosen date remains as the last valid date 🤔
> > > > 
> > > > 
> > > > I've tested this case and it works fine. I made a screen record to check, if I missed any case.
> > > > Date.Picker.clearing.the.date.mov
> > > 
> > > 
> > > Yeah this looks great! I was just wondering about the case of partial removal, so backspacing the year only completely clears the input field because the date is no longer valid-- would developers expect that date to come back to them as empty since it visually looks cleared, or is it okay that the date they get back is the last valid one? This may be a question for our designer on what functionality we expect.
> > > The other part of my question was around the two other functions that adopters hook into: `onInputChange` and `onChange`. I think I'm okay with keeping clearing the date specific to `onDateChange`, so lets leave it as is.
> > 
> > 
> > Now I see what case you're talking about. After partial removal the input is cleared but Chosen day is not. As for me, it would be more compatible if the chosen date was updated (cleared) as well. If so, I will update the PR
> 
> I want to triple check with our designer who comes back from vacation on Monday, and I'll let you know soon. Thank you! everything else looks great ⭐

Thank you. So I will wait for Monday with updating

silvalaura (approved):

Issue #1388, opened 2024-08-20
A11y Feedback: Pagination - fix aria-current attribute on current page


orion-cengage:
CENG_1961

https://react-magma.cengage.com/version/4.5.0/api/pagination/

### Issue:
The aria-current attribute is set to "true" to represent the currently selected page in the pagination navigation. This represents the current item within a set, but setting the value to "page" would represent the current page within a set of pages, which is more appropriate in this case.

### Suggestion:
Consider using aria-current="page" instead of aria-current="true" for the selected pagination control.

chris-cedrone-cengage:
_-Verify the use of aria-current_ ✅
_-Verify that aria-label is visible on the nav property_ ✅

All set

Issue #1387, opened 2024-08-20
A11y Fail: Pagination - Need aria-label


orion-cengage:
CENG_1959

https://react-magma.cengage.com/version/4.5.0/api/pagination/

### Issue:
The pagination control uses a `<nav>` element but the element does not have a label to differentiate it from other navigation landmarks.

### Suggestion:
Please use `<nav aria-label="pagination">` to label this landmark.

silvalaura:
We should be able to address this by adding `aria-label` to all our examples (it should by default trickle down to the `nav` tag), without explicitly adding a new prop. Or, we can decide that it's important enough that we add this as a prop that defaults to `pagination` but users can customize. Ideally it would be a required prop but that would be a breaking change, so maybe not at this time.

chris-cedrone-cengage:
-Verify the use of aria-current ✅
-Verify that aria-label is visible on the nav property ✅

All set

chris-cedrone-cengage:
`aria-label="pagination"` ✅

All set

PR #1403, opened 2024-08-26
fix(Pagination): Improved accessibility for classic pagination

silvalaura:
Issue: #1388 #1387 

## What I did

- Updated `aria-current` from `true` to `page`
- Added `aria-label="pagination"`. This can be customized by adopters as well in case there are multiple paginations on the page.

## Screenshots:

image

## Checklist 

## How to test

- Verify the use of `aria-current`
- Verify that `aria-label` is visible on the nav property

### DIFF
--- a/.changeset/a11y-pagination.md
+++ b/.changeset/a11y-pagination.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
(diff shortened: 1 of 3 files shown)


orion-cengage (approved):
I verified the aria-label and current-page attributes match what Perkins suggested.

chris-cedrone-cengage:
-Verify the use of aria-current ✅
-Verify that aria-label is visible on the nav property ✅

All set

Issue #1384, opened 2024-08-20
A11y Fail: Data Grid - check box labels


orion-cengage:
CENG_1958

https://react-magma.cengage.com/version/4.5.0/api/datagrid/#selectable

### Issue:
Labels used for checkboxes inside of tables that are used to select rows do not have meaningful and unique labels (i.e. "Select row undefined of undefined").

### Suggestion:
Ensure that every label is unique (i.e. Select row 2 of 10).

silvalaura:
This has been broken since v2 😓

Issue #1383, opened 2024-08-20
A11y Fail: Data Grid - Pagination fix


orion-cengage:
CENG_1956

Data Grid: https://react-magma.cengage.com/version/4.5.0/api/datagrid/

### Issue
When operating the pagination controls, content updates to indicate the number of rows displaying (i.e. 11-20 of 25). This content can be considered a live region and should be announced to screen readers when it updates.

### Suggestion
Please add aria-live="polite" to this content so that it is announced when it updates. This will give screen reader users an indication that the content in the table updated based on the action that they took.

chris-cedrone-cengage:
Looks good!

PR #1404, opened 2024-08-27
fix(DataGrid): Improved accessibility by adding `aria-live="polite"` and fixing labelText

silvalaura:
Issue: #1384 #1383 

## What I did

- Added `aria-live="polite"`
- Updated selectable datagrid to have unique labels for each checkbox
- Updated checkbox labels to update accordingly from "Select Row" to "Deselect Row" (each individual row and also the indeterminate one)
- Table Aria labels are now part of the i18n interface so they can be customized
- Updated examples in storybook and docs

## Screenshots:

image

## Checklist 

## How to test

- Test all datagrid examples in the docs site
- Test that the table component has not been affected
- Confirm that the appropriate aria labels are in place
- Confirm that the information is communicated correctly for screen readers when the pagination updated, or when items are selected

### DIFF
--- a/.changeset/a11y-datagrid.md
+++ b/.changeset/a11y-datagrid.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(DataGrid): Improved accessibility by adding `aria-live="polite"` and fixing labelText for checkboxes when `isSelectable={true}`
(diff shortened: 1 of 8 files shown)


orion-cengage (approved):
Based on my understanding, I think everything looks good from my side.

Issue #1235, opened 2024-04-11
Toggle Button / Icon only - Lacks accessible name


orion-cengage:
CENG_3317

Docs: https://react-magma.cengage.com/version/4.3.0/api/toggle-button/

Success Criteria: https://www.w3.org/WAI/WCAG21/Understanding/name-role-value

The icon only toggle button has the correct attributes of aria-checked and role but lacks an accessible name. Because of this, users of assistive technology will hear the state of what they are toggling but not the name of it.

The icon only toggle button should have an accessible name added and this can be done in a few different ways. Options include using visually hidden text, an aria-label to name the button, or aria-labelledby to name the button based on text that is elsewhere on the page.

chris-cedrone-cengage:
Verified aria-labels, all set!

PR #1405, opened 2024-08-27
fix(ToggleButton): Improved accessibility for Toggle Button icon only examples

silvalaura:
Issue: #1235 

## What I did

- Updated docs examples to include required aria-label prop for icon only (the logic to support aria label was already present)

## Checklist 

## How to test

- Confirm that all examples in docs and storybook are properly announced by screenreader
- Confirm that all icon only examples have an aria label


orion-cengage (approved):
Looks good from my side.

Issue #1238, opened 2024-04-11
Nav Tabs - indicating current page


orion-cengage:
CENG_3321

Docs: https://react-magma.cengage.com/version/4.3.0/api/nav-tabs/

Success Criteria: https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships

In all of the Nav tab examples, the current page can be determined visually but not programatically. When going through navigations like this, it can be helpful to know if a link is where you currently are.

Use aria-current="page" on the currently active link to indicate the selected state of the component to screen reader users.

silvalaura:
@orion-cengage do we think this is needed for Tabs as well?

chris-cedrone-cengage:
All set!

PR #1406, opened 2024-08-27
fix(NavTabs): Improved accessibility by adding `aria-current="page"` for active tabs.

silvalaura:
Issue: #1238 

## What I did

- Added `aria-current="page"` for active tabs
- Update `children` prop to not be required because they are not required for Icon only nav tabs, which we want to support
- Updated docs examples that were broken
- Added 2 more examples to storybook

Outstanding question: do we need aria-current for our regular tabs as well?

## Checklist 

## How to test

- Confirm new aria-current property for all nav tabs (custom and default)
- Confirm that NavTabs docs site is working as expected

### DIFF
--- a/.changeset/a11y-navTabs.md
+++ b/.changeset/a11y-navTabs.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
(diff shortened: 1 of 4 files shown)


orion-cengage (approved):
Looks good from my side.

Issue #1386, opened 2024-08-20
A11y Fail: Accordion - Lose focus on expand/collapse all


orion-cengage:
CENG_1951

https://react-magma.cengage.com/version/4.5.0/api/accordion/#controlled_accordion

### Issue:
In the Controlled Accordion example, when pressing the "Expand All" button, all of the accordions in the set expand. However, focus is no longer set on anything and the next Tab stop moves to the "Collapse All" button, while the "Expand All" button is in a disabled state. Then when pressing the "Collapse All" button, focus is again not set and the next Tab stop moves to the first accordion button.

### Suggestion:
For this control, consider using one button that changes its label from "Expand All" to "Collapse All". When the button is activated, focus should remain on it (which should be evident via a focus indicator).

silvalaura:
This should update both of these examples:
- https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/accordion--expand-collapse-all storybook 
- https://next--upbeat-sinoussi-f675aa.netlify.app/api/accordion/#controlled_accordion example

silvalaura:
Looks good ✅

PR #1422, opened 2024-09-11
fix(Accordion): Fix focus issue with  and  buttons

artemtrusov-cengage:
Issue: #1386

## What I did

Changes were made to both the Storybook and the documentation. Specifically, the "Collapse All" and "Expand All" buttons were removed and replaced with a single universal toggle button, ensuring that focus is not lost.

## Checklist 

### DIFF
--- a/.changeset/a11y-accordion.md
+++ b/.changeset/a11y-accordion.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Accordion): Fix focus issue with `Collapse All` and `Expand All` buttons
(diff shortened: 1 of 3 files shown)


.changeset/a11y-accordion.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': patch
```
silvalaura, line 2: Although technically we are updating react-magma-dom, this is a documentation only change, so let's make this `'react-magma-docs': patch`. The only reason I'm recommending this is because, if it's labeled under dom, it will become part of our user facing changelog when there's no actual change to the component. Otherwise, the change looks great!

Issue #1247, opened 2024-04-12
Breadcrumb - Current page does not receive focus


orion-cengage:
CENG_3296

Docs: https://react-magma.cengage.com/version/4.3.0/api/breadcrumb/

Success Criteria: https://www.w3.org/WAI/WCAG21/Understanding/name-role-value

Within the breadcrumb component, the current page is rendered as a `span` element. This element does not receive focus by default so assistive technology users may not reach it or understand it if they do.

Rather than using a span for the current page, use a link element with `href=""` and `aria-current="page"`.

orion-cengage:
I reviewed the examples in the docs site and this looks great.

PR #1429, opened 2024-09-12
fix(BreadCrumb): replace span with link to fix focus

moathabuhamad-cengage:
Issue: #1247

## What I did
- Bug fix replace span with link to fix focus

## Screenshots:
image

## Checklist 

## How to test
1. Ensure each BreadcrumbItem is clickable when the to prop is provided.
2. Verify the last BreadcrumbItem is non-clickable, has aria-current="page", and can be focused.
3. Confirm the correct styles and focus behavior for both clickable and non-clickable items.

### DIFF
--- a/.changeset/a111y-breadCrumbs.md
+++ b/.changeset/a111y-breadCrumbs.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(BreadCrumb): replace span with link to fix focus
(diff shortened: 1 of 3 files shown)


silvalaura (approved):

orion-cengage (changes requested):
Please make sure the current page focus is using the correct color like the links that come before it in the breadcrumb. This issue is only on the inverse version.

image

image

Issue #1396, opened 2024-08-21
A11y Fail: Date Picker - Need different keyboard command for Help


orion-cengage:
CENG_1509

https://react-magma.cengage.com/version/4.5.0/api/date-pickers/

### Issue:
The Calendar Help modal dialog provides instructions for using various key commands (such as arrows, Enter, Page Down, etc.). One of the key commands is a question mark (?), which can be used to open the Calendar Help modal. This is considered a single printable character key shortcut which can create problems for speech input users and for keyboard users who are prone to accidentally hit keys.

### Acceptance Criteria
- Help panel keyboard shortcut has been removed, so `?` should no longer trigger it
- The shortcut description is removed from the help panel list
- i18n variables and unit tests are cleaned up

### Suggestion:
~Change keyboard shortcut for the keyboard instructions to `ctrl+shift+?`, and update how and where we teach users this shortcut.~

~Edit list of keyboard instructions:~

![image](https://github.com/user-attachments/assets/cb34fcb5-ede1-4459-be64-aeee426f43fe)

~Add to Tooltip on button:~

![image](https://github.com/user-attachments/assets/36b217e0-dafa-42ea-a66b-02cc9b8f7fcc)

silvalaura:
Ensure that we do sufficient cross browser testing when implementing this

chris-cedrone-cengage:
Browsers

chris-cedrone-cengage:
Browsers:
_**Safari 18
Chrome 129
Firefox 131**_

Help panel keyboard shortcut has been removed, so ? should no longer trigger it ✅
The shortcut description is removed from the help panel list ✅
i18n variables and unit tests are cleaned up ✅

PR #1430, opened 2024-09-12
fix(DatePicker): Remove the shortcut to open the `HelperInformation` component.

nikitaorliak-cengage:
Issue: #1396

## What I did

Change the shortcut on the DatePicker to open `HelperInformation` from `?` to `Ctrl+Shift+?`. Add a tooltip for it. Write test. Change localization.

## Screenshots:

image
image

## Checklist 

### DIFF
--- a/.changeset/a11y-datePicker-shortcut.md
+++ b/.changeset/a11y-datePicker-shortcut.md
@@ -2,4 +2,4 @@
 'react-magma-dom': patch
 ---
 
-fix(DatePicker): Change shortcut to open `HelperInformation` from `?` to `Ctrl+Shift+?`
+fix(DatePicker): Remove shortcut to open `HelperInformation`.
--- a/.changeset/a11y-datePicker-shortcut.md
+++ b/.changeset/a11y-datePicker-shortcut.md
@@ -1,0 +1,5 @@
+---
+'react-magma-docs': patch
+---
+
+fix(DatePicker): Change shortcut to open `HelperInformation` from `?` to `Ctrl+Shift+?`
--- a/.changeset/a11y-datePicker-shortcut.md
+++ b/.changeset/a11y-datePicker-shortcut.md
@@ -1,5 +1,5 @@
 ---
-'react-magma-docs': patch
+'react-magma-dom': patch
 ---
 
 fix(DatePicker): Change shortcut to open `HelperInformation` from `?` to `Ctrl+Shift+?`
--- a/packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
+++ b/packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
@@ -195,8 +195,8 @@ export const CalendarMonth: React.FunctionComponent<CalendarMonthProps> = (
               </tbody>
             </Table>
             <Tooltip
-              content={'Keyboard instructions\nctrl+shift+?'}
-              tooltipStyle={{ position: 'fixed', whiteSpace: 'pre-line' }}
+              content={'Keyboard instructions'}
+              tooltipStyle={{ position: 'fixed' }}
             >
               <HelperButton theme={theme}>
                 <IconButton
--- a/packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
+++ b/packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
@@ -21,7 +21,6 @@ interface CalendarMonthProps {
   setDateFocused?: (value: boolean) => void;
 }
 
-
 const CalendarContainer = styled.div<{ isInverse?: boolean }>`
   background: ${props =>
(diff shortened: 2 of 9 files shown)


silvalaura:
The panel opens with the new command on Safari and Firefox, but not on Chrome 🤔

silvalaura:
> > The panel opens with the new command on Safari and Firefox, but not on Chrome 🤔
> 
> Sounds weird. I've tried using different keyboards and languages ​​in Chrome and Chromium and it works for me. Could you please double-check it?

I confirmed that on my PC, it works for all browsers (Chrome, Firefox and Edge). On my Mac, it does not work on Chrome but does on Firefox and Safari :/ @orion-cengage or @chris-cedrone-cengage , can you guys check if this works on your Macs?

packages/react-magma-dom/src/i18n/default.ts
```diff
@@ -144,8 +144,9 @@ export const defaultI18n: I18nInterface = {
         displayValue: 'ESC',
         explanation: 'Return to the date input field.',
       },
-      questionMark: {
```
silvalaura, line 147: changing this would be a breaking change for our adopters if they are already using `questionMark`, so maybe we should leave it as `questionMark` even if it's not 100% accurate

packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
```diff
@@ -196,8 +195,8 @@ export const CalendarMonth: React.FunctionComponent<CalendarMonthProps> = (
               </tbody>
             </Table>
             <Tooltip
-              content={'Keyboard instructions'}
-              tooltipStyle={{ position: 'fixed' }}
+              content={'Keyboard instructions\nctrl+shift+?'}
```
silvalaura, line 198: @orion-cengage Do you have a preference between lowercase or uppercase for the tooltip?

.changeset/a11y-datePicker-shortcut.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-docs': patch
```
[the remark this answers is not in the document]
nikitaorliak-cengage: I changed docs in `Internationalization` section a bit, that's why I thought that it should be `docs`. Got it, thanks!

packages/react-magma-dom/src/i18n/default.ts
nikitaorliak-cengage: Oh, okay. I thought I'd better clarify that and change from `questionMark` to `ctrlShiftAndQuestionMark` everywhere in the code. Changed it back to `questionMark` :)

packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
nikitaorliak-cengage: I used lowercase because saw them in the issue:

https://github.com/cengage/react-magma/issues/1396

image

orion-cengage: Lower case is great

orion-cengage:
> > > The panel opens with the new command on Safari and Firefox, but not on Chrome 🤔
> > 
> > 
> > Sounds weird. I've tried using different keyboards and languages ​​in Chrome and Chromium and it works for me. Could you please double-check it?
> 
> I confirmed that on my PC, it works for all browsers (Chrome, Firefox and Edge). On my Mac, it does not work on Chrome but does on Firefox and Safari :/ @orion-cengage or @chris-cedrone-cengage , can you guys check if this works on your Macs?

Nope, the new keyboard command is not working for me in Chrome on my Mac, but it is working in Firefox.

nikitaorliak-cengage:
> > > > The panel opens with the new command on Safari and Firefox, but not on Chrome 🤔
> > > 
> > > 
> > > Sounds weird. I've tried using different keyboards and languages ​​in Chrome and Chromium and it works for me. Could you please double-check it?
> > 
> > 
> > I confirmed that on my PC, it works for all browsers (Chrome, Firefox and Edge). On my Mac, it does not work on Chrome but does on Firefox and Safari :/ @orion-cengage or @chris-cedrone-cengage , can you guys check if this works on your Macs?
> 
> Nope, the new keyboard command is not working for me in Chrome on my Mac, but it is working in Firefox.

Yeap, I tried a different OS, and only Chrome on Macs has this problem. I think it happens because Mac has his own shortcuts in Chrome and doesn't allow us to use ours :/

silvalaura (approved):
Looks good. Lets just update the PR title and description before merging

silvalaura:
Lets hold off on merging for another day, as I will be releasing our next version first. I will merge this when we're ready.

orion-cengage (approved):
Update looks good to me. I confirmed the shortcut is gone.

Issue #1390, opened 2024-08-21
A11y Fail: Select - Clearable tab and focus issue


orion-cengage:
CENG_1964

https://react-magma.cengage.com/version/4.5.0/api/select/#clearable

### Issue:
The clear button is nested inside another interactive control with role="button". Focusable elements with an interactive control ancestor (any element that accepts user input such as button or anchor elements) may not be announced by screen readers and may create an empty tab stop. That is, you could tab to the element but the screen reader will not announce its name, role, or state.

I think also results in us getting two focus states at once which is a problem.

![image](https://github.com/user-attachments/assets/65ccea00-84f4-4ee5-ae49-1c7af781202e)

### Suggestion:
Please place the clear button outside of the button that operates the listbox. It can still be positioned visually where it is currently but should not be nested inside the button in the DOM.

silvalaura:
This could affect:
- Select
- Input
- Dropdown
- DatePicker

silvalaura:
With Mac Safari 17.2.1 / VoiceOver the clear button label is not announced when it gains focus:
https://github.com/user-attachments/assets/10b49b93-47fa-4ef8-8954-92a0e86b026e

It also results in errors when scanning the page with axe-core
![Image](https://github.com/user-attachments/assets/3369fc1a-5047-43b4-b48e-f17bf3b55e07)

orion-cengage:
I tested all three impacted inputs in Storybook and they're working perfectly. This one looks good to go.

PR #1434, opened 2024-09-17
fix(Select): Fix accessibility issue. 

nikitaorliak-cengage:
Issue: #1390

## What I did

Remove the `ClearIndicator` component outside from the `SelectTriggerButton`.

## Checklist 

### DIFF
--- a/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
+++ b/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
@@ -271,7 +271,7 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
           testId="clearIndicator"
           style={{
             position: 'absolute',
-            right: '2.75em',
+            right: '3.25em',
             top: '50%',
             transform: 'translateY(-50%)'
           }}
--- a/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
+++ b/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { MultiSelectProps } from '.';
+import { instanceOfDefaultItemObject, MultiSelectProps } from '.';
 import { useSelect, useMultipleSelection } from 'downshift';
 import { CloseIcon } from 'react-magma-icons';
 import { ItemsList } from './ItemsList';
@@ -9,6 +9,8 @@ import { SelectText, SelectedItemButton, IconWrapper } from './shared';
 
 import { ThemeContext } from '../../theme/ThemeContext';
 import { I18nContext } from '../../i18n';
+import { ButtonSize, ButtonVariant } from '../Button';
+import { defaultComponents } from './components';
 
 export function MultiSelect<T>(props: MultiSelectProps<T>) {
   const {
@@ -39,6 +41,7 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
     onKeyUp,
     onRemoveSelectedItem,
     placeholder,
+    isClearable,
   } = props;
 
   function checkSelectedItemValidity(itemToCheck: T) {
@@ -54,6 +57,7 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
     removeSelectedItem,
     selectedItems,
     setActiveIndex,
+    reset,
   } = useMultipleSelection<T>({
     ...props,
     ...(props.initialSelectedItems && {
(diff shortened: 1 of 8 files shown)


silvalaura:
This looks good, and I verified the screenreader reads it on Safari.

However, because we are positioning it absolutely, it breaks in the examples with `labelPosition="left"` or when `isLabelVisuallyHidden` is true:
image
image
image

orion-cengage:
We need to stop the clear button from overlapping the text and basically disappearing.

image

orion-cengage:
Not sure when this became an issue, but the regular Select is not supposed to grow vertically just because an item is long and wraps.

image

If anything, we should truncate it with ellipsis like this:

image

orion-cengage:
The positioning of the clear button is off when it is in the error state or when showing helper text.

image

image

orion-cengage:
The inverse version of the input isn't using the inverse version of the clear button so it is hard to see.

image

orion-cengage (changes requested):
I added a number of styling issues.

orion-cengage (changes requested):
The latest update fixed the previous styling issues but has broken the focus style on the main container of the Select input. If we can fix this, then we might be good to go.

image

image

image

silvalaura:
It's so close! 
- Focus is back ✅ 
- Text overflow is properly cut off ✅ 
- X is properly aligned for Combobox with left label ✅
- - X is properly aligned for Combobox with top label ✅ 
- X is properly aligned for Select with left label ✅ 
- X is properly aligned for Select with top label ❌ 
- Is inverse looks good ✅

silvalaura:
One more:
if Select `isMulti` is set to true, the isClearable button disappears. (it is visible for Combobox `isMulti`)
image

All other use cases look good to me.

packages/react-magma-dom/src/components/Select/MultiSelect.tsx
```diff
@@ -159,6 +163,44 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
...
+        allItems.push(item.label);
+      }
+    });
+
+    return allItems.join(', ');
+  }
+
+  const toggleButtonRef = React.useRef<HTMLButtonElement>();
+
+  const clearIndicatori18n =
```
silvalaura, line 185: In this multiselect file, wouldn't we always want to use the multi label?

silvalaura (commented):
**Select:**
- Focus around the entire input in Select ✅ 
- Focus is only on isClearable (no double focus) ✅
- Text overflow is properly cut off ⚠️ **This looks good, but would be even better if we could add like 8px of space between the ellipsis and the clear button so they never overlap at all**
image
- X is properly aligned for Select with left label ✅
- X is properly aligned for Select with top label ✅
- X is properly aligned for Multi Select with left label ⚠️ **How hard would it be to align to the center? It's center aligned for Combobox**
image
- X is properly aligned for Multi Select with top label ⚠️ **It can get in this weird state where the X is over the text. Maybe if we add the space noted above, this could look even better**
image 
- isInverse uses correct colors for Select ✅
- Screenreader on Safari reads isClearable ✅
- Screenreader on Chrome reads isClearable ✅

**Combobox:**
- Focus around the entire input in Combobox ✅
- Focus is only on isClearable ✅
- Text overflow is properly cut off ✅
- X is properly aligned for Combobox with left label ✅
- X is properly aligned for Combobox with top label ✅
- X is properly aligned for Multi Combobox with left label ✅
- X is properly aligned for Multi Combobox with top label ✅
- isInverse uses correct colors for Combobox ✅
- Screenreader on Safari reads isClearable ✅
- Screenreader on Chrome reads isClearable ✅

silvalaura:
We're almost there!

**Select:**
- Focus around the entire input in Select ✅ 
- Focus is only on isClearable (no double focus) ⚠️ **can it be moved _slightly_ more to the left to avoid the focus touching the arrow?**
image
- Text overflow is properly cut off ✅
- X is properly aligned for Select with left label ✅
- X is properly aligned for Select with top label ✅
- X is properly aligned for Multi Select with left label ✅
- X is properly aligned for Multi Select with top label ✅
- isInverse uses correct colors for Select ✅
- Screenreader on Safari reads isClearable ✅
- Screenreader on Chrome reads isClearable ✅

**Combobox:**
- Focus around the entire input in Combobox ✅
- Focus is only on isClearable ✅
- Text overflow is properly cut off ✅  I noticed the issue below, but I can see it's an existing issue, so it does not need to be addressed at this time. 
image
image
- X is properly aligned for Combobox with left label ✅
- X is properly aligned for Combobox with top label ✅
- X is properly aligned for Multi Combobox with left label ✅
- X is properly aligned for Multi Combobox with top label ✅
- isInverse uses correct colors for Combobox ✅
- Screenreader on Safari reads isClearable ✅
- Screenreader on Chrome reads isClearable ✅

orion-cengage (approved):

silvalaura (approved):

PR #1444, opened 2024-09-19
Build > dev patch > update

chris-cedrone-cengage:
Merging the 4.6.0 release into dev-patch


silvalaura:
We don't want 4.6.0 into dev-patch. We want 3.9.0 to avoid bringing in all the version 4 changes (package updates)

PR #1448, opened 2024-09-19
fix(Input): Remove flex-shrink to allow inputs to shrink as needed

silvalaura:
Issue: none

## What I did

Removed flex-shrink. We noticed that DatePicker had a min-width of 320px because of this, and it isn't needed.

This would affect:
- Input
- NativeSelect
- Select
- Textarea
- TimePicker
- DatePicker
- PasswordInput
- Search
- Combobox

## Checklist 
- [-] I have made corresponding changes to the documentation

## How to test

Confirm that all of the components above continue to work as expected in all screen sizes and browsers.

### DIFF
--- a/.changeset/inputwrapperstyles.md
+++ b/.changeset/inputwrapperstyles.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Input): Remove flex-shrink to allow inputs to shrink as needed.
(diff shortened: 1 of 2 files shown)


chris-cedrone-cengage:
I doubt we care about this, but I noticed that placeholder text doesn't get an ellipsis at narrow widths:

Placeholder

silvalaura:
@chris-cedrone-cengage good point. I think this is okay because it's the same behavior as before this change was introduced.

chris-cedrone-cengage:
**Verified the following components are appearing as described:**
Input ✅
NativeSelect ✅
Select ✅
Textarea ✅
TimePicker ✅
DatePicker ✅
PasswordInput ✅
Search ✅
Combobox ✅

**Tested on:** 
https://storybook-preview-1448--upbeat-sinoussi-f675aa.netlify.app/
➕
https://docs-preview-1448--upbeat-sinoussi-f675aa.netlify.app/

Issue #1233, opened 2024-04-11
Search input - Loading - Fix disappearing focus


orion-cengage:
CENG_3312

Docs: https://react-magma.cengage.com/version/4.3.0/api/search/

Success Criteria: https://www.w3.org/WAI/WCAG21/Understanding/focus-order

For the Loading example, when the search button is focused on and then pressed the button is then hidden and focus is moved to the body. When this happens it could confuse a user of assistive technology.

Perkins recommendation: If the search button is removed, focus should be programmatically moved somewhere logical or the button could be left where it is while the loading happens.

Solution: Let's just stop hiding the button and keep focus on it.

orion-cengage:
Tested this on the dev docs site. The Search button correctly shows the loading animation upon clicking, and then the focus remains on it even after it is finished. You never lose site of focus. This appears good to go for me.

PR #1450, opened 2024-09-24
fix(search): Fix losing focus during loading

artemtrusov-cengage:
Issue: #1233

## What I did

Changes were made to ensure that focus is maintained during loading by replacing the search icon with a spinner component. A condition was also added to prevent the button from being clickable while loading

## Screenshots:

Знімок екрана 2024-09-24 о 13 26 17

## Checklist 

### DIFF
--- a/.changeset/a11y-search-loading.md
+++ b/.changeset/a11y-search-loading.md
@@ -1,5 +1,5 @@
 ---
-'react-magma-docs': patch
+'react-magma-dom': patch
 ---
 
 fix(search):Fix the disappearing focus on the button when loading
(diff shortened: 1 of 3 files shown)


silvalaura (commented):
The button still looks clickable while it's loading. I wonder if we should disable it?? I'll leave it up to @orion-cengage here

orion-cengage (approved):
In this case I'm in favor of not disabling the button while it is loading. Keeping the button clickable seems less problematic than putting focus on a non-interactive element. I like how you can clearly see the focus and even move it while it loads. I also like moving the loading animation away from the left side, which was nowhere near the button you clicked.

Consider this UX-approved.

silvalaura (approved):

Issue #1439, opened 2024-09-18
A11y: Dropdown - active item index issue


orion-cengage:
CENG_1486

https://react-magma.cengage.com/version/4.5.0/api/dropdown/#active_item_index

![image](https://github.com/user-attachments/assets/a561bb6e-be9c-4baf-a6b3-2e0c1c87f3c3)

**Issue description:**
The Active Item Index menu has a selected option that is indicated with a checkmark. The selected state of this option is not communicated programmatically.

**Solution:**
Use aria-selected="true" on any option that is selected within this menu.

chris-cedrone-cengage:
`aria-current="true"` appears accordingly in the example properly within:

PR #1455, opened 2024-09-25
fix(dropdown): Fix active index communication handled programmatically

artemtrusov-cengage:
Issue: #1439

## What I did

The `menuitem` role does not support `aria-selected`, so as an alternative, `aria-checked` was added to facilitate programmatic communication of the active item and to improve accessibility.

## Checklist 

### DIFF
--- a/.changeset/a11y-dropdown-active-item.md
+++ b/.changeset/a11y-dropdown-active-item.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(dropdown): Fix active index communication handled programmatically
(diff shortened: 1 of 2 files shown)


silvalaura:
This looks good to me, but I am triple checking with our accessibility team to make sure this is the right solution. Lets hold off on merging until we get confirmation.

silvalaura (approved):
Confirmed by our a11y expert:

> I looked into the ARIA authoring practices and it looks like aria-current="true" is the way to go to indicate the selected option for that type of widget.

PR #1459, opened 2024-09-26
docs(TreeView): Update examples

silvalaura:
Issue: none

## What I did

initialExpandedItems example was using the wrong ID so the items were not expanding as expected.

## Checklist 

## How to test
Confirm all examples in tree view docs behave as expected


chris-cedrone-cengage:
Doc site example looks good with the proper initial view expanded:

initialExpandedItems example was using the wrong ID so the items were not expanding as expected. ✅

Issue #1064, opened 2023-05-15
Review Dependencies and ensure they are up to date


silvalaura:
_From Slack_
> There are a bunch of peer dependencies that are also not current (emotion, framer-motion, downshift) that are explicitly tied to older versions. Fortunately we don't already use those packages so we aren't in a position of needing to downgrade them, but I could see a situation where we'd be in a dependency bind if we did need them at the latest version for some reason or another.

From this ticket we will have a plan for how to upgrade our dependencies. We will create a list of what needs to be updated and their dependencies and will create new tickets to address separately when possible.

moathabuhamad-cengage:
we have 84 major (Potentially breaking)  dependencies 36 minor and patch updates  #1484 should cover the majority of minor and patch updates we still have the major ones left to do

| **Dependency Updates**               | **Current Version**  | **Updated Version**  | **Type**                               |
|--------------------------------------|----------------------|----------------------|----------------------------------------|
| @commitlint/cli                      | ^11.0.0              | ^19.5.0              | Major (Potentially breaking)          |
| @commitlint/config-conventional      | ^11.0.0              | ^19.5.0              | Major (Potentially breaking)          |
| @commitlint/is-ignored               | ^12.1.4              | ^19.5.0              | Major (Potentially breaking)          |
| @commitlint/load                     | ^12.1.4              | ^19.5.0              | Major (Potentially breaking)          |
| @commitlint/parse                    | ^12.1.4              | ^19.5.0              | Major (Potentially breaking)          |
| @commitlint/rules                    | ^12.1.4              | ^19.5.0              | Major (Potentially breaking)          |
| @commitlint/top-level                | ^12.1.4              | ^19.5.0              | Major (Potentially breaking)          |
| @storybook/addon-a11y               | ^6.5.16              | ^8.3.5               | Major (Potentially breaking)          |
| @storybook/addon-actions             | ^6.5.16              | ^8.3.5               | Major (Potentially breaking)          |
| @storybook/addon-controls            | ^6.5.16              | ^8.3.5               | Major (Potentially breaking)          |
| @storybook/addon-docs                | ^6.5.16              | ^8.3.5               | Major (Potentially breaking)          |
| @storybook/addon-essentials          | ^6.5.16              | ^8.3.5               | Major (Potentially breaking)          |
| @storybook/addon-measure             | ^6.5.16              | ^8.3.5               | Major (Potentially breaking)          |
[...]

PR #1464, opened 2024-09-30
chore(Lerna): Update Lerna and add local caching for build and test commands

moathabuhamad-cengage:
Issue: #1064

## What I did
- Upgrade Lerna version.
- Move to npm workspaces in place of Lerna Bootstrap.
- Add local caching for build and test commands

Screenshots:
Cool new graph with `npx nx graph`
graph

## Checklist 

## How to test
verify that all apps build and test as expected


.changeset/build-upgrade-lerna.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': patch
```
silvalaura, line 2: I know it isn't technically a docs change, but in order for our build to pass it needs a package here. If we use `docs`, our adopters won't see it in the changelog list which they don't need

packages/react-magma-dom/scripts/generateJSON.js
```diff
@@ -156,7 +156,7 @@ const filterJson = () => {
         },
       };
     }
-    console.log(child);
+    // console.log(child);
```
silvalaura, line 159: We can probably just delete this line

silvalaura (approved):

Issue #1241, opened 2024-04-11
Drawer - Focus not trapped properly when open


orion-cengage:
CENG_3326

Docs: https://react-magma.cengage.com/version/4.3.0/api/drawer/

Success Criteria: https://www.w3.org/WAI/WCAG21/Understanding/focus-order

In all drawers, the keyboard focus is trapped within the modal in both directions *except* for the navigation drawer,  Here, the focus is trapped when tabbing forward, but when tabbing backward (Shift+Tab) the focus moves into the page behind the drawer.

(Note that the inverse view of the navigation drawer *does* trap the focus in both directions.)

Ensure that the keyboard focus remains trapped within the navigation drawer when users press Shift+Tab to move the focus backward.

silvalaura:
Wonder if this is related to https://github.com/cengage/react-magma/issues/970 in some way?

silvalaura:
Confirmed that all examples in docs site and storybook trap the focus as expected ✅

Issue #970, opened 2022-11-21
Modal/Drawer: Focus lock breaks with specific content


chris-cedrone-cengage:
**Describe the bug**
With certain content in `Modal`, such as `NavTabs`, our focus lock utility doesn't trap keyboard navigation within the `Modal`. This is exemplified in our `Drawer` component which uses `Modal` and is reproducible when using it as a navigational element.

**To Reproduce**
1. Go to our documentation site.
2. Navigate to Components > `Drawer`
3. Click "Show Drawer" under "Navigation Example"
4. Tab through the opened `Drawer`

**Expected behavior**
The focus lock should allow the user to tab through the UI while staying within `Drawer`.

**Additional context**
It seems that the `focusTrapElement` variable in `Modal.tsx` causes the issue. An update made to `Modal` about a year ago regarding portals also caused animations to break and other oddities. This may have something to do with it in addition.

ccedrone:
This [PR](https://github.com/cengage/react-magma/pull/432) may have something to do with it.

PR #1467, opened 2024-10-01
fix(drawer): Fix trapped focus in the drawer

artemtrusov-cengage:
Issue: #1241

## What I did

A filtering function was added to the `useFocusLock` hook, which filters out invisible elements that were being included in the `focusableItems.current` array and preventing the focus from being properly locked on the modal window.

## Checklist 

## How to test

1) Go to our documentation site.
2) Navigate to Components > Drawer
3) Click "Show Drawer" under "Navigation Example"
4) Tab through the opened Drawer

### DIFF
--- a/.changeset/a11y-drawer-navigation-focus
+++ b/.changeset/a11y-drawer-navigation-focus
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(drawer): Fixed trapped focus in the drawer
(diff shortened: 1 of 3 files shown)


silvalaura:
This solution also addresses #970

.changeset/a11y-drawer-navigation-focus
silvalaura, line 1: We just need to add `.md` to the filename so the changeset is detected

Issue #1163, opened 2023-11-15
Update Input component to reposition "Additional Children" when labels are moved/removed


orion-cengage:
Purpose: We need to update the positioning of the [Additional Children](https://react-magma.cengage.com/version/4.6.0/api/input/#additional_children_(e.g._help_link)) when the label is moved or removed.

**Figma:** https://www.figma.com/design/VoWYiiOKKppcYm5QLgPyDC/Components---Inputs?m=auto&node-id=2113-2766&t=CEU5gyO21XoerN9F-1

![Image](https://github.com/user-attachments/assets/b7de8e26-a029-4b7c-8a35-1782c983c530)

- when label position is top, children will be top right
- when label position is left, children will be to the right of the select
- when label is visually hidden, children will be to the right of the select

We should also update the title in the docs: instead of calling it `Additional Children (e.g. Help Link)`, lets rename it to `Help Link`. We should add an example of the label being on the left, and one where the label is visually hidden.

chris-cedrone-cengage:
Update Input additional children position based on label position ✅
Check Input with help link behavior based on label position and visibility ✅

All verified on Chrome Version 130.0.6723.70, Safari 18, and Firefox 132.0 using Mac OS Sonoma 14.7 on:
https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/datepicker--default

silvalaura:
**Component issues:**
- When using labelPosition top, there is still space being left on the right size of the input, and it should be full width
![image](https://github.com/user-attachments/assets/0883a384-b110-41a7-b3f2-a30e6617abc3)

- Help link - left label: is it okay that the label is centered aligned and looks like this with a long label?
- Help link - top label: is this the expected result?
![image](https://github.com/user-attachments/assets/710140f0-c139-4100-aaf1-3ac93adc474e)


**input.mdx issues:**
- `type={InputType.button}` was passed to IconButton, and it doesn't need a type at all
- Spacer is missing as an import
- size, variant and labelPosition should take an enum instead of a string

![image](https://github.com/user-attachments/assets/25383137-5d7b-41ab-b5c5-cba481baff38)


https://codesandbox.io/p/sandbox/testing-1163-help-link-input-nwgnrt?file=%2Fexample.tsx

orion-cengage:
[...]

PR #1478, opened 2024-10-07
Update Input additional children position based on lable position

moathabuhamad-cengage:
Issue: #1163 

## What I did
Update Input additional children position based on lable position

## Screenshots:
image

## Checklist 

## How to test
check Input with help link behavior based on label position and visibility

### DIFF
--- a/.changeset/feat-input-help-text.md
+++ b/.changeset/feat-input-help-text.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': minor
+---
+
+feat(Input): Update Input additional children position based on lable position
--- a/.changeset/feat-input-help-text.md
+++ b/.changeset/feat-input-help-text.md
@@ -2,4 +2,4 @@
 'react-magma-dom': minor
 ---
 
-feat(Input): Update Input additional children position based on lable position
+feat(Input): Update Input additional children position based on label position
--- a/packages/react-magma-dom/src/components/InputBase/index.tsx
+++ b/packages/react-magma-dom/src/components/InputBase/index.tsx
@@ -3,7 +3,6 @@ import styled from '@emotion/styled';
 import { ReferenceType } from '@floating-ui/react-dom/dist/floating-ui.react-dom';
 import { transparentize } from 'polished';
 import * as React from 'react';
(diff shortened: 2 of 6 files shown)


silvalaura (approved):
Looks good to me. Let's wait for @orion-cengage's approval before merging.

orion-cengage (changes requested):
The button appears to be the correct size, but the icon inside is too large (24px). It needs to be smaller like it previously was (20px). I think it has taken on the size it was when it was previously INSIDE the input.

image

orion-cengage:
Can we please expand the examples shown on the docs site to show the three variants: Top label, left label, no label.

Also, the Select component has the same ability to show a Help button, although I think it is technically done a little differently. We should at least update the title of that section of the page to also be "Help Link."

moathabuhamad-cengage:
> The button appears to be the correct size, but the icon inside is too large (24px). It needs to be smaller like it previously was (20px). I think it has taken on the size it was when it was previously INSIDE the input.
>


its currently set to behave like the IconButton which is the `theme.iconSizes.large` 32px on large inputs and `theme.iconSizes.medium` 24px on normal ones.  

image

@orion-cengage  whats the correct behavior ? always 20px ?

orion-cengage (approved):

packages/react-magma-dom/src/components/InputBase/index.tsx
```diff
@@ -407,6 +418,24 @@ function getIconButtonSVGSize(props) {
   return `${theme.iconSizes.medium}px`;
 }
 
+function getHelpeLinkSVGSize(props) {
```
silvalaura, line 421: `getHelpLinkSVGSize` or `getHelperLinkSVGSize`

.changeset/feat-input-help-text.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': minor
+---
+
+feat(Input): Update Input additional children position based on lable position
```
silvalaura, line 5:
```suggestion
feat(Input): Update Input additional children position based on label position
```

Issue #1122, opened 2023-09-05
a11y: Checkbox labels not read in Firefox + NVDA


silvalaura:
**Describe the bug**
The screen reader does not read the checkbox label. This is only reproduced for this combination of Firefox 115/116/117 + NVDA + Windows 10/11. All other browsers and screen readers work fine.

**To Reproduce**

**Expected behavior**
Screenreader is reads checkbox label

**Screenshots**
![firefox_sandbox](https://github.com/cengage/react-magma/assets/91160746/429c06f3-bf52-4445-80fc-b12f4d755c8b)
![chrome_reproducing_issue](https://github.com/cengage/react-magma/assets/91160746/8f4a7a5b-b766-4eff-8625-61b3a8554b3a)
![firefox_reproducing_issue](https://github.com/cengage/react-magma/assets/91160746/b1849dd1-9dc7-4d40-8964-5247b0bac71f)

silvalaura:
I don't have that version of Firefox, but confirmed that labels are read on the latest FF with NVDA.

PR #1479, opened 2024-10-08
fix(CheckBox): Fix for unreadable labels in `CheckBox` for screen readers.

nikitaorliak-cengage:
Issue: #1122

## What I did

Fix for unreadable labels in `CheckBox` for screen readers.

## Checklist 

### DIFF
--- a/.changeset/a11y-checkBox-labels-not-read.md
+++ b/.changeset/a11y-checkBox-labels-not-read.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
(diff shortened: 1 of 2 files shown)


silvalaura:
Were you able to replicate the bug on your machine before fixing it? I have FF v130 on my PC and if I use NVDA on our development checkbox storybook, all the labels are read as expected (with and without this change). Wondering if we really need this, or this has been fixed on the Firefox side since the ticket was reported a year ago.

silvalaura (approved):

Issue #1080, opened 2023-06-08
Date picker - Keyboard nav issues when NVDA is turned on


orion-cengage:
**Describe the bug**
Reported by David Charters

When trying to navigate the date picker with NVDA turned on he is getting the following problems:

1. Can't navigate to the keyboard help button
2. page up/down home/end are not working

Here's a screenshot of his NVDA settings

![image](https://github.com/cengage/react-magma/assets/24397282/882fa8df-df55-4d36-a3f1-bbdf1756bb7a)

silvalaura:
This is reproducible: 

https://github.com/cengage/react-magma/assets/91160746/5cfb773d-07c8-4e5c-98fb-e3eaea25be06

silvalaura:
@chris-cedrone-cengage This ended up being a whole DatePicker navigation and accessibility overhaul, not exclusive to PC. I have confirmed it works on NVDA, so please confirm the use cases noted in the PR on Mac/VoiceOver.

chris-cedrone-cengage:
Open DatePicker -> Open HelperInformation -> Press Tab several times -> Focus inside HelperInformation. ✅

Open DatePicker -> Open HelperInformation -> Return back to DatePicker -> Press Tab several times -> Focus inside DatePicker. ✅

Button X (Close HelperInformation) in HelperInformation has the same behavior as the Return back to Calendar button. ✅

Open DatePicker -> Open HelperInformation -> Press X -> Closed DatePicker. ✅

Open DatePicker -> Press Tab several times -> Focus on dates. ✅

Open DatePicker with inactive days (minDate) -> Select the day right after inactive -> Press ArrowUp or ArrowLeft several times -> No need to press several times ArrowDown or ArrowRight to come back to the focused day. The focus shows the exact position selected. ✅

I think we should add showing the focus on the Keyboard Icon so when a user opens the help panel, then clicks back to calendar, the focus is visible. ✅

**All verified with VoiceOver using Mac OS Sonoma 14.7**

chris-cedrone-cengage:
Safari 18 Shows the following behavior:

Date picker -> Tabbing closes Date picker window at the end of focusable elements instead of looping back through them

Date picker -> Keyboard instructions -> Tabbing loses focus behind date picker

silvalaura:
I can't reproduce the issue above, but 4 people can, so this needs to be reviewed and addressed.

silvalaura:
@chris-cedrone-cengage 
[...]

PR #1487, opened 2024-10-10
fix(DatePicker): Fix navigation and accessibility issues.

nikitaorliak-cengage:
Issue: #1080

## What I did

This problem is related to the operation of the NVDA screen reader. Users with this screen reader must change the NVDA mode to `focus` to properly interact with the calendar. Added additional text to the `calendarOpenAnnounce` label.
Fix navigation and accessibility issues.

## Checklist 

#### 1. Focus is not locked in `DatePicker` and `HelperInformation`
**Previous behavior:** 
- Open `DatePicker` -> Open `HelperInformation` -> Press `Tab` several times -> Focus outside `HelperInformation`.
- Open `DatePicker` -> Open `HelperInformation` -> Return back to `DatePicker` -> Press `Tab` several times -> Focus outside `DatePicker`.

**Current behavior:**
- Open `DatePicker` -> Open `HelperInformation` -> Press `Tab` several times -> Focus inside `HelperInformation`.
- Open `DatePicker` -> Open `HelperInformation` -> Return back to `DatePicker` -> Press `Tab` several times -> Focus inside `DatePicker`.

#### 2. Button `X` (Close `HelperInformation`) in `HelperInformation` has the same behavior as the `Return back to Calendar` button.
**Previous behavior:** 
- Open `DatePicker` -> Open `HelperInformation` -> Press `X` -> Opened `DatePicker`.

**Current behavior:** 
- Open `DatePicker` -> Open `HelperInformation` -> Press `X` -> Closed `DatePicker`.

#### 3. Unable to select a date if there are disabled days in the current month.
**Previous behavior:** 
- Open `DatePicker` -> Press `Tab` several times -> No focus on dates.

**Current behavior:** 
- Open `DatePicker` -> Press `Tab` several times -> Focus on dates.

#### 4. May use arrows on inactive days.
**Previous behavior:** 
- Open `DatePicker` with inactive days (minDate) -> Select the day right after inactive -> Press `ArrowUp` or `ArrowLeft` several times -> Need to press several times `ArrowDown` or `ArrowRight` to come back to the focused day.

**Current behavior:** 
- Open `DatePicker` with inactive days (minDate) -> Select the day right after inactive -> Press `ArrowUp` or `ArrowLeft` several times -> No need to press several times `ArrowDown` or `ArrowRight` to come back to the focused day. The focus shows the exact position selected.

### DIFF
--- a/packages/react-magma-dom/src/i18n/default.ts
+++ b/packages/react-magma-dom/src/i18n/default.ts
@@ -106,8 +106,7 @@ export const defaultI18n: I18nInterface = {
   datePicker: {
     startOfWeek: 'sunday',
     calendarIconAriaLabel: 'Toggle Calendar Widget',
-    calendarOpenAnnounce:
-      'Calendar Widget is now open. Press the question mark key to get the keyboard shortcuts for changing dates. If you are using the NVDA screen reader, you may need to switch it to Focus Mode for proper interaction with the calendar.',
+    calendarOpenAnnounce: 'Calendar Widget is now open.',
     calendarCloseAriaLabel: 'Close Calendar Widget',
     previousMonthAriaLabel: 'Previous Month',
     nextMonthAriaLabel: 'Next Month',
--- a/packages/react-magma-dom/src/i18n/default.ts
+++ b/packages/react-magma-dom/src/i18n/default.ts
@@ -107,7 +107,7 @@ export const defaultI18n: I18nInterface = {
     startOfWeek: 'sunday',
     calendarIconAriaLabel: 'Toggle Calendar Widget',
     calendarOpenAnnounce:
-      'Calendar Widget is now open. Press the question mark key to get the keyboard shortcuts for changing dates.',
+      'Calendar Widget is now open. Press the question mark key to get the keyboard shortcuts for changing dates. If you are using the NVDA screen reader, you may need to switch it to Focus Mode for proper interaction with the calendar.',
     calendarCloseAriaLabel: 'Close Calendar Widget',
     previousMonthAriaLabel: 'Previous Month',
     nextMonthAriaLabel: 'Next Month',
--- a/.changeset/a11y-datePicker-nav-issues.md
+++ b/.changeset/a11y-datePicker-nav-issues.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(DatePicker): Fix navigation and accessibility issues.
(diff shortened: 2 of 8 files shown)


silvalaura:
This is looking really good, and all the changes so far make sense to me. 

A couple thoughts: 
- I think we should add showing the focus on the Keyboard Icon so when a user opens the help panel, then clicks back to calendar, the focus is visible.
- The page up/down, home/end buttons don't work after a user clicks on the help panel (this wasn't working before either). This is mentioned in the original ticket, so I think it's worth addressing.

packages/react-magma-dom/src/i18n/default.ts
```diff
@@ -107,7 +107,7 @@ export const defaultI18n: I18nInterface = {
     startOfWeek: 'sunday',
     calendarIconAriaLabel: 'Toggle Calendar Widget',
     calendarOpenAnnounce:
-      'Calendar Widget is now open. Press the question mark key to get the keyboard shortcuts for changing dates.',
+      'Calendar Widget is now open. Press the question mark key to get the keyboard shortcuts for changing dates. If you are using the NVDA screen reader, you may need to switch it to Focus Mode for proper interaction with the calendar.',
```
silvalaura, line 110: We should remove `Press the question mark key to get the keyboard shortcuts for changing dates.` since we removed that shortcut in a previous ticket but missed this line.

silvalaura (commented):
Code looks great, and I think these are good improvements.

Testing on my PC:
I can see why you added `If you are using the NVDA screen reader, you may need to switch it to Focus Mode for proper interaction with the calendar.` because it's the only way to actually interact with the calendar. But, I am seeing that when focus mode is off and I press the arrow keys, it reads the next item but doesn't move the focus. Is there _anything_ we can do to force focus mode on automatically?

silvalaura:
I confirmed that Home/End & PageUp&PageDown work on NVDA now as long as you enter insert mode.

However, now on VoiceOver, PageUp&PageDown work but Home/End close the panel: https://jam.dev/c/c87055b6-290e-4ae2-8f1b-f95897bd8822. Are you seeing this too?

silvalaura (approved):
Looks great! Lets have @orion-cengage take a look before we merge

orion-cengage (approved):
This looks and works great!

chris-cedrone-cengage:
Confirm that Inputs have -1px border outline when focused instead of 2px

Text field ✅
Text area ✅
Number ✅
Date ✅
Time ✅
Search ✅
Password ✅
Select ✅
Native select ✅
Combobox ✅

**All verified with Chrome Version 130.0.6723.70, Safari 18 and Firefox 132.0 using Mac OS Sonoma 14.7 on:**
https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/datepicker--default

Issue #1217, opened 2024-03-29
Select: Support disabling individual items


silvalaura:
Currently you can only disable the entire Select component, but not individual items within it. We should support this. 

```
items={[
          {
            { label: 'Red', value: 'red', disabled: false },
            { label: 'Blue', value: 'blue', disabled: true },
          },
}]
```

![image](https://github.com/cengage/react-magma/assets/91160746/ec89c576-d32b-4834-b6c5-5e539b4ae85d)

moathabuhamad-cengage:
@orion-cengage whats the styling  and expected behavior when it comes to keyboard navigation and screen readers for the disabled items

silvalaura:
@moathabuhamad-cengage The behavior should be similar to that of a Native Select (https://codesandbox.io/p/sandbox/native-select-disabled-item-7h9p2q?file=%2Fexample.tsx). 

As far as styling, I think we use `transparentize(0.4, magma.colors.neutral500)` everywhere we disable items.

chris-cedrone-cengage:
**Tested on:**
https://next--upbeat-sinoussi-f675aa.netlify.app/api/select/
&
https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--with-disabled-items

@moathabuhamad-cengage Both function perfectly, however do we want to add `cursor: not-allowed` to disabled items when hovered upon?

moathabuhamad-cengage:
keeping it default as per this discussion  https://github.com/cengage/react-magma/pull/1493#discussion_r1829600659

PR #1493, opened 2024-10-16
feat(Select): Support disabling individual items

moathabuhamad-cengage:
Issue: #1217 

## What I did
- added support for disabling individual items in react-magma-dom
- added a unit test for that behavior 
- added an example in the docs under api/select#disabled_items

## Screenshots:
image

## Checklist 

## How to test
check behavior in storybooks and in docs under api/select#disabled_items

### DIFF
--- a/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
+++ b/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
@@ -1,8 +1,5 @@
 import * as React from 'react';
-import {
-  instanceOfDefaultItemObject,
-  MultiSelectProps,
-} from '.';
+import { instanceOfDefaultItemObject, MultiSelectProps } from '.';
 import { useMultipleSelection, useSelect } from 'downshift';
 import { CloseIcon } from 'react-magma-icons';
 import { ItemsList } from './ItemsList';
@@ -54,10 +51,11 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
   } = props;
 
   function checkSelectedItemValidity(itemToCheck: T) {
-    return (
-      !isItemDisabled(itemToCheck) &&
-      items.findIndex(i => itemToString(i) === itemToString(itemToCheck)) !== -1
+    const itemIndex = items.findIndex(
+      i => itemToString(i) === itemToString(itemToCheck)
     );
+
+    return !isItemDisabled(itemToCheck) && itemIndex !== -1;
   }
 
   function getFilteredItemIndex(item: T, filteredItems: T[]) {
--- a/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
+++ b/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
@@ -55,7 +55,7 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
       i => itemToString(i) === itemToString(itemToCheck)
     );
 
-    return !isItemDisabled(itemToCheck) && itemIndex !== -1;
+    return !isItemDisabled(itemToCheck) && itemIndex !== -1 && !isItemDisabled(items[itemIndex]);
   }
 
   function getFilteredItemIndex(item: T, filteredItems: T[]) {
--- a/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
+++ b/packages/react-magma-dom/src/components/Select/MultiSelect.tsx
@@ -1,5 +1,8 @@
 import * as React from 'react';
-import { instanceOfDefaultItemObject, MultiSelectProps } from '.';
+import {
+  instanceOfDefaultItemObject,
+  MultiSelectProps,
+} from '.';
 import { useMultipleSelection, useSelect } from 'downshift';
 import { CloseIcon } from 'react-magma-icons';
 import { ItemsList } from './ItemsList';
@@ -11,6 +14,8 @@ import { ThemeContext } from '../../theme/ThemeContext';
 import { I18nContext } from '../../i18n';
 import { ButtonSize, ButtonVariant } from '../Button';
 import { defaultComponents } from './components';
+import { useForkedRef } from '../../utils';
+import { isItemDisabled } from './utils';
 
 export function MultiSelect<T>(props: MultiSelectProps<T>) {
   const {
@@ -45,12 +50,46 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
     setFloating,
     setReference,
     isClearable,
+    initialHighlightedIndex,
   } = props;
 
   function checkSelectedItemValidity(itemToCheck: T) {
     return (
+      !isItemDisabled(itemToCheck) &&
       items.findIndex(i => itemToString(i) === itemToString(itemToCheck)) !== -1
     );
+  }
+
+  function getFilteredItemIndex(item: T, filteredItems: T[]) {
+    const index = filteredItems.findIndex(
+      filteredItem => itemToString(filteredItem) === itemToString(item)
+    );
+
+    if (isItemDisabled(filteredItems[index])) {
+      return -1;
+    }
+    return index;
+  }
+
+  function handleOnIsOpenChange(changes) {
+    const { isOpen: changedIsOpen, selectedItem: changedSelectedItem } =
+      changes;
+
+    if (changedIsOpen && changedSelectedItem) {
+      if (isItemDisabled(changedSelectedItem)) {
+        setHighlightedIndex(-1);
+      } else {
+        setHighlightedIndex(
+          items.findIndex(
+            i => itemToString(i) === itemToString(changedSelectedItem)
+          )
+        );
+      }
+    }
+
+    onIsOpenChange &&
+      typeof onIsOpenChange === 'function' &&
+      onIsOpenChange(changes);
   }
 
   const {
@@ -71,6 +110,11 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
     ...(props.selectedItems && {
       selectedItems: props.selectedItems.filter(checkSelectedItemValidity),
     }),
+    ...(props.defaultSelectedItems && {
+      defaultSelectedItems: props.defaultSelectedItems.filter(
+        checkSelectedItemValidity
+      ),
+    }),
   });
 
   function getFilteredItems(unfilteredItems) {
@@ -85,6 +129,7 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
   const {
     stateReducer: passedInStateReducer,
     onStateChange,
+    onIsOpenChange,
     ...selectProps
   } = props;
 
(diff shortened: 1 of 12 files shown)


packages/react-magma-dom/src/components/Select/shared.ts
```diff
@@ -1,11 +1,17 @@
...
+import { Card } from '../Card';
+import { inputBaseStyles } from '../InputBase';
 
 function buildListHoverColor(props) {
+  if (props.isDisabled) {
+    if (props.isInverse) {
+      return transparentize(0.6, props.theme.colors.neutral100);
+    }
+    return transparentize(0.4, props.theme.colors.neutral500);
+  }
```
silvalaura, line 14: I don't think we want to change the background/hover color. Instead, we should just change the text color similar to how we do it in Dropdown:
image
image

silvalaura (changes requested):

silvalaura:
I noticed a little bit of a strange behavior with the focus state. In the screenshot below, I was hovering over Blue, and the blue focus border kept appearing for the item below:
image

packages/react-magma-dom/src/components/Select/shared.ts
```diff
@@ -91,19 +107,18 @@ export const StyledList = styled('ul')<{ isOpen?: boolean; maxHeight: string }>`
...
   cursor: default;
-  color: ${props =>
-    props.isInverse
-      ? props.theme.colors.neutral100
-      : props.theme.colors.neutral700};
+  color: ${props => buildListItemColor(props)};
   line-height: 24px;
   margin: 0;
   padding: 8px 16px;
+  cursor: ${props => (props.isDisabled ? 'not-allowed' : 'pointer')};
```
silvalaura, line 121: @orion-cengage do you want to use the `pointer` hand or `default` which is the existing behavior?

orion-cengage: I would like us to use whatever we're using on disabled buttons and other disabled items. Might be `cursor: not-allowed;`

silvalaura:
I was able to select disabled items:
image
https://jam.dev/c/181a698f-8a4e-4604-a764-53c7a917b634

packages/react-magma-dom/src/components/Select/shared.ts
silvalaura: @orion-cengage I meant for enabled items (I agree with not allowed for disabled items). [Existing behavior](https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--default) is `cursor` but this PR changes it to `pointer`

packages/react-magma-dom/src/components/Select/Select.stories.tsx
```diff
@@ -153,3 +153,47 @@ LeftAlignedLabelWithContainer.decorators = [
...
+    defaultSelectedItems={[
+      { label: 'Red', value: 'red' },
+      { label: 'Orange', value: 'orange', disabled: false },
+      { label: 'Yellow-Disabled', value: 'Yellow-Disabled', disabled: true },
+    ]}
+    initialSelectedItems={[
+      { label: 'Red', value: 'red' },
+      { label: 'Orange', value: 'orange', disabled: false },
+      { label: 'Yellow-Disabled', value: 'Yellow-Disabled', disabled: true },
+    ]}
```
silvalaura, line 197: Can these be part of the controls, so they can be changed on Storybook?

packages/react-magma-dom/src/components/Select/Select.tsx
```diff
@@ -48,6 +48,7 @@ export function Select<T>(props: SelectProps<T>) {
     selectedItem: passedInSelectedItem,
     setReference,
     setFloating,
+    initialHighlightedIndex,
```
silvalaura, line 51: This is already a prop, but it's not documented. Can we properly add it to the props list with an explanation of when it's needed?

silvalaura (commented):
A change somewhere here made us lose the focus back on the select input after isClearable X is pressed like we currently have: https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--default&args=isClearable:true

moathabuhamad-cengage:
> I was able to select disabled items: image https://jam.dev/c/181a698f-8a4e-4604-a764-53c7a917b634

I wasn't able to reproduce this bug but i added more checks anyway

moathabuhamad-cengage:
> A change somewhere here made us lose the focus back on the select input after isClearable X is pressed like we currently have: https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--default&args=isClearable:true

this seems to be broken on dev but I included a fix in this PR

silvalaura:
> > A change somewhere here made us lose the focus back on the select input after isClearable X is pressed like we currently have: https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--default&args=isClearable:true
> 
> this seems to be broken on dev but I included a fix in this PR

This is working as expected on dev (https://next--upbeat-sinoussi-f675aa.netlify.app/api/select/), but I see that it's now been addressed in this PR as well ✅

silvalaura:
> > I was able to select disabled items: image https://jam.dev/c/181a698f-8a4e-4604-a764-53c7a917b634
> 
> I wasn't able to reproduce this bug but i added more checks anyway

Steps to reproduce:
1. Navigate to https://storybook-preview-1493--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--with-disabled-items&args=isMulti:true
2. Click on select, then the down arrow key, then enter to select Red
3. Press the down arrow key again and notice the disabled item has focus. I expect disabled items to never get focus
4. Press arrow key down to select Green, then arrow key down to select Purple. Notice that when all that is left is disabled items, they can still gain focus

I think having `initialHighlightedIndex` is throwing things off because it's not a problem in https://storybook-preview-1493--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--multi-with-disabled-items. Although, if I change the `initialHighlightedIndex1 to 1 https://storybook-preview-1493--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--multi-with-disabled-items&args=initialHighlightedIndex:1 then `Blue-disabled` is the first item focused.

Something is off 🤔

packages/react-magma-dom/src/components/Select/shared.ts
silvalaura: This has not been addressed yet ^

packages/react-magma-dom/src/components/Select/index.tsx
```diff
@@ -142,6 +142,10 @@ export interface SelectProps<T extends SelectOptions>
    * @internal
    */
   isMulti?: false;
+  /**
+   * Index of the item that should be highlighted by default
```
silvalaura, line 146: When should adopters use this?

website/react-magma-docs/src/pages/api/select.mdx
```diff
@@ -159,6 +159,71 @@ export function Example() {
...
+      initialHighlightedIndex={1}
+      defaultSelectedItem={{ label: 'Blue', value: 'blue', disabled: true }}
+      initialSelectedItem={{ label: 'Blue', value: 'blue', disabled: true }}
+    />
+  );
+}
+```
+
+
+```tsx
```
silvalaura, line 192: Instead of 2 separate sandboxes/jsx blocks for the examples, we usually just have 1 with both components we want to showcase.

packages/react-magma-dom/src/components/Select/MultiSelect.tsx
```diff
@@ -45,14 +50,53 @@ export function MultiSelect<T>(props: MultiSelectProps<T>) {
...
+  }
+
   function checkSelectedItemValidity(itemToCheck: T) {
     return (
+      !isItemDisabled(itemToCheck) &&
       items.findIndex(i => itemToString(i) === itemToString(itemToCheck)) !== -1
     );
   }
 
+  function isDisabledItemIndex(index) {
```
silvalaura, line 67: Some of these functions are repeated across MultiSelect and Select-- is it worth having them in a shared file to avoid duplication?

moathabuhamad-cengage:
> > > I was able to select disabled items: image https://jam.dev/c/181a698f-8a4e-4604-a764-53c7a917b634
> > 
> > 
> > I wasn't able to reproduce this bug but i added more checks anyway
> 
> Steps to reproduce:
> 
> 1. Navigate to https://storybook-preview-1493--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--with-disabled-items&args=isMulti:true
> 2. Click on select, then the down arrow key, then enter to select Red
> 3. Press the down arrow key again and notice the disabled item has focus. I expect disabled items to never get focus
> 4. Press arrow key down to select Green, then arrow key down to select Purple. Notice that when all that is left is disabled items, they can still gain focus
> 
> I think having `initialHighlightedIndex` is throwing things off because it's not a problem in https://storybook-preview-1493--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--multi-with-disabled-items. Although, if I change the `initialHighlightedIndex1 to 1 https://storybook-preview-1493--upbeat-sinoussi-f675aa.netlify.app/?path=/story/select--multi-with-disabled-items&args=initialHighlightedIndex:1 then `Blue-disabled` is the first item focused.
> 
> Something is off 🤔

This was caused by mismatched indices between the filtered and unfiltered arrays, fixed by the next commit

packages/react-magma-dom/src/components/Select/MultiSelect.tsx
moathabuhamad-cengage: There is no need now, using new logic for it

website/react-magma-docs/src/pages/api/select.mdx
moathabuhamad-cengage: fixed

packages/react-magma-dom/src/components/Select/index.tsx
moathabuhamad-cengage: fixed

packages/react-magma-dom/src/components/Select/shared.ts
orion-cengage: Let's stick with the default. There doesn't seem to be an agreed-upon standard in other design systems. Some use the hand, some use the pointer. The native select uses the default, so let's just stick with that.

silvalaura:
Issues:
1. Single select > after red is selected, if isClearable is clicked, focus goes to disabled item: https://jam.dev/c/221ef11c-aa53-4484-94ee-da3c6fd428bc
2. Not sure that `initialHighlightedIndex` is now working. I change it in storybook and I don't see it affecting the first item that gets focus

packages/react-magma-dom/src/components/Select/shared.ts
```diff
@@ -91,19 +107,18 @@ export const StyledList = styled('ul')<{ isOpen?: boolean; maxHeight: string }>`
...
   cursor: default;
-  color: ${props =>
-    props.isInverse
-      ? props.theme.colors.neutral100
-      : props.theme.colors.neutral700};
+  color: ${props => buildListItemColor(props)};
   line-height: 24px;
   margin: 0;
   padding: 8px 16px;
+  cursor: ${props => (props.isDisabled ? 'not-allowed' : 'default')};
```
silvalaura, line 121: cursor is set in line 116 and then again here, so we can clean this up

silvalaura (approved):

orion-cengage (approved):

Issue #1249, opened 2024-04-15
Checkbox - Hide SVGs from screen readers


orion-cengage:
CENG_3297

Docs: https://react-magma.cengage.com/version/4.3.0/api/checkboxes/

Success Criteria: https://www.w3.org/WAI/WCAG21/Understanding/non-text-content

Within the checkbox component, SVGs are used for a custom checkbox rather than the browser default that are not named and are not hidden from assistive technology. This could confuse users who reach these elements without knowing what they are for.

Add `aria-hidden="true"` to the `span` element that wraps the SVGs within the checkbox component.

orion-cengage:
I tested this on Storybook and Docs, in Chrome on Mac. Everything looks good.

PR #1513, opened 2024-10-25
fix(CheckBox): Hide SVGs from screen readers by adding `aria-hidden="true"` to the `span` element that wraps the SVGs.

nikitaorliak-cengage:
Issue: #1249

## What I did

Hide SVGs from screen readers by adding `aria-hidden="true"` to the `span` element that wraps the SVGs.

## Checklist 

## How to test

Open `Checkbox` -> Open DevTools -> Check styles.


silvalaura:
This looks good. Can we add the same change to Indeterminate Checkbox? (it's silly, but it's a separate component)

silvalaura (approved):

Issue #1124, opened 2023-09-06
Input: Error message not read by screenreader without helperMessage


silvalaura:
**Describe the bug**
Some screenreaders are not reading input field's error message when "helperMessage" attribute is not specified with any input, although everything works when there is any helper message added (even a space will be enough)

**Expected behavior**
Screen reader reads error message despite the presence of a helper message

**Screenshots**
https://github.com/cengage/react-magma/assets/91160746/8cef6661-16a0-4f56-8f0e-9ca06ee46505

**Additional context**
_The issue is reproducible for:_
Win10 - Chrome - Jaws
Win10/11 - FF/Chrome - NVDA

_Not reproducible for:_
Mac 12.6.8 - Safari 16.6 - VoiceOver
Ubuntu 20.04.6 - FF/Chrome - Orca.

silvalaura:
The original codesandbox: https://codesandbox.io/p/sandbox/35qlkb?file=%2Fpackage.json. Notice it's using 3.1.0. The problem where error message is only read if there is helper message is no longer happening in the latest version, but actually now it never gets read despite the error message.

silvalaura:
From https://www.levelaccess.com/blog/how-to-provide-accessible-form-error-identification/ 

> This is important because after a form is submitted users of screen readers may be taken back to the top of the page. Error messages that appear below forms may not be easily detected by people with disabilities including people with low vision or cognitive disabilities. When an error message does appear at the top of the form the following is recommended:
> 
> The error message statement is indicated as a heading or ARIA region/live region. This will assist users of screen readers in quickly locating the area.
> The error message itself is focused or form field errors are also indicated inline and the first field in error is focused. To focus text such as an error message, the container will need a tabindex="-1" or tabindex="0" (allows focus to be set and places the message in the focus order) attribute and focus can be set using the JavaScript focus() method.

silvalaura:
CS used for testing: https://codesandbox.io/p/sandbox/react-magma-example-forked-m2gm7l?file=%2Fexample.tsx 

Verified that the error message is announced with and without helperMessage for:
- Windows + Chrome + NVDA
- Mac + Chrome + VoiceOver
- Mac + Safari + VoiceOver

The error message is NOT read in Firefox for either Mac or Windows. Will address in #1661

PR #1526, opened 2024-11-04
fix(Input): Clarification that developers should use `ref` with `Input`. Add a new example for Storybook.

nikitaorliak-cengage:
Issue: #1124

## What I did

Add clarification that developers should use `ref` with `Input`. Add a new example for Storybook.

## Checklist 

## How to test

Open `Input` in the docs -> Open `Error message` -> Leave the input blank and click `Submit` -> Check that the input has focus.


silvalaura:
Scheduled a meeting to discuss this as a team, so lets put this on hold for now.

packages/react-magma-dom/src/components/Input/Input.stories.tsx
```diff
@@ -304,3 +306,105 @@ NumberInput.args = {
...
+      second: '',
+      third: '',
+    });
+
+    firstInputRef.current.focus();
+  };
+
+  return (
+    <>
+      <Input
```
silvalaura, line 366: For the sake of readability (and testing with screen readers, can we update these to be `First name`, `Last name` and `Email address` for the labels, and then make the error messages match ("Please enter your ${variable}")

website/react-magma-docs/src/pages/api/form.mdx
```diff
@@ -29,23 +29,128 @@ import {
 } from 'react-magma-dom';
 
 export function Example() {
+  const [state, setState] = React.useState({
```
silvalaura, line 32: For some reason I am not seeing this example in the docs page? 🤔 
image

website/react-magma-docs/src/pages/api/input.mdx
```diff
@@ -170,6 +170,8 @@ export function Example() {
 If an input has an `errorMessage`, the input will be styled to highlight it's error state and the error message will appear below the input field.
 If an error message is present, it will replace the helper text. Can be a node or a string.
 
+For best practices, use a `ref` on the input to enhance usability when handling an `errorMessage`.
```
silvalaura, line 173: Maybe we can move this sentence to go right after `While React Magma provides the error styling, it is up to the consumer app to handle the validation` and we can elaborate a little more

It would read something like:

> While React Magma provides the error styling, it is up to the consumer app to handle the validation. We recommend using a `ref` on the input for accessibility. 
> 
> For short forms with an error, clicking submit should bring the focus back to the input with an error. For long forms, we recommend using an alert to combine the errors and focus should be moved to the alert. See example in Form.

website/react-magma-docs/src/pages/api/input.mdx
```diff
@@ -172,18 +172,22 @@ If an error message is present, it will replace the helper text. Can be a node o
 
 The `required` prop can be used to indicate when a field is required. It is also important to indicate to the user whenever a field is required.
 
-While React Magma provides the error styling, it is up to the consumer app to handle the validation.
+While React Magma provides the error styling, it is up to the consumer app to handle the validation. We recommend using a `ref` on the input for accessibility.
+
+For short forms with an error, clicking submit should bring the focus back to the input with an error. For long forms, we recommend using an alert to combine the errors and focus should be moved to the alert. See example in [Form](https://docs-preview-dev--upbeat-sinoussi-f675aa.netlify.app/api/form/).
```
silvalaura, line 177: You can use<Link to='/api/form'>Form</Link> instead

website/react-magma-docs/src/pages/api/form.mdx
```diff
@@ -29,23 +29,128 @@ import {
...
+  }, [errors]);
+
   return (
     <Form
-      onSubmit={() => alert('form submitted')}
+      onSubmit={onSubmit}
       header="Form Heading"
       description="Some Form Description"
-      errorMessage="Some Form Error"
+      errorMessage={errorMessage}
```
silvalaura, line 111: This looks good, the only thing is that the alert does not get focus / the message is not announced with a screenreader

silvalaura: This has not been addressed yet: https://jam.dev/c/7e5f61b7-cf01-4511-acdc-409d1b8760e0

I think I expect the focus to move to the alert so that the error message can be read.

silvalaura (changes requested):
Merge conflicts also need to be addressed

silvalaura (changes requested):
Just missing addressing this, since this is what we describe in the docs:
https://github.com/cengage/react-magma/pull/1526#discussion_r1890861819

silvalaura (approved):

orion-cengage (changes requested):
When you navigate to the Form page, the focus is being put on the alert in the inverse example and scrolling down to it. This is not the appropriate way to show this functionality. Focus should only be put on the Alert when appropriately triggered in the example.

Also on the Form page, we need to better represent how we want people to present form errors like this. Set a good example. The screenshot below is what I would expect to see.
- The focus outline on the Alert needs to have a 2px offset. Also might need to put the outline one div deeper in the alert to get the rounded corners.
- Change message in the Alert to what I have in the screenshot below.
- The inputs are required so we need to indicate that in their labels. Example: First Name (required)
- The inputs also need to reflect that there are errors. It can't be all on the Alert.

image

orion-cengage (approved):

PR #1527, opened 2024-11-04
docs: Update changelog to explain possible issue with emotion/css.

silvalaura:
Issue: -

## What I did

I received a message from a team having issues with upgrading from 4.3.0 to 4.6.0. Turns out they were using the emotion `css` prop, and that's not working for them anymore

```
const bodyStyle = css`
    padding: 16px 16px 0 16px;
`;

return (
        <Card
            css={cardCss}
            background={theme.colors.neutral600}
            isInverse
        >
            <CardBody css={bodyStyle}>
...
```
Turns out that with the latest version of emotion that we upgraded to, that is no longer supported. These are the supported ways of styling components:
https://codesandbox.io/p/sandbox/emotion-11-and-magma-2tg5w5?file=%2Fexample.tsx

This PR just makes a note in the changeset for adopters to be aware of this.

## Checklist 

### DIFF
--- a/.changeset/docs-emotion.md
+++ b/.changeset/docs-emotion.md
@@ -1,0 +1,5 @@
+---
+'react-magma-docs': patch
+---
+
+docs(emotion): Update changelog to explain possible issue with emotion/css.
--- a/website/react-magma-docs/src/pages/api-intro/changelog.mdx
+++ b/website/react-magma-docs/src/pages/api-intro/changelog.mdx
@@ -9,7 +9,7 @@ order: 4
 
 ### Minor Changes
 
-- ae668a3e5: chore: Updating emotion to v11. **Note: adopters will need to upgrade** their packages to the following versions `"@emotion/react": "^11.13.0", "@emotion/styled": "^11.13.0"`. Notice that `@emotion/core` has been replaced with `@emotion/styled` and that's the only breaking change.
+- ae668a3e5: chore: Updating emotion to v11. **Note: adopters will need to upgrade** their packages to the following versions `"@emotion/react": "^11.13.0", "@emotion/styled": "^11.13.0"`. Notice that `@emotion/core` has been replaced with `@emotion/styled` and that's the only breaking change, unless your project is using `@emotion/css` which may not work the same. Please review the emotion documentation when upgrading.
 - 9e38e9e7f: feat(Stepper): New Stepper component. Displays step based content for use in multi-step interfaces.
 
 ### Patch Changes


moathabuhamad-cengage (approved):

chris-cedrone-cengage:
Verified on:

https://next--upbeat-sinoussi-f675aa.netlify.app/api-intro/changelog/

This PR just makes a note in the changeset for adopters to be aware of this. ✅

garrett-whitehorn-cengage:
@silvalaura I joined that team (Skills) a few months ago. We have been stuck on Magma 4.4 for some time now due to this issue. We have a spike ticket (which links to this PR) in our backlog for researching a path forward that would allow us to upgrade our Magma version, and about a month ago I did a bit of preliminary investigation on it. I'm commenting here to share with you what I found and see what you think.

First, you said above:
> I received a message from a team having issues with upgrading from 4.3.0 to 4.6.0. Turns out they were using the emotion `css` prop, and that's not working for them anymore

That's accurate, but based on the content of this PR I want to clarify: the `css` prop and `@emotion/css` are two very different things. `@emotion/css` "accepts styles as a template literal, object, or array of objects and returns a class name" and is used as input for the `className` prop -- _**not** the `css` prop._ This is not what we are using.

The actual `css` prop is more of a bit of compilation magic, [facilitated by a Babel preset or JSX pragma](https://emotion.sh/docs/css-prop#get-started). Once that is in place, there are 2 ways to use it:
1. [Object styles](https://emotion.sh/docs/css-prop#object-styles): "The `css` prop accepts object styles directly and does not require an additional import."
2. [String styles](https://emotion.sh/docs/css-prop#string-styles): "To pass string styles, you must use `css` which is exported by `@emotion/react`" which is used as a tagged template literal. It does not return a computed class name, but an object that is understood by Emotion at a low level.

We are using the second option, though it's possible that both ways may be equally broken.

Next, you also said:
> Turns out that with the latest version of emotion that we upgraded to, that is no longer supported.

That's not entirely the case. Emotion 11 said this (emphasis mine):
> The way in which we provide TypeScript support for the `css` prop has changed. Based on the usage of our JSX factories, _**we can add support for `css` prop only for components that support `className` prop**_ (as our JSX factory functions take the provided `css` prop, resolve it and pass the generated `className` to the rendered component).

So it is still supported, but now it's only if the component receiving the prop also supports `className`.

That said, since you all seem to have recommended the use of both `@emotion/styled` and `@emotion/css` with the `className` prop, I'm not totally sure why the `css` prop would be broken.

Emotion’s [Best Practices page](https://emotion.sh/docs/best-practices#if-using-react-prefer-emotionreact-or-emotionstyled-over-emotioncss) says:
> If using React, prefer `@emotion/react` or `@emotion/styled` over `@emotion/css`
>
> `@emotion/react` and `@emotion/styled` generally offer a better developer experience over class names (`@emotion/css`) when using React.

As such, we'd like to avoid `@emotion/css`. We could certainly refactor our entire front-end codebase to use `@emotion/styled`, but why should we have to do that if our existing method is equally as good, according to Emotion? And that's not to mention what they say in the next section down:
> (If you're curious, the `css` prop is more popular among the maintainers of Emotion.)

So we're doing things the way the Emotion maintainers prefer -- yet, Magma doesn't support it.

I'd like the Magma team to take a deeper look at why the `css` prop doesn't work with Magma anymore and see what it would take to fix it. Because as far as we're concerned, it's a regression.

Issue #1341, opened 2024-07-01
A11y: Date picker calendar, days of the week aria-label


orion-cengage:
CENG_3415

[https://react-magma.cengage.com/version/4.4.0/api/date-pickers/](https://react-magma.cengage.com/version/4.4.0/api/date-pickers/)

### Description:
The table headers are terse; just letters to represent days. Although descriptive labels are used for the buttons within cells, it is not necessarily by button focus that a screen reader would be traversing the table.

### Remediation:
Recommended each column header has an `aria-label` that replaces (for example) “T” with “Tuesday”. This will produce some redundancy, since the buttons describe the full day, but is a “belt and braces” approach.

![image](https://github.com/cengage/react-magma/assets/24397282/4c002ec7-93b8-4a1c-9567-1c97ca3ca5ba)

chris-cedrone-cengage:
Verified on:

https://next--upbeat-sinoussi-f675aa.netlify.app/api/date-pickers/

&

https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/datepicker--default

Add aria-label to calendar header columns. ✅

PR #1528, opened 2024-11-05
fix(DatePicker): Add `aria-label` to calendar header columns.

nikitaorliak-cengage:
Issue: #1341

## What I did

Add `aria-label` to calendar header columns.

## Checklist 

## How to test

Open `DatePicker` -> Open DevTools -> Check that each header column has the correct `aria-label`.

### DIFF
--- a/packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
+++ b/packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
@@ -143,7 +143,12 @@ export const CalendarMonth: React.FunctionComponent<CalendarMonthProps> = (
   const startOfWeek = days.indexOf(i18n.datePicker.startOfWeek);
   const sortedDays = days.slice(startOfWeek).concat(days.slice(0, startOfWeek));
   const tableDaysHeaders = sortedDays.map((day, index) => (
-    <Th key={index} theme={theme} isInverse={context.isInverse}>
+    <Th
+      key={index}
+      theme={theme}
+      isInverse={context.isInverse}
+      aria-label={i18n.days.long[day]}
+    >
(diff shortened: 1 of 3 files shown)


packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
```diff
@@ -143,7 +143,12 @@ export const CalendarMonth: React.FunctionComponent<CalendarMonthProps> = (
   const startOfWeek = days.indexOf(i18n.datePicker.startOfWeek);
   const sortedDays = days.slice(startOfWeek).concat(days.slice(0, startOfWeek));
   const tableDaysHeaders = sortedDays.map((day, index) => (
-    <Th key={index} theme={theme} isInverse={context.isInverse}>
+    <Th
+      key={index}
+      theme={theme}
+      isInverse={context.isInverse}
+      aria-label={i18n.days.long[day]}
```
silvalaura, line 150: Do we need to account for the locale?

For example, here we have an example that uses `locale: es.` Should the days match that language?
https://docs-preview-1528--upbeat-sinoussi-f675aa.netlify.app/api/date-pickers/#internationalization

silvalaura: Upon further investigation, I think we just need to update that i18n example to include the following and it will work as expected
```
days: {
    long: {
      monday: 'Lunes',
      tuesday: 'Martes',
      wednesday: 'Miércoles',
      thursday: 'Jueves',
      friday: 'Viernes',
      saturday: 'Sábado',
      sunday: 'Domingo',
    },
```

silvalaura (approved):

PR #1541, opened 2024-11-08
feat(Table, Datagrid): Added new `tableTitle` prop

### LINKED ISSUE #1395: A11y Fail: Table + Data grid - Need to use Caption as title of table
CENG_1953

https://react-magma.cengage.com/version/4.5.0/api/datagrid/

https://react-magma.cengage.com/version/4.5.0/api/table/

### Issue:
It is a best practice to ensure that all <table> elements have a <caption> element. A caption functions like a heading for a table. Most screen readers announce the content of captions. Captions help users to find a table and understand what it’s about and decide if they want to read it. If the user uses “Tables Mode”, captions are the primary mechanism to identify tables.

### Suggestion:
We need to document and make it easy for devs to use "caption" attribute to title tables and data grids. Additionally, we need to let them know if they want the caption to also be a heading within the structure of the page, they can do both at the same time.

`<table><caption><h2>This is the title of the table</h2></caption>...`

chris-cedrone-cengage:
Issue: # [1395](https://github.com/cengage/react-magma/issues/1395)

## What I did
- New `tableTitle` prop for both `Table` and `Datagrid` to enable the use of captions and further improve accessibility across both components.

## Checklist 

### DIFF
--- a/packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
+++ b/packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
@@ -342,18 +342,20 @@ const defaultArgs = {
 };
 
 export const Default = Template.bind({});
-Default.args = defaultArgs;
+Default.args = { ...defaultArgs, tableTitle: 'Default' };
 
 export const ColoredRows = Template.bind({});
 ColoredRows.args = {
   ...defaultArgs,
   rows: coloredRows,
+  tableTitle: 'Colored rows',
 };
 
 export const Selectable = Template.bind({});
 Selectable.args = {
   ...defaultArgs,
   isSelectable: true,
+  tableTitle: 'Selectable',
 };
 
 export const SelectableAndSortable: Story<DatagridProps> = ({
@@ -541,18 +543,21 @@ export const SelectableAndSortable: Story<DatagridProps> = ({
 SelectableAndSortable.args = {
   isSelectable: true,
   isSortableBySelected: true,
+  tableTitle: 'Selectable and sortable',
 };
 
 export const ControlledSelectable = ControlledTemplate.bind({});
 ControlledSelectable.args = {
   ...defaultArgs,
   isSelectable: true,
+  tableTitle: 'Controlled selectable',
 };
 
 export const DisabledSelectableRow = Template.bind({});
 DisabledSelectableRow.args = {
   ...defaultArgs,
   isSelectable: true,
+  tableTitle: 'Disabled selectable row',
   rows: [
     ...defaultArgs.rows,
     {
@@ -569,6 +574,7 @@ DisabledSelectableRow.args = {
 export const PaginationChangedDefaults = Template.bind({});
 PaginationChangedDefaults.args = {
   ...defaultArgs,
+  tableTitle: 'Pagination changed defaults',
   paginationProps: {
     defaultPage: 2,
     defaultRowsPerPage: 5,
@@ -579,6 +585,7 @@ PaginationChangedDefaults.args = {
 export const ControlledPagination = ControlledPaginatedTemplate.bind({});
 ControlledPagination.args = {
   ...defaultArgs,
+  tableTitle: 'Controlled pagination',
   paginationProps: {
     rowsPerPageValues: [5, 10, 20],
   },
@@ -587,6 +594,7 @@ ControlledPagination.args = {
 export const WithoutPagination = Template.bind({});
 WithoutPagination.args = {
   ...defaultArgs,
+  tableTitle: 'Without pagination',
   hasPagination: false,
 };
 
@@ -632,6 +640,7 @@ const CustomPaginationComponent: React.FunctionComponent<
 export const PaginationWithCustomComponent = Template.bind({});
 PaginationWithCustomComponent.args = {
   ...defaultArgs,
+  tableTitle: 'Pagination with custom component',
   components: {
     Pagination: CustomPaginationComponent,
   },
--- a/packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
+++ b/packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
@@ -340,13 +340,17 @@ const defaultArgs = {
   isInverse: false,
   isSortableBySelected: false,
   paginationProps: {},
+  tableTitle: 'Default'
 };
 
 export const Default = Template.bind({});
-Default.args = { ...defaultArgs, tableTitle: 'Default' };
+Default.args = { ...defaultArgs };
 
 export const TitleWithNode = Template.bind({});
+
 TitleWithNode.args = { ...defaultArgs, tableTitle: <Heading level={1}>Title with node</Heading> };
+
+TitleWithNode.parameters = { controls: { exclude: ['tableTitle'] } };
 
 export const ColoredRows = Template.bind({});
 ColoredRows.args = {
--- a/packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
+++ b/packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
@@ -282,6 +282,11 @@ export default {
         type: 'boolean',
       },
     },
+    tableTitle: {
+      control: {
+        type: 'text',
+      },
+    },
   },
 } as Meta;
 
--- a/packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
+++ b/packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
@@ -15,6 +15,7 @@ import { magma } from '../../theme/magma';
 import { Spacer, SpacerAxis } from '../Spacer';
 import { Announce } from '../Announce';
 import { VisuallyHidden } from '../VisuallyHidden';
+import { Heading } from '../Heading';
 
 const rowsForPagination = [
   {
@@ -343,6 +344,9 @@ const defaultArgs = {
 
 export const Default = Template.bind({});
 Default.args = { ...defaultArgs, tableTitle: 'Default' };
+
+export const TitleWithNode = Template.bind({});
+TitleWithNode.args = { ...defaultArgs, tableTitle: <Heading level={1}>Title with node</Heading> };
 
 export const ColoredRows = Template.bind({});
 ColoredRows.args = {
--- a/packages/react-magma-dom/src/components/Datagrid/Datagrid.test.js
+++ b/packages/react-magma-dom/src/components/Datagrid/Datagrid.test.js
@@ -4,7 +4,12 @@ import { Datagrid } from '.';
 import { TableRowColor } from '../Table';
 import { Button } from '../Button';
 import { usePagination } from '../Pagination/usePagination';
-import { render, screen, fireEvent } from '@testing-library/react';
+import {
+  render,
+  screen,
+  fireEvent,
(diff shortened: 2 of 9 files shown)


packages/react-magma-dom/src/components/Datagrid/Datagrid.tsx
```diff
@@ -51,6 +53,7 @@ export interface DatagridRow {
    * Used to allow each unique column field as a key with the row content as the value
    */
   [key: string]: any;
+  tableTitle?: any;
```
silvalaura, line 56: I think we want users to pass a node (React.ReactNode), and maybe also a string? but I don't think we should make this as broad as `any`

packages/react-magma-dom/src/components/Datagrid/Datagrid.tsx
```diff
@@ -138,6 +141,11 @@ export interface UncontrolledSelectedRowsProps {
   defaultSelectedRows?: (string | number)[];
 }
 
+export const StyledCaption = styled.caption`
+  display: flex;
+  flex: 1;
+`;
```
silvalaura, line 147: We should check with @orion-cengage what the default heading title styles should be

packages/react-magma-dom/src/components/Datagrid/Datagrid.tsx
```diff
@@ -278,6 +287,8 @@ export const Datagrid = React.forwardRef<HTMLTableElement, DatagridProps>(
 
     return (
       <>
+        {tableTitle && <StyledCaption>{tableTitle}</StyledCaption>}
+
```
silvalaura, line 291: Should we by default add some spacing?

packages/react-magma-dom/src/components/Datagrid/Datagrid.tsx
silvalaura: Since you're already exporting it, we should define the StyledCaption in only one place and use it in both Table and Datagrid so they will always be consistent.

website/react-magma-docs/src/pages/api/table.mdx
silvalaura, line 1: We need to update every example in our docs site to use the heading, and they should each be unique: Basic Usage Table, Vertical Borders Table, etc.

packages/react-magma-dom/src/components/Datagrid/Datagrid.tsx
silvalaura: We also need to add a description to explain to adopters when to use this

website/react-magma-docs/src/pages/api/table.mdx
silvalaura: And we should update datagrid.mdx as well

packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
```diff
@@ -282,6 +282,11 @@ export default {
         type: 'boolean',
       },
     },
+    tableTitle: {
+      control: {
+        type: 'text',
```
silvalaura, line 287: Can we add a default value as well so we can always see it?

silvalaura (changes requested):

packages/react-magma-dom/src/components/Datagrid/Datagrid.test.js
```diff
@@ -836,6 +841,18 @@ describe('Datagrid', () => {
     });
   });
 
+  it('should have a title above the Datagrid', () => {
```
silvalaura, line 844: `should display the title` is more accurate to what the test is doing (we aren't checking position, just that it's present

packages/react-magma-dom/src/components/Table/Table.stories.tsx
silvalaura, line 1: Can we have some examples that pass a node with a Heading component / Paragraph? All of these use a string

website/react-magma-docs/src/pages/api/table.mdx
```diff
@@ -116,32 +118,31 @@ import {
...
-          <TableCell>Lorem ipsum dolor sit amet</TableCell>
-          <TableCell>Lorem ipsum dolor</TableCell>
-          <TableCell>Lorem ipsum dolor</TableCell>
-          <TableCell>Lorem ipsum</TableCell>
-        </TableRow>
-      </TableBody>
-    </Table>
-  );
+  return;
+  <Table tableTitle="Vertical borders table" hasVerticalBorders>
```
silvalaura, line 122: This doesn't look right

silvalaura: Op, it crashes the whole Table page
image

website/react-magma-docs/src/pages/api/table.mdx
```diff
@@ -162,7 +163,7 @@ import {
 
 export function Example() {
   return (
-    <Table hasHoverStyles>
+    <Table tableTitle=`Hover style table` hasHoverStyles>
```
silvalaura, line 166: You should be able to pass a regular string `"Hover style table"` without ticks

website/react-magma-docs/src/pages/api/datagrid.mdx
silvalaura, line 1: Not seeing the titles in these examples:
image

packages/react-magma-dom/src/components/Table/Table.tsx
```diff
@@ -122,6 +127,24 @@ export const TableContainer = styled.div<{
           : props.theme.colors.focus};
   }
 `;
+export const StyledCaption = styled.caption<{ isInverse: boolean }>`
+  color: ${props =>
```
silvalaura, line 131: If we are copying basically all the styles from `headingMediumStyles`, we should be able to do something like this so we don't duplicate code and if one gets updated, they both change simultaneously:
```
export const StyledCaption = styled.caption<{ isInverse: boolean }>`
    ${headingMediumStyles(props)}
    display: flex;
    flex: 1;
`;
```

silvalaura (changes requested):
Just a few more changes

website/react-magma-docs/src/pages/api/table.mdx
chris-cedrone-cengage: Thanks Prettier

packages/react-magma-dom/src/components/Table/Table.tsx
```diff
@@ -122,6 +128,12 @@ export const TableContainer = styled.div<{
           : props.theme.colors.focus};
   }
 `;
+export const StyledCaption = styled.caption<{ isInverse: boolean; tableTitleNode: boolean; }>`
+  ${headingMediumStyles};
+  margin: ${props => props.tableTitleNode ? `` : props.theme.spaceScale.spacing04};
```
silvalaura, line 133:
```suggestion
  margin: ${props => props.tableTitleNode || props.theme.spaceScale.spacing04};
```

The logical OR operator || can be used to check if props.tableTitleNode is truthy. If it is truthy, an empty string is used as the value of margin. If props.tableTitleNode is falsy, props.theme.spaceScale.spacing04 is used as the value of margin

OR

you can leave as is but you can just use `""` instead of ticks

packages/react-magma-dom/src/components/Table/Table.tsx
```diff
@@ -122,6 +128,12 @@ export const TableContainer = styled.div<{
           : props.theme.colors.focus};
   }
 `;
+export const StyledCaption = styled.caption<{ isInverse: boolean; tableTitleNode: boolean; }>`
```
silvalaura, line 131: By default, captions are centered aligned.

@orion-cengage are you okay with this or should we force a different alignment?
image

website/react-magma-docs/src/pages/api/table.mdx
silvalaura, line 1: Lets have at least one example here and in DataGrid where we use a node

silvalaura (changes requested):

packages/react-magma-dom/src/components/Table/Table.stories.tsx
```diff
@@ -868,4 +886,56 @@ export const NoRowsPerPageControl = args => {
       />
     </Card>
   );
-};
\ No newline at end of file
+};
+
+export const TableTitleWithNode = args => {
+  return (
+    <Table tableTitle={<Heading level={1}>Heading component example</Heading>} {...args}>
```
silvalaura, line 893: Because `{...args}` is at the end, it's replacing tableTitle with the one that gets passed in so it reads `Row colors` instead of `Heading component example`

packages/react-magma-dom/src/components/Table/Table.stories.tsx
```diff
@@ -51,7 +52,7 @@ const rows = [
 
 const Template: Story<TableProps> = args => (
   <Card isInverse={args.isInverse}>
-    <Table {...args}>
+    <Table tableTitle="Basic usage" {...args}>
```
silvalaura, line 55: This isn't needed as it will always get overwritten by `args` as long as the Default has tableTitle

packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
```diff
@@ -337,18 +343,23 @@ const defaultArgs = {
 };
 
 export const Default = Template.bind({});
-Default.args = defaultArgs;
+Default.args = { ...defaultArgs, tableTitle: 'Default' };
+
+export const TitleWithNode = Template.bind({});
+TitleWithNode.args = { ...defaultArgs, tableTitle: <Heading level={1}>Title with node</Heading> };
```
silvalaura, line 349: We could do something like this to exclude tableTitle from this story in order for it to not get replaced
`TitleWithNode.parameters = { controls: { exclude: ['tableTitle'] } };`

packages/react-magma-dom/src/components/Datagrid/Datagrid.stories.tsx
```diff
@@ -337,18 +343,23 @@ const defaultArgs = {
 };
 
 export const Default = Template.bind({});
-Default.args = defaultArgs;
+Default.args = { ...defaultArgs, tableTitle: 'Default' };
```
silvalaura, line 346: `tableTitle: 'Default'` should be moved inside `defaultArgs` where all the default arguments are

packages/react-magma-dom/src/components/Table/Table.tsx
```diff
@@ -45,6 +46,11 @@ export interface TableProps extends React.HTMLAttributes<HTMLTableElement> {
    */
   minWidth?: number;
   rowCount?: number;
+  /**
+   * Title that appears above the Datagrid
```
silvalaura, line 50: I wonder if we should be clear that this will be inserted inside `<caption>`. Something like 

- The title or caption of a table inside a `<caption>` HTML element that provides the table an accessible description

(this applies to both Table and DataGrid)

silvalaura (changes requested):
Looking back at the original ticket...

I think it's expected for caption to be _inside_ the table tag
`<table><caption><h2>This is the title of the table</h2></caption>...`

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption

silvalaura (approved):
Code looks good. Please do NOT merge until after the release (and @orion-cengage 's approval)

orion-cengage:
This was definitely not the intended result on this one, and that's probably my fault for not providing a design. I'll take a closer look when I get a bit of time freed up.

image

PR #1595, opened 2024-12-09
feat(Popover): New `Popover` component

### LINKED ISSUE #927: New component: Popover
Need to create a new component that is visually similar to a tooltip, but is better for containing interactive elements and more complex content.

Figma: https://www.figma.com/file/z2H2qv9iYQX942LqToCuQE/Components---Data-Display?type=design&node-id=625%3A21846&t=O0EUkNbsDC83olKU-1

Inspiration:

- https://mantine.dev/core/popover
- https://carbondesignsystem.com/components/toggletip/usage/
- https://getbootstrap.com/docs/5.3/components/popovers/

Image

### AC

- 1.0 Popover is triggered by clicking on a focusable element (button, input, etc.).
-- 1.1 you can set popover to be triggered on hover if necessary.
- 2.0 Positioning
-- 2.1 Default positioning is centered below the trigger, but you can force it to appear above by default.
-- 2.2 Use FloatingUI to adjust the position

Image

- 3.0 Keyboard
-- 3.1 Pressing Space or Enter also opens and closes the popover
-- 3.2 When focus is inside the popover then you can press Escape to close, and focus goes back to the triggering element.
-- 3.3 If there aren't any focusable elements inside the popover and focus is still on the triggering element, Escape does not close the popover. (note: this needs to be tested; if the screenreader can't read the content, then the focus does need to go inside the container like we do for Informational dropdowns right now)

Image

-- 3.4 If there are focusable elements inside the popover, you should set focus automatically to the first item and trap focus inside the popover.

Image

- 4.0 Popover width
-- 4.1 Width can be manually set to any pixel value

Image

-- 4.2 Width can also be set to automatically match the width of the triggering element

Image

- 5.0 Disabled or unfocusable elements
-- 5.1 (per Bootstrap) Elements with the disabled attribute aren’t interactive, meaning users cannot hover or click them to trigger a popover. As a workaround, you’ll want to trigger the popover from a wrapper <div> or <span>, ideally made keyboard-focusable using tabindex="0".

Image

- 6.0 Popover anatomy and contents
-- 6.1 You can put just about anything inside a popover, although putting something like a dropdown or tooltip is probably not advised. 
-- 6.2 Popover has a main content area, and optional header and footer
--- 6.2.1 If scrolling is needed, in a dropdown like situation for instance, then the body section should scroll vertically with the header and footer remaining fixed in place if those elements are present. Do not scroll horizontally or let content bleed off the page.

Image

-- 6.3 There should be an option to turn off the pointer like in the example below:

Image

- 7.0 Accessibility (Taken from Mantine. We can tweak if necessary)
-- Popover element has role="dialog" and aria-labelledby="target-id" attributes
-- Target element has aria-haspopup="dialog", aria-expanded, aria-controls="popover-id" attributes

- 8.0 We need a function to show the Popover as open by default when the page loads or after a particular action is triggered. This can be useful for tutorial style prompts
-- As a result, we need the user to be able to add a button inside the popover to manually close the popover.
-- We need to make sure developers can add "don't show this again" functionality so the user can tell the system to stop opening the popover by default.

image

- 9.0 Dark mode inverse available.

image

- 10.0 Update Dropdown examples that are similar to Popover
-- In the Dropdown docs, add suggestion for users to use Popover instead of Dropdown for content like Informational Content and filters (tbd?) 

This component should be tested by Perkins before it's released.

nikitaorliak-cengage:
Issue: #927

## What I did

Created a new `Popover` component.

## Checklist 


packages/react-magma-dom/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,226 @@
...
+    popoverTriggerId.current = `${popoverId}_trigger`;
+    popoverContentId.current = `${popoverId}_content`;
+
+    const ref = useForkedRef(forwardedRef, ownRef);
+
+    React.useEffect(() => {
+      if (openByDefault) {
+        openPopover();
+      }
+    }, [openByDefault]); // Do I need this?
```
silvalaura, line 113: What's the question here?

packages/react-magma-dom/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,251 @@
...
+import { ThemeInterface } from '../../theme/magma';
+import { PopoverHeader } from './PopoverHeader';
+import { PopoverFooter } from './PopoverFooter';
+
+/**
+ * @children required
+ */
+export interface PopoverContentProps
+  extends React.HTMLAttributes<HTMLDivElement> {
+  // isInverse?: boolean;
```
silvalaura, line 18: Do we still need this, or do we get it from the context?

packages/react-magma-dom/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,251 @@
...
+  position: relative;
+  background: ${props =>
+    props.isInverse
+      ? props.theme.colors.primary600
+      : props.theme.colors.neutral100};
+  display: ${props => (props.isOpen ? 'block' : 'none')};
+  max-height: ${props => (props.maxHeight ? props.maxHeight : '100%')};
+  opacity: ${props => (props.isOpen ? '1' : '0')};
+  outline: 0;
+  /* padding: ${props => props.theme.spaceScale.spacing05}; */
```
silvalaura, line 44: We can delete this line if we don't need it

packages/react-magma-dom/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+const StyledAnnounce = styled(Announce)<{ maxHeight?: string }>`
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  /* display: grid; */
+  /* grid-template-rows: 1fr auto 1fr; */
+  /* flex-wrap: wrap; */
+  /* justify-content: center; */
+  /* align-items: center; */
```
silvalaura, line 141: Can be deleted?

packages/react-magma-dom/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,251 @@
...
+  max-height: ${props => (props.maxHeight ? props.maxHeight : '100%')};
+`;
+
+const StyledPopoverContent = styled.div`
+  z-index: 2;
+`;
+
+const PopoverChildrenContent = styled.div``;
```
silvalaura, line 149: If there are no styles, we can just use a div without the name

packages/react-magma-dom/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,251 @@
...
+        withoutPointer={context.withoutPointer}
+      >
+        <PopoverChildrenContent
+          aria-labelledby={context.popoverTriggerId.current}
+          role="dialog"
+          ref={focusTrapRef}
+        >
+          <StyledAnnounce maxHeight={context.maxHeight}>
+            {!content.length ? (
+              'Content must be passed'
```
silvalaura, line 238: Not sure we should announce an "error" for the developers. I think we can assume there is content?

packages/react-magma-dom/src/components/Popover/PopoverHeader.tsx
silvalaura, line 1: Are the header and footer practically the same?

packages/react-magma-dom/src/components/Popover/Popover.stories.tsx
silvalaura, line 1: We should add more examples to match all the ones provided in the designs

silvalaura (commented):
Looking good!

packages/react-magma-dom/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,226 @@
...
+  onOpen?: () => void;
+  testId?: string;
+  positioning?: PopoverPositioning;
+  maxHeight?: string | number;
+  width?: string | number;
+  hoverable?: boolean;
+  focusable?: boolean;
+  isInverse?: boolean;
+  isDisabled?: boolean;
+  matchedWidth?: boolean;
```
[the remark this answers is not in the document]
silvalaura: Mantine does `width="target"` for matching the width to the target element, or it takes a number for it to be a fixed width. Maybe we can do the same?

chris-cedrone-cengage:
Needs doc site example:

"10.0 Update Dropdown examples that are similar to Popover
-- In the Dropdown docs, add suggestion for users to use Popover instead of Dropdown for content like Informational Content and filters (tbd?)"

packages/react-magma-dom/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+const StyledPopoverContent = styled.div`
+  z-index: 2;
+`;
+
+const ScrollableContent = styled.div`
+  display: flex;
+  flex-direction: column;
+  overflow-y: auto;
+  padding: 16px;
```
chris-cedrone-cengage, line 142: Probably want to use the theme `props => props.theme.spaceScale.spacing05` here.

packages/react-magma-dom/src/components/Popover/PopoverSection.tsx
```diff
@@ -0,0 +1,98 @@
...
+  children: React.ReactChild | React.ReactChild[];
+  /**
+   * @internal
+   */
+  style?: React.CSSProperties;
+}
+
+export const StyledSection = styled.section`
+  position: relative;
+  padding: 16px;
```
chris-cedrone-cengage, line 31: Probably want to use the theme props => props.theme.spaceScale.spacing05 here.

chris-cedrone-cengage:
For these Storybook examples, `hoverable` doesn't seem to work, should it?

**Profile Example,
Dont Show Again**

packages/react-magma-dom/src/components/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,959 @@
...
+  );
+};
+
+export const DisabledTrigger = DisabledTriggerExample.bind({});
+DisabledTrigger.args = {
+  width: 320,
+  isDisabled: true,
+};
+
+const ProfileExampleTemplate = args => {
```
silvalaura, line 769: This made me laugh and I love it so much 💯 
image

packages/react-magma-dom/src/components/Popover/Popover.stories.tsx
silvalaura, line 1: Great job on all these examples!

packages/react-magma-dom/src/components/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,959 @@
...
+            style={{
+              display: 'flex',
+              flexDirection: 'column',
+              gap: '8px',
+            }}
+          >
+            <div style={{ display: 'flex', gap: '8px', alignItems: 'center' }}>
+              {errorState.length ? (
+                <CloseIcon
+                  color={args.isInverse ? '#FAAEB0' : '#D32821'}
```
silvalaura, line 584: We can use theme variables here and in the lines below with hex colors

packages/react-magma-dom/src/components/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,959 @@
...
+        height: '300px',
+        justifyContent: 'center',
+        flexDirection: 'row',
+        gap: '8px',
+      }}
+      isInverse={args.isInverse}
+    >
+      <img
+        src="https://upload.wikimedia.org/wikipedia/commons/3/33/Mark_Kassen%2C_Tony_C%C3%A1rdenas_and_Chris_Evans_%28cropped%29.jpg"
+        alt="Mark_Kassen"
```
silvalaura, line 784:
```suggestion
        alt="Chris Evans"
```

packages/react-magma-dom/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,224 @@
...
+  flex-direction: column;
+`;
+
+const ErrorMessage = styled.div`
+  width: fit-content;
+  padding: 1em;
+`;
+
+const StyledPopoverContent = styled.div`
+  z-index: 2;
```
silvalaura, line 135: I haven't pulled the code but this reminds me, how does the popover behave/interact with other elements with z-indexes? What if it's inside a drawer or next to a dropdown?

packages/react-magma-dom/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,224 @@
...
+        data-popover-placement={context.position}
+        hasPointer={context.hasPointer}
+      >
+        <div
+          aria-labelledby={context.popoverTriggerId.current}
+          role="dialog"
+          ref={focusTrapRef}
+        >
+          {!content.length ? (
+            <ErrorMessage>Content must be passed</ErrorMessage>
```
silvalaura, line 208: I'm debating whether we should show an error message like this, or just show empty space. I don't think we do this for any other component, and I worry about this being visible to actual users accidentally. Just curious about your thought process in including this.

packages/react-magma-dom/src/components/Popover/PopoverTrigger.tsx
```diff
@@ -0,0 +1,150 @@
...
+export const PopoverTrigger = React.forwardRef<
+  HTMLButtonElement,
+  PopoverTriggerProps
+>((props, forwardedRef) => {
+  const context = React.useContext(PopoverContext);
+  const theme = React.useContext(ThemeContext);
+
+  const ref = useForkedRef(context.toggleRef, forwardedRef);
+
+  const { icon = <InfoIcon />, children, ...other } = props;
```
silvalaura, line 61: @orion-cengage do we want to default to the InfoIcon or always expect users to choose?

silvalaura (commented):
This is more a requirements question for @orion-cengage. There is an image of an opened popover with a disabled trigger
```
-- 5.1 (per Bootstrap) Elements with the disabled attribute aren’t interactive, meaning users cannot hover or click them to trigger a popover. As a workaround, you’ll want to trigger the popover from a wrapper
or , ideally made keyboard-focusable using tabindex="0".
```
What is the expected behavior?

silvalaura:
I chatted with Orion and he agrees that it would be fine to have a `trapFocus` prop that users have to turn on manually if they use interactive elements, similar to how Mantine handles [focus traps](https://mantine.dev/core/popover/#focus-trap). 

When building out a solution for disabled Popovers, I think we need to make a distinction between:
1. **A disabled trigger**: the trigger (likely a button) is disabled, but a popover is still visible
-- this would happen when `isDisabled` is passed to the `PopoverTrigger` OR if `PopoverTrigger` has no disabled prop but it wraps a disabled item such as `<PopoverTrigger><Button disabled>Button with Popover</Button></PopoverTrigger>`
-- use case: a button is disabled and we want to explain to users why
-- [Bootstrap example](https://mantine.dev/core/popover/#disabled)

2. **Popover is disabled**: the trigger exists on the page but hovering/clicking on it will not make the popover visible 
-- this would happen when `isDisabled` is passed to the `Popover`
-- use case: popovers are conditional on items, and users may or may not want it to be visible
-- [Mantine example](https://mantine.dev/core/popover/#disabled)

Ideally, we want to support both. Please let me know is making `trapFocus` a conditional prop helps in achieving these disabled scenarios, or what all you discover while looking into this.

Other notes:
- I don't think we need `aria-label`s on our Popover trigger, not seeing that in other systems. If we do, we probably want it to be something our users can customize.

silvalaura (commented):
This is looking great! I am so glad you got the disabled trigger working. I think this is ready for docs examples 🤩

website/react-magma-docs/src/pages/api/popover.mdx
```diff
@@ -0,0 +1,467 @@
...
+
+**Any additional props supplied will be provided to the wrapping section element.**
+
+<PopoverSectionProps />
+
+## Popover Trigger Props
+
+**Props supplied to the PopoverTrigger component will be provided to the underlying button element.**
+
+<PopoverTriggerProps />
```
silvalaura, line 465: For some reason, these are not visible in the docs page
image

website/react-magma-docs/src/pages/api/popover.mdx
```diff
@@ -0,0 +1,467 @@
...
+
+## Popover Content Props
+
+**Props supplied will be passed to the wrapping div element containing the popover content.**
+
+<PopoverContentProps />
+
+## Popover Section Props
+
+**Any additional props supplied will be provided to the wrapping section element.**
```
silvalaura, line 457: For this to be true, we'd need to add the rest of the props:

`<PopoverSection section={SectionEnum.header} style={style} {...others}>`

orion-cengage:
This is really looking great so far!

Can we use the arrow positioning capabilities of floating-ui to keep the popover from getting cutoff when the trigger is close to the right or left edge of the browser. Here's an illustration of what I mean:

image

orion-cengage:
Can we please decrease the size of the pointer a little bit? The implemented pointer is about 22px wide and the one in my design is only about 12px.

image

image

orion-cengage:
Bug: If you turn on the popover header OR footer, the arrow isn't positioned correctly.

image

image
image

orion-cengage:
Bug: If you change the width of the popover, and you get further away from the default of 320, the position of the arrow is incorrect.

200px
image

600px
image

orion-cengage:
Please rewrite the opening sentence on the docs site / Implementation tab.

**Current:**
The Popover component provides an overlay to display content (such as menus, tooltips or custom interactive sections) in a floating container positioned relative to a trigger element. It supports multiple interaction modes including click and hover, and can be customized with headers, footers, focus trapping, and more.

**Change to:**
A popover is a small UI element that appears on top of content when triggered, showing additional information or actions.

orion-cengage:
Pointer

I would like the pointer being on to be considered the default, with the ability to hide the pointer.

- Please add the pointer on all of the examples.
- In the section about the pointer-related prop, rephrase the text from the point of view of hiding the pointer and reflect that in the example.

orion-cengage:
**Bug: Safari / Mac OS**

When you click on the trigger, the popover appears. When you click on the trigger again, the popover closes and then immediately reopens. This is incorrect behavior.

website/react-magma-docs/src/pages/api/dropdown.mdx
```diff
@@ -653,6 +653,11 @@ Dropdowns can contain elements other than `DropdownMenuItem` components. They ca
 While users may navigate between `DropdownMenuItem` components using up and down arrow keys,
 browser default keyboard behavior applies to other content.
 
+<Alert variant="info">
+  For displaying informational content or filters, consider using the{' '}
+  <Link to="/api/popover">Popover</Link> component instead of Dropdown.
+</Alert>
+
 ### Dropdown with Filter
```
silvalaura, line 661: Lets get rid of the Dropdown with Filter + Dropdown with Informational Content examples, plus the added Popover ones. With this new alert, users should go over to the Popover page for those examples.

packages/react-magma-dom/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,320 @@
...
+   */
+  isDisabled?: boolean;
+  /**
+   * If true, a pointer (arrow) is displayed pointing towards the trigger element.
+   * @default false
+   */
+  hasPointer?: boolean;
+  /**
+   * If true, the popover is open by default when the component is first rendered.
+   * @default true
```
silvalaura, line 75: I think this is supposed to be false

packages/react-magma-dom/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,320 @@
...
+   */
+  isInverse?: boolean;
+  /**
+   * If true, the popover will be disabled and cannot be opened.
+   * @default false
+   */
+  isDisabled?: boolean;
+  /**
+   * If true, a pointer (arrow) is displayed pointing towards the trigger element.
+   * @default false
```
silvalaura, line 70: This should be true

website/react-magma-docs/src/pages/api/popover.mdx
```diff
@@ -0,0 +1,464 @@
...
+
+<PopoverSectionProps />
+
```
silvalaura, line 463: We are missing PopoverTriggerProps

website/react-magma-docs/src/pages/api/popover.mdx
```diff
@@ -0,0 +1,464 @@
...
+        </FormGroup>
+      </PopoverContent>
+    </Popover>
+  );
+}
+```
+
+## Disabled Popover / Disabled Trigger
+
+The `isDisabled` property will disable the ability to open the popover. For disabled trigger and enabled content, you need to place the disable button inside the trigger and use the `hoverable` property.
```
silvalaura, line 336:
```suggestion
The `isDisabled` property will disable the ability to open the popover. For disabled triggers with enabled content, you need to place the disabled button inside the `PopoverTrigger` and use the `hoverable` property.
```

website/react-magma-docs/src/pages/api/popover.mdx
```diff
@@ -0,0 +1,464 @@
...
+        <div>This popover has a fixed width and maximum height</div>
+      </PopoverContent>
+    </Popover>
+  );
+}
+```
+
+## Popover Positions
+
+The `position` prop determines where the popover is displayed relative to its trigger. The available options are `PopoverPosition.bottom` (default) and `PopoverPosition.top`.
```
silvalaura, line 228:
```suggestion
The `position` prop sets the default position for the popover in relation to its trigger. It uses floating-ui to move the position if there is not enough space on the screen. The available options are `PopoverPosition.bottom` (default) and `PopoverPosition.top`.
```

website/react-magma-docs/src/pages/api/popover.mdx
silvalaura, line 1: Something is causing the examples to error on the docs site: `ReferenceError: Popover is not defined`

silvalaura (commented):
Component looks and behaves great! Just a few more finishing touches on the examples.

orion-cengage:
**_Popover - Text only with hover:_** When you tab to the triggering element with the keyboard and the popover opens automatically, the screen reader also needs to read the content of the popover. This is essentially the exact same behavior as our current Tooltip component.

orion-cengage:
Just found this issue. The focus ring on the disabled trigger shouldn't be gray. We choose our focus style colors specifically for a11y visibility.
image

silvalaura (approved):
Looks good after the focus color update!

orion-cengage:
I'm seeing issues on the "Profile Example."

- The name in the example (Chris Evans) looks like a link, but it doesn't have a hover or focus state. Can't we just use the hyperlink component here?
- When you click on the name to open the Popover, clicking outside the popover doesn't close it the way it should. You can currently only close it by clicking on the trigger again, or getting focus inside the popover and then clicking outside of it.

image

silvalaura:
> I'm seeing issues on the "Profile Example."
> 
> * The name in the example (Chris Evans) looks like a link, but it doesn't have a hover or focus state. Can't we just use the hyperlink component here?
> * When you click on the name to open the Popover, clicking outside the popover doesn't close it the way it should. You can currently only close it by clicking on the trigger again, or getting focus inside the popover and then clicking outside of it.
> 
> image

I would recommend we consider if this is an actual issue with Popover or just with the way this example is set up. If it's an issue with the example only, it may not be worth fixing (it's only in Storybook, so not user facing). As fun as it is, we may be able to delete it and move forward.

orion-cengage (approved):

moathabuhamad-nix (dismissed):

moathabuhamad-cengage (approved):

chris-cedrone-cengage (approved):

PR #1622, opened 2025-01-17
fix(FormGroup): Fix validation for `InputMessage`

nikitaorliak-cengage:
Issue: #

## What I did

Validation for `InputMessage` in `FormGroup` created an empty container when there was no message. Fixed this validation.

## Checklist 


silvalaura (approved):
I don't see any reason why we should keep the message container even when it's empty

PR #1652, opened 2025-01-31
fix(Indeterminate Checkbox): Fixed issue with voice over not recognizing the mixed state when using keyboard navigation.

### LINKED ISSUE #1231: Indeterminate Checkbox - Incorrect message announced to assistive tech
CENG_3302

Docs: https://react-magma.cengage.com/version/4.3.0/api/indeterminate/

Success Criteria: https://www.w3.org/WAI/WCAG21/Understanding/name-role-value

Within the Indeterminate checkbox, when the state of a checkbox is `indeterminate` that message is not announce to assistive technology. The message that is announced is not checked.

Use aria-checked="mixed" for an indeterminate checkbox when some of, but not all of, it's children are selected. A detailed example of this can be found here: https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox-mixed/

### LINKED ISSUE #1566: Indeterminate Checkbox: screenreader reads "unticked" on `indeterminate` state
**Component:**
Indeterminate Checkbox

**Description of bug:**
screen readers read unticked on `indeterminate` state

**Steps to reproduce:**
Navigate to https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/indeterminate-checkbox--default 
test with screenreader 

Relates to https://github.com/cengage/react-magma/issues/1231

chris-cedrone-cengage:
Closes #1231 

## 🐢 What I did 🐢
Restructured DOM to recognize the "mixed" state when Voice Over tabs over indeterminate checkboxes

## 📺 Screenshots: 📺
Example

## ✅ Checklist ✅
- [N/A] I have made corresponding changes to the documentation
- [N/A] I have added tests that prove my fix is effective or that my feature works

## 🐎 How to test 🐎
1. Open Storybook
2. Go to Indeterminate Checkbox
3. Open Voice Over
4. Navigate to a Indeterminate Checkbox with keyboard
5. Verify "mixed" state is probably read
6. Go to Datagrid > Selectable
7. Verify Indeterminate Checkbox works appropriately here
8. Go to TreeView > First Item Branch
9. Verify Indeterminate Checkbox works appropriately here

### DIFF
--- a/packages/react-magma-dom/src/components/IndeterminateCheckbox/index.tsx
+++ b/packages/react-magma-dom/src/components/IndeterminateCheckbox/index.tsx
@@ -77,7 +77,7 @@ export const IndeterminateCheckbox = React.forwardRef<
   }
 
   function handleOnKeyDown(event: React.KeyboardEvent) {
-    if (event.key === ' ') {
+    if (event.key === ' ' && !disabled) {
       event.preventDefault();
       const syntheticEvent = {
         ...event,
@@ -173,7 +173,7 @@ export const IndeterminateCheckbox = React.forwardRef<
           tabIndex={-1}
         />
         <div
-          tabIndex={0}
+          tabIndex={!disabled && 0}
           role="checkbox"
           aria-checked="mixed"
           onKeyDown={handleOnKeyDown}
--- a/packages/react-magma-dom/src/components/IndeterminateCheckbox/index.tsx
+++ b/packages/react-magma-dom/src/components/IndeterminateCheckbox/index.tsx
@@ -172,43 +172,42 @@ export const IndeterminateCheckbox = React.forwardRef<
           onChange={handleChange}
           tabIndex={-1}
         />
-        <div
-          tabIndex={!disabled && 0}
-          role="checkbox"
-          aria-checked="mixed"
-          onKeyDown={handleOnKeyDown}
-        >
-          <StyledLabel htmlFor={id} isInverse={isInverse} style={labelStyle}>
-            <StyledFakeInput
-              isChecked={isChecked}
-              color={color}
-              disabled={disabled}
-              hasError={hasError}
-              hideFocus={props.hideFocus}
-              isIndeterminate={isIndeterminate}
-              isInverse={isInverse}
-              style={inputStyle}
-              theme={theme}
-              aria-hidden="true"
-            >
-              {isIndeterminate ? (
-                <IndeterminateCheckBoxIcon
-                  testId="indeterminateIcon"
-                  size={theme.iconSizes.medium}
-                />
-              ) : isChecked ? (
-                <CheckBoxIcon size={theme.iconSizes.medium} />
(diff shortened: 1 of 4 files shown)


moathabuhamad-cengage:
it does read mixed,  I was able to check the disabled checkbox in the storybook using my keyboard

image

packages/react-magma-dom/src/components/IndeterminateCheckbox/index.tsx
```diff
@@ -159,37 +170,45 @@ export const IndeterminateCheckbox = React.forwardRef<
...
-          >
-            {isIndeterminate ? (
-              <IndeterminateCheckBoxIcon
-                testId="indeterminateIcon"
-                size={theme.iconSizes.medium}
-              />
-            ) : isChecked ? (
-              <CheckBoxIcon size={theme.iconSizes.medium} />
+        <div
+          tabIndex={!disabled && 0}
```
silvalaura, line 176: I'm not sure about this tabindex here. It's changing the keyboard navigation behavior in TreeView which we don't want. Speaking of which, we should note in the testing notes to consider TreeView and DataGrid

packages/react-magma-dom/src/components/IndeterminateCheckbox/index.tsx
```diff
@@ -159,37 +170,45 @@ export const IndeterminateCheckbox = React.forwardRef<
...
+          role="checkbox"
+          aria-checked="mixed"
```
silvalaura, line 178: When we add this, we now have an input with aria-checked mixed AND a div with it. Is this okay? How do other systems handle this?

silvalaura: Also, this mixed status is hardcoded so even when it becomes checked, `mixed` is still in the dom.

silvalaura (changes requested):

packages/react-magma-dom/src/components/IndeterminateCheckbox/index.tsx
chris-cedrone-cengage: I removed the div and moved these attributes to `StyledFakeInput`.

silvalaura (commented):
1. Having tabIndex here still breaks the keyboard navigation in TreeView. With TreeView, we expect `Tab key moves to the tree and then out of the tree` and the tree should be navigated with up and down arrows. 

2. The focus around the indeterminate checkbox looks weird/different to the focus around regular checkboxes
image
image

3. Have you found a good example in the wild that does what we are trying to do? That will help us ensure we are doing this the right way.

moathabuhamad-cengage:
weird checkbox border animation / behavior 

[Screencast from 2025-02-10 14-44-28.webm](https://github.com/user-attachments/assets/71c26223-9de6-4392-9427-29fe8e395df2)

PR #1694, opened 2025-03-10
feat(treeview): improve dynamic child rendering and expandability det…

### LINKED ISSUE #1619: TreeView: Investigate why using context for data works in 4.6 but not 4.7.0
Allison Payne reported issues with seeing TreeView when using context in version 4.7.0. When she downgraded to 4.6.0, it worked.

This is NOT reproducible in CodeSandbox with the same exact code and data as Allison's: https://codesandbox.io/p/sandbox/treeview-allison-payne-l6d5cg?file=%2FResources.tsx. The tree is rendered as expected in both 4.7 and 4.6.

Slack thread: https://cengage.slack.com/archives/C8Y4FR63U/p1736781287666309

We need to investigate why this is happening and make any necessary updates. 

AC:
- Will potentially be fixed by latest changes in TreeView when they're merged.
- Verify fix when TreeView tickets are merged.
- Slack thread mentions this information.

### LINKED ISSUE #1668: TreeView: Support dynamically adding children
**Is your feature request related to a problem? Please describe.**
(From Tatyana T.)

We are using the TreeView component to display dynamic content, allowing users to add, remove, and update items. However, we faced an issue where TreeView does not re-render when adding a new child to a parent node that previously had no children.
I have prepared a CodeSandbox example demonstrating the issue:
https://codesandbox.io/p/devbox/react-magma-example-forked-p3wcrc?workspaceId=ws_8x7AiYRQ4ujcZkYVz1p3Mn

The issue occurs only when adding the first child. If the parent already has children, adding more works as expected. However, when using a keyboard to navigate the tree, newly added items are not visually indicated.

**Describe the solution you'd like**
TreeView is automatically updated when a new child is added to a parent with and without existing children.

**Acceptance Criteria**
- User can dynamically add tree items to an existing tree
- TreeView is updated automatically
- All accessibility remains the same (keyboard navigable, announcements, etc.)
- This is supported for all tree view selection types

moathabuhamad-cengage:
Closes: #1668
Closes: #1619

## What I did

* Fixed TreeView component to properly handle dynamic addition of children to initially empty parent nodes
* Added detection of expandability status changes when children are added at runtime
* Ensured expand/collapse controls appear appropriately for dynamically added children
* Added unit tests to verify the fix

## Checklist 
- [X] Pull request description is descriptive and testing steps are listed
- [X] Corresponding changes to the documentation have been made
- [X] Tests that prove the fix is effective or that the feature works have been added

## How to test

How to test

navigate to and test story/treeview--dynamic-tree-items 

or in codesanbox
- Create a TreeView with empty parent nodes (no children)
- Dynamically add child elements to previously empty parents
- Verify that expand/collapse button appears on the parent
- Verify expansion functionality works correctly
- Verify that parent nodes are properly rendered with their dynamic children


packages/react-magma-dom/src/components/TreeView/TreeView.stories.tsx
```diff
@@ -2384,3 +2384,111 @@ ComplexWithTopLevelNotSelectable.args = {
...
+  };
+
+  const handleClick2 = () => {
+    const newTree = tree.map((item, index) => {
+      if (index === 1) {
+        return {
+          ...item,
+          children: [
+            ...item.children,
+            { id: 22, name: 'New child', children: [] },
```
silvalaura, line 2462: ids need to be unique for the tree to work properly

silvalaura (changes requested):
- The newly added children cannot be selected: if I add child to a parent and then try to select it, the parent collapses and the child is still not selected.
- Existing children, when selected, collapses the entire parent: expand parent item with children, select child item, notice parent is collapsed.
- We need an example in the docs.
- Can we add the ability to change the selectable prop to the new example to easily test that it works for not selectable, single select and multi select? We also need to test this with allll the other props we offer, so we should allow them all in the example.
- Some edge cases to consider: 
-- what if the parent that has a child added to is part of the initialExpandedItems?
-- do checkParents/checkchildren/isTopLevelSelectable still work?
-- do the icons work as expected?
-- what about disabled?
-- can I dynamically add a child to a nested item?
-- prob so many more. we need to think through them and add them to testing notes too

Our TreeView has so much functionality so we need to be very careful whenever we add anything new to it.

orion-cengage:
2 questions:

1. I know these are just examples in Storybook. But does this mean that we literally only support dynamically adding top level parents, and children to those parents, or is it actually possible for the consumer to use this functionality at any level of the tree?

2. In an example where you start in this scenario:
image

If you then click "add unchecked children to the last parent", it actually adds the child as already checked. I don't think that is expected behavior.

packages/react-magma-dom/src/components/TreeView/useTreeView.ts
```diff
@@ -31,6 +31,7 @@ export interface TreeViewApi {
   showLess(): void;
   expandAll(): void;
   collapseAll(): void;
+  addNewItem(item: TreeViewItemInterface): void;
```
silvalaura, line 34: How about just calling this `addItem`?

silvalaura (commented):
- In storybook, add checked/unchecked parent does not work as expected for single select
- In the docs site, if I `add child to the last parent` while the item is expanded, it automatically closes it. Actually, it closes all expanded items in the tree

silvalaura:
Just a random thought and it doesn't have to be a new story that gets committed, but if you don't mind testing adding dynamic children on a large tree, I'm curious if there are any performance concerns?

nikitaorliak-cengage (approved):

olehnoskov-cengage (approved):

orion-cengage (approved):

silvalaura (approved):
Looks GREAT! Wonderful job @nikitaorliak-cengage & @moathabuhamad-cengage !

Issue #1686, opened 2025-03-03
SkipLink: Should not accept "to" prop


silvalaura:
**Describe the bug**
SkipLink component requires the `to` prop because it extends `HyperlinkProps`. The `to` prop gets overwritten anyway, so it shouldn't be required.

**Expected behavior**
SkipLink should not require the `to` prop.

PR #1699, opened 2025-03-12
SkipLink: Should not accept "to" prop

olehnoskov-cengage:
Closes: #1686

## What I did
Made the `to` prop optional

## Checklist 

### DIFF
--- a/.changeset/fix-SkipLink-prop.md
+++ b/.changeset/fix-SkipLink-prop.md
@@ -1,5 +1,5 @@
 ---
-'react-magma-dom': minor
+'react-magma-dom': patch
 ---
 
-feat(SkipLink): Make the `to` prop optional
+fix(SkipLink): Make the `to` prop optional
--- a/.changeset/fix-SkipLink-prop.md
+++ b/.changeset/fix-SkipLink-prop.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': minor
+---
+
+feat(SkipLink): Make the `to` prop optional
--- a/packages/react-magma-dom/src/components/Hyperlink/index.tsx
+++ b/packages/react-magma-dom/src/components/Hyperlink/index.tsx
@@ -36,7 +36,7 @@ export interface HyperlinkProps
   /**
    * The href value of the link
    */
-  to?: string;
+  to: string;
   hasUnderline?: boolean;
   /**
    * Icon to display within the component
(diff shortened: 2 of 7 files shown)


packages/react-magma-dom/src/components/Hyperlink/index.tsx
```diff
@@ -36,7 +36,7 @@ export interface HyperlinkProps
   /**
    * The href value of the link
    */
-  to: string;
+  to?: string;
```
silvalaura, line 39: We need to keep the `to` prop for Hyperlink required, but it needs to be optional or just omitted for SkipLink

.changeset/fix-SkipLink-prop.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': minor
+---
+
+feat(SkipLink): Make the `to` prop optional
```
silvalaura, line 5: This is just a bug fix, so it's not a minor change.

```suggestion
---
'react-magma-dom': patch
---

fix(SkipLink): Make the `to` prop optional
```

olehnoskov-cengage: Thanks, didn't notice it )

packages/react-magma-dom/src/components/Hyperlink/index.tsx
olehnoskov-cengage: Oh, got it!
Refactored.

packages/react-magma-dom/src/components/SkipLink/index.tsx
```diff
@@ -29,6 +29,10 @@ export interface SkipLinkProps extends Omit<HyperlinkProps, 'children'> {
    * @internal
    */
   testId?: string;
+  /**
+   * The href value of the link
+   */
+  to?: string;
```
silvalaura, line 35: Actually, I think just omitting it above is sufficient and we don't need this. Looking below, we always set `to={`#${TARGET_ID}`}` and wouldn't use this prop at all.

We also need to update any SkipLink example in docs or storybook that uses the `to` prop.

olehnoskov-cengage: Got it!
Updated.

silvalaura (approved):

silvalaura (approved):

nikitaorliak-cengage (approved):

Issue #1456, opened 2024-09-25
Tooltip: Replace Popper with Floating UI


silvalaura:
We are now using [Floating UI](https://floating-ui.com/) for viewport detection in Combobox, Select, Dropdown and DatePicker (https://github.com/cengage/react-magma/pull/1451). We should update Tooltip to use this same library as Popper is no longer supported. 

**Acceptance Criteria**
- No breaking changes for our users
- Make any notes if there are any major differences with behavior
- Tooltips continue to behave the same way
- Popper is removed as a dependency from react-magma-dom

PR #1708, opened 2025-03-18
fix(tooltip): replace popper with floating-ui

nikitaorliak-cengage:
Closes: #1456 

## What I did

Replaced `Popper` with `Floating-UI`

## Checklist 

## How to test

Open Tooltip -> Check that everything works fine (as before)


olehnoskov-cengage (approved):

packages/react-magma-dom/src/components/Tooltip/index.tsx
```diff
@@ -239,26 +256,31 @@ export const Tooltip = React.forwardRef<any, TooltipProps>((props, ref) => {
...
       {isVisible && (
-        <div
-          ref={setPopperElement}
-          style={combinedTooltipStyles}
-          {...attributes.popper}
-        >
+        <div ref={refs.setFloating} style={combinedTooltipStyles}>
+          <FloatingArrow
+            ref={arrowElement}
+            data-testid="tooltipArrow"
```
silvalaura, line 262: I prefer testIds to be unique. Can we use `${testId}-arrow` or something like that?

.changeset/fix-tooltip-replace-popper-with-floating-ui.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Tooltip): Replace Popper with Floating UI.
```
silvalaura, line 5: Maybe we should make a note here that the default tooltip `arrowSize` has changed in case adopters are using a custom size in their theme.

silvalaura (commented):
- Can we now remove `react-popper` from our package.jsons?
- Do we have any examples where the arrow is customized in some way that we can prove that this update doesn't break anything for that use case?

orion-cengage (approved):

silvalaura (approved):

chris-cedrone-cengage (approved):

Issue #1573, opened 2024-11-15
Dropdown > Expandable Menu Items > Icon missing at parent level


chris-cedrone-cengage:
**Describe the bug**

Initially from: https://github.com/cengage/react-magma/issues/1014

_**-- 1.5 If you include icons on the parent level, the children do not also have icons.**_

![235749135-982996c9-982e-413d-a52d-cd9865094837](https://github.com/user-attachments/assets/b824bccd-5e8e-4e65-a4db-6c22ee760691)

**To Reproduce**
1. Go to https://codesandbox.io/p/sandbox/dtffrn?file=%2Fexample.tsx
2. In the code, see Line 54
3. There's an error having an icon at this level

**Expected behavior**
At the moment, the `DropdownExpandableMenuListItem` as a parent doesn't allow for this prop to be used and it should be available.
If there are children to this parent with an icon, the children do not have the allowance for the same prop.

**Screenshots**
<img width="877" alt="Bug" src="https://github.com/user-attachments/assets/ac7423c9-89a2-45a5-bac3-97cb8807b879">

nikitaorliak-cengage:
**Test link:** 
https://storybook-preview-1710--upbeat-sinoussi-f675aa.netlify.app/, https://docs-preview-1710--upbeat-sinoussi-f675aa.netlify.app/

**Tested on:** 
Mac OS Sequoia in Chrome, Firefox, and Safari
Windows 10/11 in Chrome and Firefox

**Bugs:**
Everything is working as expected. No bugs found.

silvalaura:
@nikitaorliak-cengage when QAing a ticket, please be sure to use the dev links (https://next--upbeat-sinoussi-f675aa.netlify.app/ and https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/accordion--default) instead of the PR previews. This will ensure the changes work in conjunction with all other merged changes.

PR #1710, opened 2025-03-19
fix(Dropdown): Expandable Menu Items > Icon missing at parent level

olehnoskov-cengage:
Closes: #1573

## What I did
- Added the `icon` prop for `DropdownExpandableMenuListItem` component.

## Screenshots
React Magma Docs

Screenshot from 2025-03-19 17-23-43

Storybook
Screenshot from 2025-03-25 16-29-47

## Checklist 

## How to test
- Go to `Storybook`
- Open Dropdown -> Dropdown Expandable Menu List Item With Icons :
- for Specialty Pizza -> Capricciossa -> should be invisible icon (because is inside list of items)
- for Cheese Pizza -> should be visible icon

### DIFF
--- a/.changeset/fix-Dropdown-add-icon-parent-level.md
+++ b/.changeset/fix-Dropdown-add-icon-parent-level.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Dropdown): Added icon prop for DropdownExpandableMenuListItem
--- a/packages/react-magma-dom/src/components/Dropdown/Dropdown.stories.tsx
+++ b/packages/react-magma-dom/src/components/Dropdown/Dropdown.stories.tsx
@@ -700,8 +700,10 @@ export const DropdownExpandableMenuListItemWithIcons = args => {
         </DropdownExpandableMenuListItem>
         <DropdownDivider />
         <DropdownExpandableMenuListItem icon={<LocalPizzaIcon />}>
-          Pizza seafood (no icon)
-          <p style={{ marginLeft: '10px' }}>children</p>
+          Pizza seafood
+          <DropdownExpandableMenuListItem icon={<LocalPizzaIcon />}>
+            Child (no icon)
+          </DropdownExpandableMenuListItem>
         </DropdownExpandableMenuListItem>
       </DropdownContent>
     </Dropdown>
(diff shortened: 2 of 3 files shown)


silvalaura (changes requested):
The interesting part about this task is that we expect `DropdownExpandableMenuListItem` to accept the icon prop ONLY if it is a parent item. 

In the example below, I should be able to pass the Pizza Icon to Pizza, but not to the `I should not have an icon` element.
image

So the work here is correct so far, the icon prop _is_ allowed, but we need to only display the icon IF the component doesn't have children / is not a child.

chris-cedrone-cengage:
Example

Could this example have a child menu item under the "Child (no icon)" parent for clarity?

silvalaura (commented):
Just want to clarify the expectation for this component:


This isn't a real way someone would use the component:
```
<DropdownExpandableMenuListItem icon={<LocalPizzaIcon />}>
          Pizza seafood
          <DropdownExpandableMenuListItem icon={<LocalPizzaIcon />}>
            Child (no icon)
          </DropdownExpandableMenuListItem>
        </DropdownExpandableMenuListItem>
```

`DropdownExpandableMenuListItem` should never have children-- `DropdownExpandableMenuButton` would be the component used as a parent, and `DropdownExpandableMenuListItem` would always either be a child OR it could be a sibling to `DropdownExpandableMenuButton` in which case that's when it's allowed to have an icon.

I think this would be sufficient to show that:
Capricciosa gets an icon but it's not visible, yet Cheese Pizza also gets an icon but it **is** visible:
```
 <DropdownContent>
        <DropdownExpandableMenuGroup isMulti={false} defaultIndex={0}>
          <DropdownExpandableMenuItem>
            <DropdownExpandableMenuButton icon={<LocalPizzaIcon />}>
              Specialty Pizza
            </DropdownExpandableMenuButton>
            <DropdownExpandableMenuPanel>
              <DropdownExpandableMenuListItem>
                Margherita
              </DropdownExpandableMenuListItem>
              <DropdownExpandableMenuListItem icon={<LocalPizzaIcon />}>
                Capricciosa
              </DropdownExpandableMenuListItem>
            </DropdownExpandableMenuPanel>
          </DropdownExpandableMenuItem>
        </DropdownExpandableMenuGroup>
        <DropdownDivider />
        <DropdownExpandableMenuListItem icon={<LocalPizzaIcon />}>
          Cheese Pizza
        </DropdownExpandableMenuListItem>
      </DropdownContent>
```

orion-cengage (approved):

silvalaura (approved):

chris-cedrone-cengage (approved):

Issue #1637, opened 2025-01-29
TreeView: Support `additionalContent` prop


silvalaura:
In order to support the following design, the TreeView component needs to support adopters passing extra data to a TreeItem:
![Image](https://github.com/user-attachments/assets/b7b09c63-ea34-4df3-9968-31deceb85afc)

I have built out this example in CodeSandbox: https://codesandbox.io/p/sandbox/caldera-table-of-contents-tree-poc-j94dvx. I am achieving that look by having all the content passed to the `label` prop. Although this looks like it works fine, the problem is that we call both `event.preventDefault()` and `event.stopPropagation()` when the enter and space keys are pressed, and therefore the additional elements are not keyboard accessible. 

**Acceptance Criteria**
- An adopter can pass `additionalContent` React.Node that is displayed below the label
- The additionalContent could contain anything
- When navigating the Tree, pressing the up and arrow keys highlight the entire element (label and additional content all together). Pressing tab would then navigate through the items in that item.
- If additional content contains buttons/dropdowns/any interactive component, they must be accessible with keyboard navigation 
- There is bottom margin on the additional content (Orion to provide the size)
- Add illustration in the docs about what this prop can do/space that takes up/etc

**Test Cases to consider**
- additionalContent is a React node that contains a button. Navigating the tree with a keyboard allows me to reach that button and the action gets called when pressing enter or space.

**Notes**
- This is needed for Caldera, their epic is committed to 2025 Q2 so we need to deliver this by the beginning of May

<details>
<summary>Code Sample</summary>
```
import React from "react";
import {
  TreeView,
  TreeItem,
  Heading,
  TreeViewSelectable,
  Hyperlink,
  ButtonSize,
  Dropdown,
  magma,
  DropdownButton,
  DropdownMenuItem,
  DropdownContent,
  ButtonColor,
  Paragraph,
  ButtonGroup,
  TypographyVisualStyle,
  HyperlinkIconPosition,
  Flex,
  FlexBehavior,
  ButtonShape,
  FlexJustify,
  TypographyColor,
  FlexWrap,
} from "react-magma-dom";
import {
  SpaceDashboardIcon,
  VerifiedIcon,
  OpenInNewIcon,
  BookIcon,
  MoreHorizIcon,
  EditIcon,
  QuizIcon,
  QueuePlayNextIcon,
} from "react-magma-icons";

export function Example() {
  const treeLabel = () => {
    return (
      <Flex
        behavior={FlexBehavior.container}
[...]

PR #1736, opened 2025-04-04
feat(Treeview): Support additionalContent prop.

olehnoskov-cengage:
Closes: #1637

## What I did
- Added the new `additionalContent` prop for `TreeItem` component
- Fixed a bug causing `TreeItemProps` not to display correctly in `React Magma Docs`.

## Screenshots
React Magma Docs

Screenshot from 2025-04-04 12-55-12
Screenshot from 2025-04-04 13-45-29

Storybook

Screenshot from 2025-04-04 15-48-44

## Checklist 

## How to test
- Go to Storybook
- Open Treeview -> Complex with Additional Content -> inside some TreeItem components should be showed content

- Go to React Magma Docs
- Open Components -> Treeview -> additionalComponent
- Take a look at the TreeItemProps section as well.

### DIFF
--- a/.changeset/feat-TreeView-support-additionlContent-prop.md
+++ b/.changeset/feat-TreeView-support-additionlContent-prop.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': minor
+---
+
+feat(TreeView): Add additionalContent prop.
--- a/packages/react-magma-dom/src/components/TreeView/TreeItem.test.js
+++ b/packages/react-magma-dom/src/components/TreeView/TreeItem.test.js
@@ -4,7 +4,6 @@ import { render, getByTestId } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { transparentize } from 'polished';
 
-import { axe } from '../../../axe-helper';
 import { magma } from '../../theme/magma';
 
 import { TreeItem, TreeView } from '.';
@@ -71,6 +70,37 @@ describe('TreeItem', () => {
 
       expect(getByTestId(`${testId}-itemwrapper`)).toHaveStyle(
         `backgroundColor: ${backgroundColor}`
+      );
+    });
+  });
+
+  describe('additional content styles', () => {
+    it('should apply default styles', () => {
+      const { getByTestId } = render(
+        <TreeItem label={labelText} testId={testId} itemId={itemId} />
+      );
+
+      expect(getByTestId(`${testId}-itemwrapper`)).toHaveStyle(
+        `flexDirection: row`
+      );
+    });
+
+    it('should apply custom styles when additional contens is provided', () => {
+      const { getByTestId, getByText } = render(
+        <TreeItem
+          additionalContent={<>Content</>}
+          label={labelText}
+          testId={testId}
+          itemId={itemId}
+        />
+      );
+
+      expect(getByText('Content')).toBeInTheDocument();
+      expect(getByTestId(`${testId}-itemwrapper`)).toHaveStyle(
+        `flexDirection: column`
+      );
+      expect(getByTestId(`${testId}-additionalcontentrapper`)).toHaveStyle(
+        `marginBottom: 16px`
       );
     });
   });
--- a/packages/react-magma-dom/src/components/TreeView/TreeItem.tsx
+++ b/packages/react-magma-dom/src/components/TreeView/TreeItem.tsx
@@ -4,19 +4,19 @@ import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import { transparentize } from 'polished';
 import {
-  FolderIcon,
   ArticleIcon,
-  ExpandMoreIcon,
   ChevronRightIcon,
+  ExpandMoreIcon,
+  FolderIcon,
 } from 'react-magma-icons';
 
 import { TreeItemContext } from './TreeItemContext';
 import { TreeViewContext } from './TreeViewContext';
 import { TreeViewSelectable } from './types';
 import {
-  UseTreeItemProps,
-  useTreeItem,
   checkedStatusToBoolean,
+  useTreeItem,
+  UseTreeItemProps,
 } from './useTreeItem';
 import { useIsInverse } from '../../inverse';
 import { ThemeInterface } from '../../theme/magma';
@@ -29,12 +29,12 @@ import {
 import { Transition } from '../Transition';
 import {
   calculateOffset,
-  TreeNodeType,
   getTreeItemLabelColor,
   getTreeItemWrapperCursor,
+  TreeNodeType,
 } from './utils';
 
-export type TreeItemProps = UseTreeItemProps;
+export interface TreeItemProps extends UseTreeItemProps {}
 
 const StyledTreeItem = styled.li<{
   theme?: ThemeInterface;
@@ -107,7 +107,6 @@ const StyledTreeItem = styled.li<{
           inset-inline-start: 0;
         }
       `}
-
     &:hover {
       background: ${props =>
         !props.isDisabled
@@ -127,7 +126,8 @@ const IconWrapper = styled.span<{
   color: ${props =>
     getTreeItemLabelColor(props.isInverse, props.isDisabled, props.theme)};
   margin-right: ${props => props.theme.spaceScale.spacing03};
-  margin-left: 0px;
+  margin-left: 0;
+
   svg {
     height: ${props => props.theme.iconSizes.medium}px;
     width: ${props => props.theme.iconSizes.medium}px;
@@ -169,6 +169,7 @@ const StyledCheckboxWrapper = styled.div<{ theme?: ThemeInterface }>`
 `;
 
 const StyledItemWrapper = styled.div<{
+  additionalContent?: React.ReactNode;
   theme?: ThemeInterface;
   selectable?: TreeViewSelectable;
   nodeType: TreeNodeType;
@@ -177,6 +178,7 @@ const StyledItemWrapper = styled.div<{
   isDisabled: boolean;
 }>`
   display: flex;
+  flex-direction: ${props => (props.additionalContent ? 'column' : 'row')};
   align-items: flex-start;
   cursor: ${props =>
     getTreeItemWrapperCursor(
@@ -186,9 +188,14 @@ const StyledItemWrapper = styled.div<{
     )};
 `;
 
+const AdditionalContentWrapper = styled.div<{ theme?: ThemeInterface }>`
+  margin-bottom: ${props => props.theme.spaceScale.spacing05};
+`;
+
 export const TreeItem = React.forwardRef<HTMLLIElement, TreeItemProps>(
   (props, forwardedRef) => {
     const {
+      additionalContent,
       children,
       icon,
       index,
@@ -315,6 +322,24 @@ export const TreeItem = React.forwardRef<HTMLLIElement, TreeItemProps>(
       selectable === TreeViewSelectable.multi &&
       (isTopLevelSelectable !== false || !topLevel);
 
+    const handleAdditionalContentKeyDown = e => {
+      if (['Enter', ' '].includes(e.key)) {
+        const interactiveElement =
+          e.target instanceof HTMLElement &&
+          e.target.closest('button, [role="button"], input, select, textarea');
+
+        if (interactiveElement) {
+          e.stopPropagation();
+
+          if (e.key === ' ' && interactiveElement.tagName !== 'BUTTON') {
+            e.preventDefault();
+          }
+
+          interactiveElement.click();
+        }
+      }
+    };
+
     return (
       <TreeItemContext.Provider value={contextValue}>
         <StyledTreeItem
@@ -337,6 +362,7 @@ export const TreeItem = React.forwardRef<HTMLLIElement, TreeItemProps>(
           onKeyDown={handleKeyDown}
         >
           <StyledItemWrapper
+            additionalContent={additionalContent}
             data-testid={`${testId ?? itemId}-itemwrapper`}
             depth={itemDepth}
             id={`${itemId}-itemwrapper`}
@@ -384,7 +410,18 @@ export const TreeItem = React.forwardRef<HTMLLIElement, TreeItemProps>(
                 )}
               </StyledCheckboxWrapper>
             ) : (
-              <>{labelText}</>
+              <>
+                {labelText}
+                {additionalContent && (
+                  <AdditionalContentWrapper
+                    theme={theme}
+                    data-testid={`${testId ?? itemId}-additionalcontentrapper`}
+                    onKeyDown={handleAdditionalContentKeyDown}
+                  >
+                    {additionalContent}
+                  </AdditionalContentWrapper>
+                )}
+              </>
             )}
           </StyledItemWrapper>
 
--- a/packages/react-magma-dom/src/components/TreeView/TreeView.stories.tsx
+++ b/packages/react-magma-dom/src/components/TreeView/TreeView.stories.tsx
@@ -3,41 +3,62 @@ import React from 'react';
 import { Meta } from '@storybook/react/types-6-0';
 import {
   ArticleIcon,
-  FolderIcon,
-  FavoriteIcon,
-  StarIcon,
+  BookIcon,
+  EditIcon,
   EmergencyIcon,
+  FavoriteIcon,
+  FolderIcon,
   KeyboardArrowDownIcon,
   KeyboardArrowUpIcon,
+  LockIcon,
+  MoreHorizIcon,
+  QueuePlayNextIcon,
+  QuizIcon,
+  SpaceDashboardIcon,
+  StarIcon,
+  VerifiedIcon,
 } from 'react-magma-icons';
 
 import {
-  Tag,
-  TagSize,
+  Accordion,
+  AccordionButton,
+  AccordionItem,
+  AccordionPanel,
   Button,
+  ButtonGroup,
+  ButtonVariant,
   Flex,
   FlexBehavior,
+  IconButton,
   IndeterminateCheckboxStatus,
-  TreeItemSelectedInterface,
-  TreeViewProps,
-  ButtonVariant,
-  ButtonGroup,
   Spacer,
   SpacerAxis,
-  Accordion,
-  AccordionItem,
-  AccordionButton,
-  AccordionPanel,
-  IconButton,
+  Tag,
+  TagSize,
+  TreeItemSelectedInterface,
+  TreeViewProps,
 } from '../..';
 import { magma } from '../../theme/magma';
-import { ButtonColor, ButtonSize } from '../Button';
+import { ButtonColor, ButtonShape, ButtonSize } from '../Button';
 import { Card } from '../Card';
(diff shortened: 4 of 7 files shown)


packages/react-magma-dom/src/components/TreeView/TreeView.stories.tsx
```diff
@@ -2384,3 +2405,305 @@ ComplexWithTopLevelNotSelectable.args = {
...
+    );
+  };
+
+  const additionalContent = () => {
+    return (
+      <div style={{ width: '100%' }}>
+        <Paragraph noTopMargin visualStyle={TypographyVisualStyle.bodyXSmall}>
+          Due: xx/xx/xxx · Submitted: 12 · Missing: 3
+        </Paragraph>
+        <ButtonGroup style={{ marginBottom: '16px' }}>
```
silvalaura, line 2503: We can remove this spacing since now it's built into the `AdditionalContentWrapper`

silvalaura (changes requested):
Visually this is looking great, and the keyboard navigation works as expected. The challenge with this task is that we call both `event.preventDefault()` and `event.stopPropagation()` when the enter and space keys are pressed, and therefore the additional elements can't be pressed or opened via keyboard

packages/react-magma-dom/src/components/TreeView/TreeItem.test.js
```diff
@@ -75,6 +74,37 @@ describe('TreeItem', () => {
     });
   });
 
+  describe('additional content styles', () => {
```
silvalaura, line 77: we should add tests for confirming that the interactive elements get called when clicked or space/enter are pressed

packages/react-magma-dom/src/components/TreeView/TreeItem.tsx
```diff
@@ -384,7 +410,18 @@ export const TreeItem = React.forwardRef<HTMLLIElement, TreeItemProps>(
                 )}
               </StyledCheckboxWrapper>
             ) : (
-              <>{labelText}</>
+              <>
+                {labelText}
+                {additionalContent && (
+                  <AdditionalContentWrapper
+                    theme={theme}
+                    data-testid={`${testId ?? itemId}-additionalcontentrapper`}
```
silvalaura, line 418:
```suggestion
                    data-testid={`${testId ?? itemId}-additionalcontentwrapper`}
```

moathabuhamad-cengage:
I know this doesn't have a `ready for review` label yet but double check keyboard navigation, pressing `enter` does not open the dropdowns, but pressing `space` does

silvalaura (commented):
- I can open the Resources and Standards dropdown with `space` but not with `enter`
- I cannot open the overflow menu with either `space` or `enter`
- Pressing the Click button with either `space` or `enter` gets called twice (two logs on the console)
- Looks like the tests are still missing / were not committed

silvalaura (commented):
This is looking really good! 

The only outstanding thing I can see is that the keyboard accessibility gets out of order a little bit when you start to use tab. For example in the Storybook additional content example:

if I press tab to enter the tree view, then I press the down key to navigate to 5.1 elements, then I press tab to get to the overflow menu, now I'm stuck tabbing through the interactive elements but I can't get back to the Tree's navigation. I am not exactly sure how I expect this to work so I will look at it some more tomorrow but just throwing it out there.

website/react-magma-docs/src/pages/api/tree-view.mdx
```diff
@@ -1397,6 +1397,120 @@ export function Example() {
 }
 ```
 
+## additionalContent
```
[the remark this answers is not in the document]
olehnoskov-cengage: Updated.
Should we update [isTopLevelSelectable](https://github.com/cengage/react-magma/blob/dev/website/react-magma-docs/src/pages/api/tree-view.mdx#:~:text=)%3B%0A%7D-,isTopLevelSelectable,-When%20set%20to) as well?

packages/react-magma-dom/src/components/TreeView/TreeView.stories.tsx
```diff
@@ -2384,3 +2405,305 @@ ComplexWithTopLevelNotSelectable.args = {
...
+              <Dropdown>
+                <DropdownButton
+                  aria-label="Extra icon example"
+                  color={ButtonColor.secondary}
+                  size={ButtonSize.small}
+                  icon={<MoreHorizIcon />}
+                  shape={ButtonShape.fill}
+                />
+                <DropdownContent>
+                  <DropdownMenuItem icon={<EditIcon aria-hidden />}>
```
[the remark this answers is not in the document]
olehnoskov-cengage: Added.

website/react-magma-docs/src/pages/api/tree-view.mdx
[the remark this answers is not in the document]
silvalaura: Yes please, good catch!

orion-cengage (changes requested):
1. I'm not sure if this is only because of the varying font sizes in this example, or if this is just making an existing thing more obvious, but there are some visual alignment issues on a number of these tree items. We need to be doing whatever we have to to keep the text vertically centered with the icons and check boxes, and that includes if the label is long enough to wrap to a second or third line. That means the top line of the label needs to be centered with those icons and check boxes.

Part of the solution could be changing the line-height of the text labels, regardless of size to 24px, which would then match up with the height of the icons on the left. Also I don't think we want any of these labels to be wrapped in paragraphs. That's the not the appropriate place to use a paragraph and it brings a bunch of styling you then have to override.

image

image

image

image

3. Why is there no space between the icon and label in these dropdown buttons? 
image

4. I doubt this is the intended layout result when reducing the size of the window, but is this on us, or the consumer?
image

orion-cengage (approved):

orion-cengage:
> @orion-cengage I'm gonna hop in here because I wrote the original example:
> 
> 1. I believe the original requirements stated that everything would be top aligned because of the possibility of labels with two lines, but it's particularly obvious with the different size labels. I agree that it can look odd though, so we can change this if you prefer. The labels are not paragraphs by default, though they are in this example because I was styling them.
> 2. `Why is there no space between the icon and label in these dropdown buttons?` because our dropdown buttons don't technically support two icons, so I had to shove the left one in there. I will add better spacing when this gets implemented in the app. Not really a magma concern unless we want to support 2 icons
> 3. This is on the consumer and again, it was a POC and a quick example; will be done correctly when fully implemented. We can get rid of all the clutter for the sake of the magma example if we want.

Thanks @silvalaura . This + our conversation has resolved my issues. I'll create a ticket to get the left icon prop added to the Dropdown buttons.

silvalaura:
Previously, if there were interactive elements in the tree, pressing the left arrow would bring focus back on the tree: for example in https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/treeview--flat-tree), if I bring focus into the tree, then tab to focus the action button, I can press the left arrow key to come back to focus on the tree. It looks like this no longer works on https://storybook-preview-1736--upbeat-sinoussi-f675aa.netlify.app/?path=/story/treeview--flat-tree.

I think this is what's missing from this implementation. What do you think?

chris-cedrone-cengage (approved):

silvalaura:
When swapping from single to multi select, the additional content disappears. Is this just an issue with the storybook example, or is this always the case?
image
image

silvalaura (commented):
I don't have a solution yet (will keep looking on Monday), but some things I found in the meantime:

**Multi select**
 `checkboxProps` has `hideFocus` and `tabIndex`: I wonder if we need to allow the focus for checkboxes if the user is "isInsideTreeItem"?

**Single select**
I am not sure.. it's interesting because it's not just because we're using hyperlink; it has double focus even with a native anchor, or if the label text is not interactive, then the overflow menu gets the double tab 🤔 

New bug to consider: 
- Pressing space to open the overflow menu selects the item on single select
- The first time you press ESC on an item, it doesn't escape, so it requires two presses (probably related to single select issue above)

olehnoskov-cengage:
> I don't have a solution yet (will keep looking on Monday), but some things I found in the meantime:
> 
> **Multi select** `checkboxProps` has `hideFocus` and `tabIndex`: I wonder if we need to allow the focus for checkboxes if the user is "isInsideTreeItem"?
> 
> **Single select** I am not sure.. it's interesting because it's not just because we're using hyperlink; it has double focus even with a native anchor, or if the label text is not interactive, then the overflow menu gets the double tab 🤔
> 
> New bug to consider:
> 
> * Pressing space to open the overflow menu selects the item on single select
> * The first time you press ESC on an item, it doesn't escape, so it requires two presses (probably related to single select issue above)

For first bug - as I can see it's default behavior because of [this peace of code](https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/TreeView/TreeItem.tsx#L266)

silvalaura:
> > I don't have a solution yet (will keep looking on Monday), but some things I found in the meantime:
> > **Multi select** `checkboxProps` has `hideFocus` and `tabIndex`: I wonder if we need to allow the focus for checkboxes if the user is "isInsideTreeItem"?
> > **Single select** I am not sure.. it's interesting because it's not just because we're using hyperlink; it has double focus even with a native anchor, or if the label text is not interactive, then the overflow menu gets the double tab 🤔
> > New bug to consider:
> > 
> > * Pressing space to open the overflow menu selects the item on single select
> > * The first time you press ESC on an item, it doesn't escape, so it requires two presses (probably related to single select issue above)
> 
> @silvalaura For the first bug, as I can see , it's the default behavior because of [this piece of code](https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/TreeView/TreeItem.tsx#L266) Should we change this behaviour?

Yes. Pressing space should select the item in single select if the focus is on the entire item, and not if it's `isInsideTreeItem`

silvalaura (commented):
It's looking **really** good!

Just two things I noticed: 
1. I want to try is adding a button or any interactive element at the bottom of the `Complex with Additional Content` example so I can be sure that we can get the focus in and out of the TreeView as expected
2. https://jam.dev/c/d1ab9295-fff1-4682-9e55-4850eedfef58 I noticed that for items like `5.2 Water (pp. 137-141)` which is a parent, when tabbing to the overflow menu, it goes to it's first child's interactive elements instead of being "stuck" there like it does for items that don't have children. You can also see that with `Expanded Tree Item with Button` where the focus goes to `Button Label`'s checkbox then pressing tab takes you to the `Button` and then pressing tab again goes to `5.2 Water (pp. 137-141)` checkbox. I think we just need to isolate the focus lock to the element itself without its children if possible?

olehnoskov-cengage:
> It's looking **really** good!
> 
> Just two things I noticed:
> 
> 1. I want to try is adding a button or any interactive element at the bottom of the `Complex with Additional Content` example so I can be sure that we can get the focus in and out of the TreeView as expected
> 2. https://jam.dev/c/d1ab9295-fff1-4682-9e55-4850eedfef58 I noticed that for items like `5.2 Water (pp. 137-141)` which is a parent, when tabbing to the overflow menu, it goes to it's first child's interactive elements instead of being "stuck" there like it does for items that don't have children. You can also see that with `Expanded Tree Item with Button` where the focus goes to `Button Label`'s checkbox then pressing tab takes you to the `Button` and then pressing tab again goes to `5.2 Water (pp. 137-141)` checkbox. I think we just need to isolate the focus lock to the element itself without its children if possible?

@silvalaura 
I a couple of questions:

1. What conditions must be met to ensure that focus can reliably move into and out of the TreeView, without getting trapped inside a TreeItem or its interactive elements?

2. When focus is trapped inside a TreeItem, navigation keys like `Home` and` End` don’t work as expected — we can't move focus to the first or last TreeItem. These keys only work when the entire TreeItem is focusable. Is it okay?

silvalaura:
> > It's looking **really** good!
> > Just two things I noticed:
> > 
> > 1. I want to try is adding a button or any interactive element at the bottom of the `Complex with Additional Content` example so I can be sure that we can get the focus in and out of the TreeView as expected
> > 2. https://jam.dev/c/d1ab9295-fff1-4682-9e55-4850eedfef58 I noticed that for items like `5.2 Water (pp. 137-141)` which is a parent, when tabbing to the overflow menu, it goes to it's first child's interactive elements instead of being "stuck" there like it does for items that don't have children. You can also see that with `Expanded Tree Item with Button` where the focus goes to `Button Label`'s checkbox then pressing tab takes you to the `Button` and then pressing tab again goes to `5.2 Water (pp. 137-141)` checkbox. I think we just need to isolate the focus lock to the element itself without its children if possible?
> 
> @silvalaura I have a couple of questions:
> 
> 1. What conditions must be met to ensure that focus can reliably move into and out of the TreeView, without getting trapped inside a TreeItem or its interactive elements?
> 2. When focus is trapped inside a TreeItem, navigation keys like `Home` and` End` don’t work as expected — we can't move focus to the first or last TreeItem, because in case of being inside `TreeItem` we change the `onKeyDown` handler. These keys only work when the entire TreeItem is focusable. Is it okay?

1. I think the expectation for moving focus out of the TreeView is what the existing behavior on dev is: you tab into the tree once, and pressing tab again moves focus out of it.
2. Mmm, good point. We do need Home and End to continue to work, but maybe it's okay that it doesn't while you're focus locked "inside" one of the elements. Maybe we can get @orion-cengage 's opinion on this one.

silvalaura:
Decisions from meeting with Orion:

1. We will use control/command + enter to get focus into an item (https://www.telerik.com/kendo-angular-ui/components/treeview/keyboard-navigation)
2. Lets remove the focus from checkboxes in multi select -> getting focus into the item goes to the next interactive element (example: text). This will also fix the current issue where simple trees with multi select allow tabbing in which is odd
3. Tabbing will allow you to get out of the tree completely (like the existing dev behavior)
4. Children items should not be part of the focus lock

olehnoskov-cengage:
@silvalaura 
> Decisions from meeting with Orion:
> 
> 1. We will use control/command + enter to get focus into an item (https://www.telerik.com/kendo-angular-ui/components/treeview/keyboard-navigation)
> 2. Lets remove the focus from checkboxes in multi select -> getting focus into the item goes to the next interactive element (example: text). This will also fix the current issue where simple trees with multi select allow tabbing in which is odd
> 3. Tabbing will allow you to get out of the tree completely (like the existing dev behavior)
> 4. Children items should not be part of the focus lock

Fixed all of these issues, you can test it.
The one last thing that I need to fix is disappearing focus in `multi` mode.

Example:
[Screencast from 22.05.25 17:52:32.webm](https://github.com/user-attachments/assets/bd59584d-d7d1-44b2-9258-b0f1cbe38b5b)

packages/react-magma-dom/src/components/TreeView/TreeItem.tsx
```diff
@@ -162,13 +164,19 @@ const StyledExpandWrapper = styled.div<{
...
   margin-right: ${props => props.theme.spaceScale.spacing03};
   vertical-align: middle;
-  display: inline-flex;
+  display: ${props => (props.hasAdditionalContent ? 'flex' : 'inline-flex')};
+  flex-direction: column;
+  width: ${props => `calc(100% - ${props.theme.spaceScale.spacing03})`};
 `;
 
 const StyledItemWrapper = styled.div<{
+  additionalContent?: React.ReactNode;
```
silvalaura, line 179: What gets passed here could just be a boolean `hasAdditionalContent?: boolean` since the styled component doesn't need the full react node

packages/react-magma-dom/src/components/TreeView/TreeItem.tsx
```diff
@@ -242,6 +256,107 @@ export const TreeItem = React.forwardRef<HTMLLIElement, TreeItemProps>(
...
+      return Array.from(
+        container.querySelectorAll<HTMLElement>(selector)
+      ).filter(
+        el =>
+          !el.hasAttribute('tabindex') ||
+          (el.tabIndex !== undefined && el.tabIndex >= 0)
+      );
+    };
+
+    const handleLabelAndAdditionalContentKeyDown = (
```
silvalaura, line 278: Can we add a comment explaining what this is all doing? I understand it today but we'll want to remember in the future why we built it this way

packages/react-magma-dom/src/components/TreeView/TreeItem.tsx
```diff
@@ -315,6 +432,46 @@ export const TreeItem = React.forwardRef<HTMLLIElement, TreeItemProps>(
       selectable === TreeViewSelectable.multi &&
       (isTopLevelSelectable !== false || !topLevel);
 
+    const onKeyDownHandler = (event: React.KeyboardEvent) => {
```
silvalaura, line 435: Lets also add a comment to this function

packages/react-magma-dom/src/components/SelectionControls/StyledLabel.tsx
```diff
@@ -23,6 +23,7 @@ const StyledLabelComponent = styled.label<StyledLabelProps>`
   line-height: ${props => props.theme.typeScale.size03.lineHeight};
   margin: 0;
   padding: ${props => props.theme.spaceScale.spacing03} 0;
+  width: 100%;
```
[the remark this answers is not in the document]
olehnoskov-cengage: Yes, because for having the posibility for displaying in `TreeViewSelectable.multi` I added this style.
For `TreeViewSelectable.single and off` we have `StyledLabelWrapper` with `width: 100%`.

silvalaura (commented):
The animation from the checkbox is coming from `DisplayInputActiveStyles`. We've been trying to get away from unnecessary navigations so I wonder if @orion-cengage would allow us to fully remove it from all checkboxes

orion-cengage:
> The animation from the checkbox is coming from `DisplayInputActiveStyles`. We've been trying to get away from unnecessary navigations so I wonder if @orion-cengage would allow us to fully remove it from all checkboxes. Or, if we just need to make those styles more specific in `StyledFakeInput`
> 
> ```
> ${HiddenInput}:not(:disabled):active + label & {
>     &:after {
>       ${DisplayInputActiveStyles}
>     }
>   }
> ```

Yeah, you can take the animation off the checkbox, and radio too while you're in there.

.changeset/feat-TreeView-support-additionlContent-prop.md
silvalaura, line 1: Can we add one more changeset stating that we removed animations from checkbox and radio buttons? just so adopters can refer to this if they are wondering

packages/react-magma-dom/src/components/TreeView/TreeItem.tsx
```diff
@@ -334,19 +515,21 @@ export const TreeItem = React.forwardRef<HTMLLIElement, TreeItemProps>(
           selected={selectedItem}
           theme={theme}
           tabIndex={tabIndex}
-          onKeyDown={handleKeyDown}
+          onKeyDown={onKeyDownHandler}
+          ref={treeItemRef}
```
silvalaura, line 519: Oooh! It looks like this may help with this new bug I just opened, yay https://github.com/cengage/react-magma/issues/1800

silvalaura (approved):
Looks great!! Feel free to merge after pulling dev/fixing conflicts and adding the additional changeset. 

Great job Oleh 🎉

Issue #1737, opened 2025-04-07
Dropdown: Displaying dropdown content under other elements.


olehnoskov-cengage:
**Describe the bug**
The `Dropdown` component is displayed under another component.
Now, the `DropdownContent` has [z-index: 2](https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/Dropdown/DropdownContent.tsx#L96), which was added to make sure that the DropdownContent is displayed above other elements.

**The issue:** In case when another component has `z-index` more than 2, it will be displayed above of `Dropdown` component.

**Images:**
![Image](https://github.com/user-attachments/assets/2d33c5a3-fb3c-4535-8c57-ced1fd8703d5)

![Image](https://github.com/user-attachments/assets/0e25d51b-67ec-4fa0-a072-773eb8fbbaad)

**Acceptance criteria**
`Dropdown` should always display on top of other elements.

Link to [Slack](https://cengage.slack.com/archives/C8Y4FR63U/p1744039496757689)

silvalaura:
This was also reported to me for caldera-ui and mts-ui, so we'll want to fix it asap

nikitaorliak-cengage:
**Test links:** 
https://next--upbeat-sinoussi-f675aa.netlify.app/ and https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/accordion--default

**Tested on:** 
Mac OS Sequoia in Chrome, Firefox, and Safari
Windows 10/11 in Chrome and Firefox

**Bugs:**
Everything is working as expected. No bugs found.

PR #1739, opened 2025-04-08
fix(Dropdown): Displaying dropdown content under other elements.

olehnoskov-cengage:
Closes: #1737

## What I did
- Updated the z-index value for `DropdownContent` to ensure it is displayed above all other elements.
- Added test and updated snapshot.

## Checklist 

## How to test
- To verify it, you need to add some content for Dropdown into storybook.
- Go to storybook -> Dropdown -> get your updated example ->  DropdownContent should be displayed above all suggested components.

### DIFF
--- a/.changeset/fix-Dropdown-not-displayed-above-elements.md
+++ b/.changeset/fix-Dropdown-not-displayed-above-elements.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Dropdown): Make to display Dropdown content above other elements.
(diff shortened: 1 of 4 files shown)


olehnoskov-cengage:
> Before we just change the z-index, I want to understand the structure changes in the component. It looks like:
> 
> 1. we added a new `div` inside the Container div since 4.4.0
> 2. There's a new `<div style="position: absolute; left: 0px; top: 0px; z-index: 2; transform: translate(0px, 42px);">` for the dropdown itself. Is this because of popper?
> 
> image

Yes, we added this new `<div>`, which is now controlled by the `Floating-UI` library.
Almost all of these styles were added by this library except only `z-index` style.

https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/Dropdown/DropdownContent.tsx#L96

nikitaorliak-cengage:
Our Popover uses FLoating-UI and `z-index: 2` too. Can we clarify that we don't need the same change for it? (File: PopoverContent.tsx)

olehnoskov-cengage:
> Our Popover uses FLoating-UI and `z-index: 2` too. Can we make sure we don't need the same change for it? (File: PopoverContent.tsx)

I think it's a good catch!
Because all components that have z-index higher than **2** will be above the **Popover** component.

moathabuhamad-cengage (approved):

chris-cedrone-cengage (approved):

Issue #1721, opened 2025-03-25
Input: Does not update as expected


silvalaura:
**Describe the bug**
The logic of minutes field is: if manual minutes input value > 60, than update hours accordingly and left only minutes < 60
Minutes field works OK usually, but when:
NEW minutes value should be the same as previous minutes value input, but with updated hours, than minutes not updated accordingly.
E.g. 66, 133, 333, 666 etc....
Logic which not works for 66:
a) 6 in edit field
b) we add one more 6 to edit field (so to be 66) => so hours should be +1 and minutes = 6
c) but actually hours is +1, but minutes left as 66 on UI (edited) 

**To Reproduce**
1. Go to https://codesandbox.io/p/sandbox/react-magma-example-forked-rqz86z?file=%2Fexample.tsx 
2. Notice that entering 66 in the RM input does not update the hour
3. Refresh the page and now enter 66 in the native input, and notice the updated hour field

**Expected behavior**
RM Input should update the same as the native input

**Additional context**
Reported by hurricane team

**AC**
- Specific to Input.
- Double check TimePicker.

**Additional important AC**
- Check other projects for successful testing: MTS-UI, Caldera-UI, Authoring-UI etc.

nikitaorliak-cengage:
**Test link:** 
https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/accordion--default

**Tested on:** 
Mac OS Sequoia in Chrome, Firefox, and Safari
Windows 10/11 in Chrome and Firefox

**Bugs:**
Everything is working as expected. No bugs found.

nikitaorliak-cengage:
**Test links:** 
https://next--upbeat-sinoussi-f675aa.netlify.app/ and https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/accordion--default

**Tested on:** 
Mac OS Sequoia/Sonoma in Chrome, Firefox, and Safari
Windows 10/11 in Chrome and Firefox

**Bugs:**
Everything is working as expected. No bugs found.

PR #1782, opened 2025-05-14
fix(Input): Does not update value as expected

olehnoskov-cengage:
Closes: #[1721](https://github.com/cengage/react-magma/issues/1721)

## What I did
- Provided a fix for this issue on the customer side.
- Added example in Storybook.
- Added additional tests of using `Input` on the customer side to track break changes.

## Screenshots
[Screencast from 11.04.25 10:50:08.webm](https://github.com/user-attachments/assets/e12404ca-caea-41db-9014-60a3f96b502c)

Storybook:
image
Screenshot from 2025-05-20 11-03-25

## Checklist 

## How to test
- Go to Storybook
- Input -> Time Input
- Type 133, 66 or 333 into the `Minutes` field (RM input)
- In case of 66, in **Hours** field should be `1` hour and in **Minutes** should be `6`
- The values of minutes and hours have to be the same for the native input component.

**On the customer side need add this line to fix this issue:**

`isHours ? setHours(currentUnitValue) : setMinutes(currentUnitValue);`

You can take a look at the [implementation](https://github.com/cengage/react-magma/pull/1782/files#diff-96fd9e1b6321b5f876d0e6926e79651a7c8f7dfa8f510e5eeda0c985ed3c6fe2R730) `onChangeTimedDuration` function 

**Why should the customer update the current Implementation?**

The main issue lies in this line: `const currentUnitValue = numericValue && numericValue >= 0 ? numericValue : 0;`

Let’s say `numericValue === 66`. Based on totalDurationMilliseconds, the parsed minutes become 6, so we call setMinutes(6). However, if the current state minutes is already 6, React sees no change and skips the re-render.

**As a result:**
1) Even though the user typed 66, the internal minutes state is 6
2) React skips the update as a no-op (minutes didn’t change)
3) The input still shows 66, but logically the time is 1h 6m, causing a mismatch.

**Why This Works Differently in Native Inputs vs. React Magma Input**

Native inputs update their displayed value immediately, independent of React state — they're "uncontrolled" in that sense. In contrast, `InputBase` and `Input` components in React Magma are state-driven and semi-controlled. Their displayed value is tied to React state and props.

If the parent component fails to re-set the value explicitly, or if React optimizes away a no-op update, the visual value may fall out of sync with the actual state — especially when parsing alters the value (like converting 66 to 6).

### DIFF
--- a/.changeset/fix-Input-not-properly-update-value.md
+++ b/.changeset/fix-Input-not-properly-update-value.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Input): Improve the logic for updating the value in the handleChange function.
--- a/packages/react-magma-dom/src/components/Input/Input.stories.tsx
+++ b/packages/react-magma-dom/src/components/Input/Input.stories.tsx
@@ -796,3 +796,151 @@ export function TimeInput() {
     </>
   );
 }
+
+export const CustomTopicsRow = ({
+  topicList,
+  isRemoveButtonDisabled,
+  shouldValidate,
+  topicTitle,
+  testTopic,
+  studyMaterialsTopic,
+  order,
+  setTopicTitle,
+  setTestTopic,
+  setStudyMaterialsTopic,
+  removeTopicRow,
+}: {
+  topicList: { reference: string; title: string }[];
+  isRemoveButtonDisabled: boolean;
+  shouldValidate?: boolean;
+  topicTitle: string;
+  testTopic?: string;
+  studyMaterialsTopic?: string;
+  order: number;
+  setTopicTitle: (title: string) => void;
+  setTestTopic: (reference: string) => void;
+  setStudyMaterialsTopic: (reference: string) => void;
+  removeTopicRow: () => void;
+}) => {
+  const getTopicTitleError = (): string | undefined => {
+    if (!topicTitle.length) return 'Enter a topic title';
+    if (topicTitle.length >= 100) return 'Title is too long';
+    return undefined;
+  };
+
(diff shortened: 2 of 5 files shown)


silvalaura:
What changes were made after reverting it? Do we know the root cause of the previous issue?

olehnoskov-cengage:
> What changes were made after reverting it? Do we know the root cause of the previous issue?

As @moathabuhamad-cengage explained to me, maybe these changes added some breaking change in other Cengage repos that's way it was reverted, I am right, Moa?

silvalaura:
> > What changes were made after reverting it? Do we know the root cause of the previous issue?
> 
> As @moathabuhamad-cengage explained to me, maybe these changes added some breaking change in other Cengage repos that's way it was reverted, I am right, Moa?

Right. We don't want to have breaking changes, so we need to investigate what about this pull request is causing our adopters to break before we merge it back.

moathabuhamad-cengage:
@olehnoskov-cengage I added a storybook example and a few tests to match the errors we saw in the adopters repo, now you can fix the ticket with these included. those tests pass on current dev but fail on this branch

image

moathabuhamad-cengage (approved):

olehnoskov-cengage:
> Just curious-- I am seeing new tests, but no changes to the input component. Do we know the root cause of what caused our adopter's tests to fail with the code changes we introduced?

Yes, @moathabuhamad-cengage suggested adding new tests to monitor breaking changes on the adopter's side.
The main issue of failing tests was that in tests we were waiting for text like `text`, but in `handleChange` during every keystroke we called this function and in result in tests we had a result like: `t`, `e`, `x`, `t` because we passed in props new value every keystroke. That's why visually it worked good but tests failed.

PR #1798, opened 2025-05-28
fix(Drawer): Fix animation for sliding in.

### LINKED ISSUE #1702: Bug: Drawer Component - Missing animation for sliding in
**Describe the bug**
When opening the Drawer component, there should be an animation that shows it sliding in from off-screen. This existed in old versions of Magma (2.5) and somehow got lost along the way.

**To Reproduce**
1. Go to ['Drawer component'](https://react-magma.cengage.com/version/4.8.0/api/drawer/)
2. Open an example drawer
3. See total lack of animation

olehnoskov-cengage:
Closes: #1702

## What I did
- Fixed sliding in animation for **Drawer**

## Checklist 

## How to test
- Go to Storybook -> Open Drawer -> Default, Site Navigation-> Click on `Show Drawer` -> Drawer should appear and disappear smoothly.


silvalaura:
> > I am not sure we want to modal to also have the animation, I think it's supposed to just be the drawer. I see now that the ticket had no clear description of what is expected, which means it should've never made it to our iteration. We should confirm with @orion-cengage what he expects when he's back tomorrow.
> 
> Yes, but some modal components are wrapped in code by `Transition`, and looking at the previous implementation, `Modal` had such an animation before.

I did notice that your implementation is just fixing broken code and keeping the same transition that was already there, but we recently removed a different animation from Modal because we are trying to reduce the number of animations in our components so I want to get more clarity on this.

orion-cengage:
> > > I am not sure we want to modal to also have the animation, I think it's supposed to just be the drawer. I see now that the ticket had no clear description of what is expected, which means it should've never made it to our iteration. We should confirm with @orion-cengage what he expects when he's back tomorrow.
> > 
> > 
> > Yes, but some modal components are wrapped in code by `Transition`, and looking at the previous implementation, `Modal` had such an animation before.
> 
> I did notice that your implementation is just fixing broken code and keeping the same transition that was already there, but we recently removed a different animation from Modal because we are trying to reduce the number of animations in our components so I want to get more clarity on this.

Aside from some fine-tuning, the animation on the drawer looks great, but Laura is correct... I didn't intend for the animation to return to the modal. The modal doesn't really have a directional aspect to it like the drawer, therefore the animation doesn't really add any value or helpful visual cues. So if we can remove the animation from the modal, that would be great.

Regarding the animation on the drawer, can we slow down the animation when it closes to essentially match what it does when it appears? It's kind of alarming when it moves that fast. I'm picturing something more like [the modal on MUI](https://mui.com/material-ui/react-drawer/#anchor).

packages/react-magma-dom/src/theme/components/reducedMotionTransition.ts
silvalaura, line 1: I think these should stay at 0, with reduced motion turned on, there should not be any animations

orion-cengage:
I'm not sure if this will help, but here's a [Figma](https://www.figma.com/proto/z2H2qv9iYQX942LqToCuQE/Components---Data-Display?page-id=2319%3A14876&node-id=2383-16070&viewport=-1714%2C439%2C0.42&t=fuiX9cleLmMnwpi2-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=2383%3A16070) link that illustrates what I want the animations to be.

1. background darkens (fade in and out)
2. Drawer slides in and out at the same speed as the background fades.

Jun-18-2025 10-46-15

orion-cengage (approved):

silvalaura (commented):
- In the all the storybook modal examples, I am no longer able to close the modal by clicking the background. It seems to work fine on the docs site. Just want to make sure we didn't break this?
- I can't tell if reduced motion is working for the drawer example. What is the expected behavior?

silvalaura (approved):

Issue #1840, opened 2025-06-30
CarbonChart: Support custom colors


silvalaura:
_This is the work resulting from #1822_ 

Typhoon team needs to be able to customize colors for their charts by specifically mapping a color to a dataset. Our current implementation prevents that from happening.

In addition, they will be using the threshold option, and we need to fix the styles so the text can be legible:
![Image](https://github.com/user-attachments/assets/06c0bac6-99fa-4e47-abf1-2c906e9e972e)
![Image](https://github.com/user-attachments/assets/71742f9d-8125-4f78-bfa8-a61348b6d188)

**Acceptance Criteria**
- CarbonCharts can have custom colors to support Typhoon's use case
- Threshold is styled for accessibility

silvalaura:
Typhoon team does not need this for now, they will do the customizations on their end.

PR #1863, opened 2025-07-24
feat(CarbonChart): Add supporting custom colors.

olehnoskov-cengage:
Closes: #1840

## What I did
Added supporting custom colors

## Checklist 

## How to test
Go to **Storybook** -> **CarbonChart** -> **Area** -> **Area Time Series with custom colors** -> example should have custom colors: `red`, `green`, and `yellow`.

### DIFF
--- a/.changeset/feat-CarbonCharts-support_custom-colors.md
+++ b/.changeset/feat-CarbonCharts-support_custom-colors.md
@@ -1,5 +1,5 @@
 ---
-'react-magma-dom': minor
+'@react-magma/charts': minor
 ---
 
 feat (CarbonCharts): Add supporting custom colors.


olehnoskov-cengage:
> My initial thought for this ticket was to support `color` in `options: {}` the way that Carbon does instead of color within each data set object (passing an array of colors). Right now we overwrite the color prop with our custom function. I want to hear your thoughts on pros/cons to doing it this way instead?
> 
> The downside I see to this solution is that here I only passed color to one object-- my other ones are not defaulting to the other colors: image

My initial thought was that user have to pass an equal number of colors and groups.
I'll refactor it this way: if the user doesn't pass an equal number of colors and groups, the remaining colors will be taken from the defaults.

moathabuhamad-cengage (changes requested):

Issue #1207, opened 2024-03-08
Date Picker: Updates to Calendar


orion-cengage:
**Purpose:** Due to future updates for things like Date Range, we need to make some updates to the Calendar dropdown and the elements within.

In progress

olehnoskov-cengage:
Need to add some additional `aria-current` and `aria-selected` attributes 
https://cengage.slack.com/archives/C8Y4FR63U/p1754042617307779

Issue #1209, opened 2024-03-11
Date Picker - Change month and year


orion-cengage:
**Purpose:** Provide ability to more quickly change the month and/or year in the Date Picker

**Figma:** https://www.figma.com/file/VoWYiiOKKppcYm5QLgPyDC/Components---Inputs?type=design&node-id=1625%3A6121&mode=design&t=NrViqpH1mAFQ3U0F-1

### 1.0 Triggering elements
-- 1.1 The Month and Year will now be clickable, allowing the user to make a selection
-- 1.2 Month and Year should actually be Native Selects, even though they are styled differently
--- 1.2.1 Month and Year will each have default, hover, and focused states
---- 1.2.1.1 The styling matches that of a subtle button with no container
-- 1.3 Aria labels
--- 1.3.1 Month: aria-label="[month]. Select a month"
---- 1.3.1.1 Example: "September. Select a month"
--- 1.3.2 Year: aria-label="[year]. Select a year"
---- 1.3.2.1 Example: "2023. Select a year"

### 2.0 Options list
-- 2.1 Use dropdown list
-- 2.2 Month
---2.2.1 List includes all 12 months
-- 2.3 Year
--- 2.3.1 List includes 1900 - 2099
-- 2.4 Restrict height of each dropdown to 240px
-- 2.5 Use Active Item Index to show which option is currently selected

![image](https://github.com/cengage/react-magma/assets/24397282/69c8b910-6ef1-4412-af4e-ed39b1edd674)

### 2.0 Options list
-- 2.1 Uses Native Select so the computer and browser will control how the list is presented
-- 2.2 Month
---2.2.1 List includes all 12 months
-- 2.3 Year
--- 2.3.1 List includes 1900 - 2099

![image](https://github.com/cengage/react-magma/assets/24397282/714a43d3-11cf-4604-a3ed-cccfc930ad19)

PR #1871, opened 2025-08-04
feat(Date Picker): Add today buttton and update styles.

olehnoskov-cengage:
Closes: #1207

## What I did
- Added new `Today` button for navigation to current date
- Updated styles for `Date Picker` component.

## Screenshots
Screenshot from 2025-08-04 16-02-14
Screenshot from 2025-08-04 16-01-59

## Checklist 

## How to test
Go to `Storybook` -> `Date Picker` -> `Default` and `Inverse` 
- `Today` button should move to the current day, and you can trigger it by clicking or pressing `Space` and `Enter`
- New styles should be applied to `Date Picker` component

[Link to Figma](https://www.figma.com/design/VoWYiiOKKppcYm5QLgPyDC/Components---Inputs?node-id=1615-15367&m=dev)

### DIFF
--- a/packages/react-magma-dom/src/components/DatePicker/HelperInformation.tsx
+++ b/packages/react-magma-dom/src/components/DatePicker/HelperInformation.tsx
@@ -121,6 +121,7 @@ const BackToCalendarWrapper = styled.span<{
   theme?: ThemeInterface;
   isInverse?: boolean;
 }>`
+  margin-top: ${props => props.theme.spaceScale.spacing03};
   font-weight: 600;
   font-size: ${props => props.theme.typeScale.size02.fontSize};
   line-height: ${props => props.theme.typeScale.size02.lineHeight};


packages/react-magma-dom/src/components/DatePicker/CalendarHeader.tsx
```diff
@@ -82,44 +89,62 @@ export const CalendarHeader = React.forwardRef<
...
           )}`}
-          icon={<ArrowBackIcon />}
+          icon={<KeyboardArrowLeftIcon />}
           type={ButtonType.button}
           variant={ButtonVariant.link}
           onClick={onPrevMonthClick}
-          style={{ margin: '6px' }}
+          style={{
+            margin: '6px',
+            color: props.isInverse
```
silvalaura, line 125: Can we use `color={ButtonColor.subtle}` instead of using a custom color?

packages/react-magma-dom/src/components/DatePicker/CalendarHeader.tsx
```diff
@@ -82,44 +89,62 @@ export const CalendarHeader = React.forwardRef<
...
       <CalendarHeaderText
+        data-testid="calendar-header"
         tabIndex={-1}
         theme={theme}
         ref={ref}
         isInverse={props.isInverse}
       >
-        <Announce>{capitalizeCurrentMonth}</Announce>
+        <Announce>
+          <MonthYearWrapper theme={theme}>
```
silvalaura, line 106: Do we want this to receive focus? I know it currently does, I just wasn't sure if we want to keep that

silvalaura (commented):
1. In the month that is not current, the numbers not within the month are the wrong font color (should be #707070 instead of #454545)
image

Maybe we should check with @orion-cengage -- what are the styles for days that are from the previous month vs. disabled dates (if min/max width are used)?

2. After I click Today, I can't get the focus to go back to the calendar numbers (the focus does go to the numbers before I click Today)

3. I had left this comment on the ticket `When we do this, we need to update our DatePicker examples in the docs and unit tests. We use string for dates when they should be of type Date which makes it hard for our adopters to have off the shelf examples.` Could we make this update in this PR?

packages/react-magma-dom/src/components/DatePicker/CalendarHeader.tsx
olehnoskov-cengage: As I can see, maybe,  it'll be changed in [Date Picker - Change month and year](https://github.com/cengage/react-magma/issues/1209) where `month` and `year` will be as `button`.

orion-cengage (changes requested):

1. For the keyboard controls button, let's change it to the medium/default sized button. The recent change in icon sizes was done way after the redesign for the date picker, and that icon is not rendering well at that size.
image

2. The close button should use the "subtle" color, not "secondary"
image

3. Left/Right buttons: Make sure the space between the buttons is only 4px. It actually looks like there's 6px of margin on all sides of the button which shouldn't be there. It should just be the 4px of space between them.
image
image

4. Use text-transform to make the days of the week all caps
image

5. Remove margin on the bottom of the table. It's creating too much space.
image

6. The days on the calendar should be 40px tall, and then they should fill up the space within the overall container. That should result in each day being approximately 43px wide. There also shouldn't be any space between the days horizontally or vertically.
image
image

7.  The numbers on the active days should not be black. They should be our default text color which uses neutral-700 (#454545)
image

olehnoskov-cengage:
> Great progress on the fixes. Just a few things:
> 
> 1. There should be an 8px space between the days of the week and the first row of dates. I was able to do this by adding 8px of bottom padding to the `th` elements.
> 
> image
> 2. The button for the Keyboard instructions should also use the "Subtle" color variant, not "Secondary"
> 
> image
> 3. The fonts are rendering a little different in the browser than they did in Figma. So for that reason, let's give a font-weight of 400 to the disabled dates, and dates that are technically part of the previous and next month. That should provide an even clearer contrast between the active dates and the others.
> 
> image

@orion-cengage 
In this case, we should also update the `Back to calendar` button to the "Subtle" color variant as well.

packages/react-magma-dom/src/components/DatePicker/CalendarDay.tsx
```diff
@@ -67,37 +99,45 @@ const CalendarDayCell = styled.td<{
...
   &:focus {
     outline: 2px solid
       ${props =>
         props.isInverse
           ? props.theme.colors.focusInverse
           : props.theme.colors.focus};
+    border: ${props => buildChosenDayBorder(props)};
+    width: ${props => props.theme.spaceScale.spacing09};
+    height: calc(${props => props.theme.spaceScale.spacing09} - 4px);
+    margin: 1px;
```
moathabuhamad-cengage, line 140: Magic numbers

packages/react-magma-dom/src/components/DatePicker/CalendarDay.tsx
```diff
@@ -67,37 +99,45 @@ const CalendarDayCell = styled.td<{
...
-  height: calc(${props => props.theme.spaceScale.spacing09} - 4px);
+  height: ${props => props.theme.spaceScale.spacing09};
   justify-content: center;
-  margin: ${props => props.theme.spaceScale.spacing01};
   overflow: hidden;
   outline-offset: 0;
   position: relative;
   transition: background 0.5s ease-in-out 0s;
-  width: calc(${props => props.theme.spaceScale.spacing09} - 4px);
+  width: calc(${props => props.theme.spaceScale.spacing09} + 3px);
```
moathabuhamad-cengage, line 129: +3px magic number

packages/react-magma-dom/src/components/DatePicker/CalendarMonth.tsx
```diff
@@ -211,35 +276,66 @@ export const CalendarMonth: React.FunctionComponent<CalendarMonthProps> = (
...
+                  theme={theme}
+                  isInverse={context.isInverse}
+                />
+                <Hyperlink
+                  aria-label={i18n.datePicker.navigateToCurrentDateAriaLabel}
+                  target="_self"
+                  to="#"
+                  hasUnderline={false}
+                >
+                  Today
```
silvalaura, line 319: i18n

packages/react-magma-dom/src/components/DatePicker/HelperInformation.tsx
```diff
@@ -83,15 +96,24 @@ const StyledContent = styled.div<{
       : props?.theme?.colors?.neutral700};
 
   h2 {
-    margin: 10px 0 12px 0;
+    margin: 16px 0 12px 0;
```
silvalaura, line 99: spacing05 0 spacing04 0

silvalaura (approved):

orion-cengage:
> https://storybook-preview-1871--upbeat-sinoussi-f675aa.netlify.app

Great catch. I actually wanted to update that button to use a link instead. Is that possible? Same styling as the "Today" link but with the icon too.

image

orion-cengage:
Regarding all the comments about 3px and 43px, those aren't meant to be hard widths. The only hard width is really on the overall picker at 320px. Within that the dates are supposed to "fill" the space to be as wide as they need to be to fill up the container and honor the 8px of padding on the edges. In Figma, this results in each date being 43.43px wide but I don't know if that's how it will exactly come to in code.

Since the dates look to be part of a table, I think you just want the width of the table to be 100%, and then each cell in each row are just equal width to fill it up. 

Hopefully that helps clarify.

Issue #1882, opened 2025-08-18
Charts: module resolution errors in @react-magma/charts


moathabuhamad-cengage:
**Describe the bug**
The `exports` field in @react-magma/charts package.json is misconfigured, causing TypeScript module resolution failures. The current configuration points to directories instead of specific files, violating the Node.js package exports specification.

**To Reproduce**
1. Create a new TypeScript project with `moduleResolution: "bundler"` and `module: "esnext"`
2. Install `@react-magma/charts` package
3. Try to import the package: `import charts from '@react-magma/charts'`
4. See TypeScript resolution errors in IDE and during compilation

**Expected behavior**
- TypeScript should successfully resolve the module types
- No TypeScript errors when importing `@react-magma/charts`
- IntelliSense and auto-completion should work properly
- The package should resolve correctly for both ESM and CommonJS consumers

**Current Configuration Issues**
Looking at the package.json:
```json
"exports": {
  ".": "./src",
  "./src/components/CarbonChart/": "./styles.min.css"
}

PR #1888, opened 2025-08-25
fix(charts): resolve bug in ChartDataTable and update package exports

moathabuhamad-cengage:
Closes: #1882 

## What I did

* Fixed logic bug in ChartDataTable: Resolved an "always falsy" expression 
* Updated package exports: Modified `package.json` exports to point to compiled CSS (./dist/charts.css) instead of source files, providing a cleaner import path for consumers.

## Screenshots

image

image

## Checklist 

### Testing the package exports:
1. Install the updated package in a consumer project
2. Import styles using: `import '@react-magma/charts/dist/charts.css';`
3. Verify that the CSS file loads correctly and styles are applied

### Components affected:
- `ChartDataTable` component - Core logic fix
- Package consumers - Improved import experience

### Edge cases to verify:
- Chart data with zero values
- Chart data with mixed zero/non-zero values  
- Chart data with `null` or `undefined` values
- CSS import in consumer applications

### DIFF
--- a/.changeset/chore-charts-imports copy.md
+++ b/.changeset/chore-charts-imports copy.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+chore: updated @react-magma/charts package to include all necessary styles directly in the package and corrected named exports for better usability, eliminating the need to import styles from /dist


silvalaura (commented):
Charts in docs don't work: 
image

Do we need to change how we import the styles in our doc page? We had a line in the examples to uncomment the styles, is this no longer needed or is this still applicable?
```
// IMPORTANT: Uncomment the line below for the correct styles
// import '@react-magma/charts/dist/charts.css';
```

moathabuhamad-cengage:
> How would adopters use this package, if differently to how it was previously? Do they not need to import styles separately? We should have instructions since this is a major change

The only difference now is that they don't need the line ` import '@react-magma/charts/dist/charts.css';` the base carbon charts styles are applied out of the box

Issue #1917, opened 2025-09-12
DatePicker: Voice over announces value months and years twice.


olehnoskov-cengage:
**Test links:** 
https://next--upbeat-sinoussi-f675aa.netlify.app/ 
https://storybook-preview-dev--upbeat-sinoussi-f675aa.netlify.app/?path=/story/accordion--default

**Tested on:** 
Mac OS Sequoia in Chrome, Firefox, and Safari
Windows 10/11 in Chrome and Firefox

**Bugs:**
Windows 10/11 + Chorme/Firefox -- NVDA announces all enabled months/years when you open a select:

<img width="1140" height="408" alt="Image" src="https://github.com/user-attachments/assets/1344cd9f-a2f5-4218-87ac-cbea51dd72ac" />

MacOS Sequoia + Safari/Firefox/Chrome -- Voiceover announces month/year twice

<img width="1171" height="600" alt="Image" src="https://github.com/user-attachments/assets/b00df0af-5309-4e08-a72f-adc317b7a54b" />

<img width="1171" height="600" alt="Image" src="https://github.com/user-attachments/assets/10a4558e-58a3-422f-bccd-3ffac6ac23e2" />

PR #1936, opened 2025-09-22
fix(DatePicker): Announcement during the change to the next or previous month.

olehnoskov-cengage:
Closes: #1917

## What I did
- Fixed announcement during the change to the next or previous month.
- Update logic of showing month in calendar header. 

## Checklist 

## How to test
Storybook -> DatePicker -> Default -> Turn on Voice Over -> Click on `next` or `previous` month button -> After changing month, VO should announce updated month and year.

### DIFF
--- a/.changeset/fix-DatePicker-add-announcement-next-previous-buttons.md
+++ b/.changeset/fix-DatePicker-add-announcement-next-previous-buttons.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix (Date Picker): Add an announcement during the change to the next or previous month.
--- a/.changeset/fix-DatePicker-add-announcement-next-previous-buttons.md
+++ b/.changeset/fix-DatePicker-add-announcement-next-previous-buttons.md
@@ -1,5 +1,0 @@
----
-'react-magma-dom': patch
----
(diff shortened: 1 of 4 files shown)


nikitaorliak-cengage:
> > Found a bug:
> > When we have `minDate`/`maxDate`, we can still use the next/previous buttons, even if we shouldn't have that option
> > DatePicker_next-previous_bug.webm
> 
> In what way it should work: it should be disabled if we reach `min` or `max` values?

We could do this in 4.9.1, but in our case we simply don't see any changes in the “Year” selection, so we should probably decide whether to change the year selection or disable the buttons

silvalaura:
> > When clicking the previous/next arrows, the month+year are announced twice
> > @silvalaura
> 
> I was investigating this issue and have a result: It happens when the user doesn't wait for the announcement of all the text for the `nextMonth` button and service information like `you are currently on a button...`.
> 
> If the user listens to all this info only once completely, and after that presses on the `nextMonth` button, the user will hear the current month and year only once.
> 
> It's only related to `macOS` systems and all browsers.

Are you calling me impatient??? 😆 You're right! I just retested.

silvalaura (approved):

PR #1990, opened 2025-10-29
feat(Date picker): Add new keyboard control (v5)

### LINKED ISSUE #1179: Date Picker: Add new keyboard controls
Originally requested in #1157 .

Description: Reinforce desired date formatting by adding native HTML behavior into the date input, allowing the user to manually input the date or use arrow keys + tab.

Based on functionality found here:

- https://react-spectrum.adobe.com/react-spectrum/DatePicker.html
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

**Figma:** https://www.figma.com/file/VoWYiiOKKppcYm5QLgPyDC/Components---Inputs?type=design&node-id=1472%3A1837&mode=design&t=Lwpr8x7ZchEKQ2YL-1

image

### AC

- [x] When you tab into the input, the focus is on the first placeholder.
- [x] You can press up or down to change the value.
- [x] The first value is equal to today's value. So if it is May, then the value is 05. If today is the 23rd, then the value is 23. If the current year is 2024, then the value is 2024.
- [x] You can switch between month, day and year placeholders using Tab or left/right arrows.
- [x] You can also type the value you want using numbers on the keyboard.
- [x] The input is not considered complete until you somehow set a value for month, day and year.

olehnoskov-cengage:
Closes: #1179 

## What I did
- Added new keyboard control for Date Picker via adding new prop: `isDateFieldInput`
- Added in docs Date Field Input example and description about the new property.
- Removed caret for input

## Screenshots
Screenshot from 2025-10-28 18-24-46

Screenshot from 2025-10-31 11-55-37

Screenshot from 2025-10-28 18-25-19

## Checklist 

## How to test
**Storybook**
 DatePicker -> Non Default Format -> update in controls `isDateFieldInput=true` -> verify supporting all date formats
DatePicker -> Date Field Default -> test all functionality and styles which are described [here](https://github.com/cengage/react-magma/issues/1179#:~:text=t%3DLwpr8x7ZchEKQ2YL%2D1-,AC,-When%20you%20tab)
DatePicker -> Inverse -> verify styles

When the date picker has `MMMM d, yyyy` format, it works in such a way:
- user presses 1 or `J` -> `January`
- user presses D(d) -> `December`
- user presses 1 -> `January` -> `2` -> `December` (1+2=12)
-  user presses `J` -> `January` ->  `2` -> `February`
- user presses `J` -> `January` -> `u` -> `June` -> `l` -> `July`.
OR
- user presses `M` -> `March` -> `a` -> `March` -> `y` -> `May`.

The current behavior was obtained here > [MUI-date-picker](https://mui.com/x/react-date-pickers/date-field/#localization)

**React Magma Docs:**
React Magma Docs -> Components -> Date Picker -> Date Field Input


website/react-magma-docs/src/pages/api/date-pickers.mdx
```diff
@@ -209,6 +209,51 @@ export function Example() {
...
+        {chosenDate && (
+          <span>
+            {`${
+              chosenDate.getMonth() + 1
+            }/${chosenDate.getDate()}/${chosenDate.getFullYear()}`}
+          </span>
+        )}
+      </p>
+      <DatePicker
+        isDateFieldInput
```
silvalaura, line 244: The description says `useDateField` but the example uses `isDateFieldInput`. I wonder what we should call this prop. Maybe `useDateMask`??? I'm not sure, what do other systems call this?

silvalaura (commented):
1. On my machine, the space for `mm` is not wide enough so it looks like `mn`
image
image

2. After the year has a value, it's really hard to change it. I accidentally typed a 1 so it defaulted to some value and then I couldn't change it to the 2000s

website/react-magma-docs/src/pages/api/date-pickers.mdx
olehnoskov-cengage: Opps!
Mistyped into docs. Should be `isDateFieldInput`. 
Updated docs.

olehnoskov-cengage:
> 1. On my machine, the space for `mm` is not wide enough so it looks like `mn`
> 
> image image
> 2. After the year has a value, it's really hard to change it. I accidentally typed a 1 so it defaulted to some value and then I couldn't change it to the 2000s

1. Updated.
2. Found an interesting issue: 
` * @param year The full year designation is required for cross-century date accuracy. If year is between 0 and 99 is used, then year is assumed to be 1900 + year.` It's described in **DateConstructor**. And when we start updating year field, date automatically was updated to 1900 + year value.

orion-cengage (changes requested):
Please address the following styling issues:

- There is extra padding on each value, causing the whole date to be spread out. Each one should hug the sides of the value and only be as wide as it needs to be for each value. Yes, this means it will potentially grow and shrink.
- The blue bottom border when the value is highlighted is too thick. It should only be 2px.
- We need to compensate for that border, because it's making the values jump up and down
- Something happened to our leading zeros on single digits. It should "03/02/2025", not "3/2/2025"

orion-cengage:
The only issue I see now is that the arrow keys don't work when the focus is on the first part of this example. I don't see this format often, but I don't think the month and day are meant to be combined like this.

image

olehnoskov-cengage:
> The only issue I see now is that the arrow keys don't work when the focus is on the first part of this example. I don't see this format often, but I don't think the month and day are meant to be combined like this.
> 
> image

You don't see any data in case using the arrow keys, because it's a `text` field, and I tried to implement this `MMMM d, yyyy` date format in new implementation.
Could you provide an example of what it might look like?

From my perspective, it can be a `text` field for month and `number` field for day input, and in result we'll have 3 inputs, as for all date formats.

orion-cengage:
> > The only issue I see now is that the arrow keys don't work when the focus is on the first part of this example. I don't see this format often, but I don't think the month and day are meant to be combined like this.
> > image
> 
> You don't see any data in case using the arrow keys, because it's a `text` field, and I tried to implement this `MMMM d, yyyy` date format in new implementation. Could you provide an example of what it might look like?
> 
> From my perspective, it can be a `text` field for month and `number` field for day input, and in result we'll have 3 inputs, as for all date formats.

[Here's a link](https://mui.com/x/react-date-pickers/date-field/#localization) to a working example on MUI. It's definitely broken into 3 parts.

image

orion-cengage:
I checked all the updates that were based on my previous feedback and I think we're looking really good.

But I did find a new issue. When you have `isDateFieldInput` turned on, you can tab forward through the input and beyond. But when you `shift + tab` to go backwards, the focus gets stuck at the first position and can't get out of the input going backwards.

orion-cengage (approved):

nikitaorliak-cengage (approved):
Looks good to me!

silvalaura:
Still seeing "mn" on Chrome. It looks fine on Safari and Firefox
image

silvalaura:
My only challenge with this functionality is that I can't ever fully clear a field. If I have a 7 on the month and I want to clear it, I press backspace. That swaps a 7 for a 1. Is this intended?

olehnoskov-cengage:
> My only challenge with this functionality is that I can't ever fully clear a field. If I have a 7 on the month and I want to clear it, I press backspace. That swaps a 7 for a 1. Is this intended?

It's strange.
Can't reproduce it.
[Screencast from 23.12.25 15:57:50.webm](https://github.com/user-attachments/assets/f0ae0038-e387-4e0d-9218-b27f94b9ceb1)


It looks like an old version, as for this and for previous issue.
Can't reproduce it.
Screenshot from 2025-12-23 16-09-17

silvalaura (approved):

PR #1995, opened 2025-11-06
fix(Table): Change logic for `tabIndex`

### LINKED ISSUE #1769: Table: Remove tabindex and update docs
**Describe the bug**
The element is gained focus via the Tab key, but is not interactive. 

**To Reproduce**
-> Go to React Magma Docs
-> Open Table -> Basic Usage
-> Move to table example via `Tab`
-> Table example is focusable

**Additional context**
[Linked spike](https://github.com/cengage/react-magma/issues/1748)
[Non-interactive content gains focus](https://perkins.caspio.com/dp/ca3640005e6f2f41f3634a75b225?ID=ID_0016668)

From US Gov site:
> **Scrollable tables need to be focusable.** When you use the .usa-table-container--scrollable variant with a table, you must add the tabindex="0" attribute to the scrollable element. This attribute assures that users navigating with a keyboard are able to select and scroll the table. tabindex="0" enables focus on elements that do not get focus by default. This attribute does not change the tab order. It places the element in the logical navigation flow.

- Example: see difference between top tables and bottom tables https://designsystem.digital.gov/components/table/#sortable-table-rows-2

**Acceptance Criteria**
- **scrollable table, no interactive elements: needs tabIndex=0 and should get focused when navigating the page**
- scrollable, with interactive elements: no tabIndex needs to be set
- non-scrollable, no interactive elements: no tabIndex needs to be set and should not get focus when navigating a page
- non-scrollable, with interactive elements: no tabIndex needs to be set and should not get focus when navigating a page

We will achieve this by removing the existing tab index AND updating our docs page to explain when users need to manually add tab index to their tables.

Once complete, test and close https://github.com/cengage/react-magma/issues/1797.

nikitaorliak-cengage:
Closes: #1769

## What I did

Change logic for Table:
- scrollable table, no interactive elements: has tabIndex=0 and gets focused
- scrollable, with interactive elements: no tabIndex
- non-scrollable, no interactive elements: no tabIndex and doesn't get focus
- non-scrollable, with interactive elements: no tabIndex and doesn't get focus

## Checklist 

## How to test

Open Storybook -> Table -> Check `tabIndex` behavior on different examples and try to resize page

### DIFF
--- a/packages/react-magma-dom/src/components/Table/Table.tsx
+++ b/packages/react-magma-dom/src/components/Table/Table.tsx
@@ -262,10 +262,10 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
       checkScrollability();
       updateActiveElements();
 
-      let resizeObserver: ResizeObserver | null = null;
+      let resizeObserver: any = null;
 
       if (typeof window !== 'undefined' && 'ResizeObserver' in window) {
-        resizeObserver = new ResizeObserver(() => {
+        resizeObserver = new (window as any).ResizeObserver(() => {
           checkScrollability();
         });
 
--- a/packages/react-magma-dom/src/components/Table/Table.tsx
+++ b/packages/react-magma-dom/src/components/Table/Table.tsx
@@ -241,6 +241,10 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
 
     const tableWrapper = `table-wrapper-${testId}`;
 
+    // - scrollable table, no interactive elements: needs tabIndex=0 and should get focused when navigating the page
+    // - scrollable, with interactive elements: no tabIndex needs to be set
+    // - non-scrollable, no interactive elements: no tabIndex needs to be set and should not get focus when navigating a page
+    // - non-scrollable, with interactive elements: no tabIndex needs to be set and should not get focus when navigating a page
     React.useEffect(() => {
       const checkScrollability = () => {
         if (tableWrapperRef.current) {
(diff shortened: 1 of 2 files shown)


nikitaorliak-cengage:
I changed the ResizesObesrver type to `any` and `(window as any).ResizeOberver`, because typescript cannot find name 'ResizeObserver'. 

Another possible solution -- install it and add it to our `lib` array in the `tsconfig.json` file. 

Which approach is better? @silvalaura

silvalaura:
Missing the part of the ticket that requests an update to our docs page: `updating our docs page to explain when users need to manually add tab index to their tables.`

packages/react-magma-dom/src/components/Table/Table.tsx
```diff
@@ -229,12 +230,59 @@ export const Table = React.forwardRef<HTMLTableElement, TableProps>(
...
     const theme = React.useContext(ThemeContext);
 
     const isInverse = useIsInverse(props.isInverse);
 
+    const tableWrapperRef = React.useRef<HTMLTableElement>(null);
+
     const tableWrapper = `table-wrapper-${testId}`;
 
+    React.useEffect(() => {
+      const checkScrollability = () => {
```
silvalaura, line 249: I wonder if we should add an inline comment with the expected functionality so we remember why we did this
```

- scrollable table, no interactive elements: needs tabIndex=0 and should get focused when navigating the page
- scrollable, with interactive elements: no tabIndex needs to be set
- non-scrollable, no interactive elements: no tabIndex needs to be set and should not get focus when navigating a page
- non-scrollable, with interactive elements: no tabIndex needs to be set and should not get focus when navigating a page
```

silvalaura:
> I changed the ResizesObesrver type to `any` and `(window as any).ResizeOberver`, because typescript cannot find name 'ResizeObserver'.
> 
> Another possible solution -- install it and add it to our `lib` array in the `tsconfig.json` file.
> 
> Which approach is better? @silvalaura

I think this is fine, does anyone else feel strongly either way?

silvalaura (approved):

olehnoskov-cengage (approved):

PR #2121, opened 2025-12-09
fix(treeview): add height prop for virtualization support

moathabuhamad-cengage:
Closes: #

## Checklist 

### DIFF
--- a/packages/react-magma-dom/src/components/TreeView/TreeView.tsx
+++ b/packages/react-magma-dom/src/components/TreeView/TreeView.tsx
@@ -22,6 +22,13 @@ export interface TreeViewProps
    * @default false
    */
   enableVirtualization?: boolean;
+  /**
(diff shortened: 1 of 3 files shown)


packages/react-magma-dom/src/components/TreeView/TreeView.tsx
```diff
@@ -308,6 +316,12 @@ export const TreeView = React.forwardRef<HTMLUListElement, TreeViewProps>(
                 }}
                 role="tree"
                 theme={theme}
+                style={{
```
olehnoskov-cengage, line 319: As I see, it sets the `height` for `Treeview`, but how can it improve the performance?

moathabuhamad-cengage: Lowering the Overscan affects performance, not height

Issue #1816, opened 2025-06-13
SPIKE: Investigate and Recommend Performance Testing Strategy


moathabuhamad-cengage:
The goal of this spike is to detect, prevent, and avoid future performance issues. for reference #1813  

Acceptance Criteria:
- Research tools and frameworks for front-end performance testing (e.g., Lighthouse, Playwright,  custom scripts).
- Prototype a minimal setup demonstrating how to measure render time, interaction latency, and memory usage for key UI components.
- Recommend a performance testing strategy, including:
    - Which tools to use
    - How to integrate them into our CI pipeline
    - How to define and enforce performance budgets
- Document findings in our wiki

silvalaura:
We need to audit all components to understand where to focus performance testing.

PR #2133, opened 2025-12-15
chore(test): add playwright performance tests

moathabuhamad-cengage:
Closes: #1816

## Checklist 


olehnoskov-cengage (approved):

tests/playwright/tests/performance/checkbox-performance.spec.ts
silvalaura, line 1: Should we compare the performance between native checkboxes vs our checkboxes? I think that's an interesting data point.

Issue #2345, opened 2026-03-17
Safari issues in 4.11


silvalaura:
1. The modal's focus shifts beyond its content.

https://github.com/user-attachments/assets/41289201-e309-422f-ad30-cd85b49805b8

2. Duplications in Input with maxCount
<img width="2639" height="1710" alt="Image" src="https://github.com/user-attachments/assets/c2b5e7f1-1b7f-43a0-ae98-78ae09818ba4" />

<img width="2504" height="1709" alt="Image" src="https://github.com/user-attachments/assets/b8689bbd-c6df-49e6-b94b-87e01f68cb41" />

olehnoskov-cengage:
Reverted `useFocusLock` changes to fix focus behavior in this [PR](https://github.com/cengage/react-magma/pull/2393)

PR #1731, opened 2025-04-01
fix(Character Counter): Fixed screenreader properly conveying the remainder in a character counter on user input.

chris-cedrone-cengage:
Closes: #1682 

## What I did
- Added ability for screenreaders to give proper feedback to the remaining characters as the user types.

## Screenshots
https://github.com/user-attachments/assets/038ed3fb-9a9a-4c85-a10d-9ba5916f90a7

## Checklist 
- [X] changeset has been added
- [X] Pull request is assigned, labels have been added and ticket is linked
- [X] Pull request description is descriptive and testing steps are listed
- [N/A] Corresponding changes to the documentation have been made
- [X] New and existing unit tests pass locally with the proposed changes
- [N/A] Tests that prove the fix is effective or that the feature works have been added

## How to test
- With your favorite screenreader open, go to Storybook
- Character Counter > Start typing
- Verify the screenreader dictates the proper remaining count as you type
- Rejoice

PR #2346, opened 2026-03-18
fix(Modal & CharacterCounter): Fix Safari-specific focus and renderin…

moathabuhamad-cengage:
Closes: #2345 

### 1 Modal focus escaping in Safari

  Safari does not respect aria-modal="true" for focus containment (unlike Chrome/Firefox). PR #2183 (4005029) reordered the close button before the body content in DOM order. The old useFocusLock only intercepted Tab at the edges (wrap on last item, wrap on first item with Shift+Tab)
  and relied on native browser behavior for all intermediate items. Chrome handled this via aria-modal, but Safari let focus escape after the close button.

  Fix: Rewrote the Tab handler in useFocusLock.ts to explicitly manage all Tab navigation with event.preventDefault() + indexOf, so native browser Tab behavior is never invoked. Also added a rootNode.contains() guard so nested modals don't interfere with each other's focus traps.
  Added a unit test for the nested modal regression.

### 2 Character counter read twice by VoiceOver

  PR #1731 (70b8ba0) added a hidden aria-live region (HiddenLabelText) for screen reader announcements inside the same container that aria-describedby points to. After the 3-second debounce, that container held both the visible text and the hidden text with identical content.
  VoiceOver reads all text inside an aria-describedby target, so it announced the counter twice (e.g. "5 characters left 5 characters left").

  Fix: Moved HiddenLabelText outside the id'd container so aria-describedby only picks up the visible text. Set aria-live="off" on the visible element to prevent duplicate live region announcements. Removed the isMacOS platform check — the hidden live region now handles announcements
  consistently on all platforms.

## Checklist 

## How to test

  Modal focus trap (Safari only):
  1. Open any Modal story in Safari (e.g. Default Modal)
  2. Open the modal
  3. Verify focus lands on the modal heading
  4. Press Tab repeatedly — focus should cycle through modal items (close button → body content → back to close button) and never escape to elements behind the modal
  5. Press Shift+Tab — focus should cycle in reverse
  6. Verify the same behavior works in Chrome (regression check)

  CharacterCounter duplication (Safari only):
  1. Open the Input story with maxCount prop in Safari
  2. Verify the character counter text (e.g. "10 characters allowed") appears once, not duplicated
  3. Type in the input — verify the counter updates without visual duplication
  4. Exceed the maxCount — verify the error state displays correctly
  5. Verify screen reader announcements still work (VoiceOver on macOS: counter is announced after a brief debounce)

### DIFF
--- a/.changeset/chore-fix-modal-safari-focus.md
+++ b/.changeset/chore-fix-modal-safari-focus.md
@@ -2,7 +2,5 @@
 'react-magma-dom': patch
 ---
 
-fix(Modal & CharacterCounter): Fix Safari-specific issues.
-
-- Modal: Fix focus escaping the modal in Safari by handling forward Tab from the header in `useFocusLock`. Safari does not respect `aria-modal` for focus containment, and the DOM reorder from #2183 exposed a gap in the focus trap.
-- CharacterCounter: Fix visual duplication when `maxCount` is applied in Safari by removing `aria-live` from the visible element and using a single hidden live region for screen reader announcements.
+- fix(Modal): Fix focus escaping the modal in Safari.
+- fix(CharacterCounter): Fix voice over announcement duplication when `maxCount` is applied in Safari.
--- a/.changeset/chore-fix-modal-safari-focus.md
+++ b/.changeset/chore-fix-modal-safari-focus.md
@@ -1,0 +1,8 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Modal & CharacterCounter): Fix Safari-specific issues.
+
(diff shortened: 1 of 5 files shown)


olehnoskov-cengage (commented):
**Character counter**
Character counter text appears twice
Image

**Modal**
Focus escapes from the modal
[Screencast from 18.03.26 11:26:01.webm](https://github.com/user-attachments/assets/89b8a569-b880-4634-954a-8045bf0e3314)

nikitaorliak-cengage:
**Character Counter:**
image

image

Duplication exists on the first focus and after typing anything. 

In my opinion, it's how Safari + VO works. We shouldn't spend a lot of time trying to fix this duplication announcement if everything works fine on other OS/Browsers/Screenreaders.

nikitaorliak-cengage (approved):

olehnoskov-cengage (approved):

PR #2362, opened 2026-03-19
V4/fix(Dropzone): fix docs example

nikitaorliak-cengage:
Fix docs example

### DIFF
--- a/.changeset/fix-Dropzone-docs-example.md
+++ b/.changeset/fix-Dropzone-docs-example.md
@@ -1,0 +1,5 @@
+---
+"@react-magma/dropzone": patch
+---
+
+fix(Dropzone): Fix docs example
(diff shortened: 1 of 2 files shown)


.changeset/fix-Dropzone-docs-example.md
```diff
@@ -0,0 +1,5 @@
+---
+"@react-magma/dropzone": patch
```
silvalaura, line 2: docs

nikitaorliak-cengage: I thought about it, but it's just an additional pr for this #2360 one, so I wasn't sure

silvalaura: We don't want this change in the changelog (our adopters don't care about a docs change) and therefore I "hide" these changesets under docs because no one reads those.

silvalaura: If you want, feel free to merge and we can clean this up in the clean up PR

PR #2469, opened 2026-04-28
fix(Toast): Toast does not close on click or after timeout

### LINKED ISSUE #2388: Toast: The component doesn’t close on click or after timing out.
Found an issue with Toast component when using React.StrictMode
React Magma version 4.11.0
The issue is that we cannot close the Toast; even after the timer is up, it still shows. This is likely because of the double mounting of components in Strict Mode.
Here is a place in a react magma code where we are likely facing the issue:
https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/Toast/index.tsx#L196

During first mount, the cleanup function is called and the mountedRef.current become false

https://cengage.slack.com/archives/C8Y4FR63U/p1774623249579259

artemtrusov-cengage:
Closes: #2388

## What I did

In React 18 and higher, Strict Mode, `useEffect` runs mount → unmount → mount again in development, which caused `mountedRef.current` to stay in an incorrect state. Updated logic to properly reset the flag on each mount/unmount cycle so the toast behavior works correctly and state is not retained between Strict Mode re-mounts.

## Checklist 

## How to test

Wrap the Toast component in <React.StrictMode> in Storybook -> Open Storybook/Docs -> Toast -> trigger Toast → try to close it via click and via timeout

### DIFF
--- a/.changeset/fix-Toast-not-closing.md
+++ b/.changeset/fix-Toast-not-closing.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(Toast): Fix: Toast not closing on click or timeout (Strict Mode issue with unmount)
--- a/packages/react-magma-dom/src/components/Toast/Toast.stories.tsx
+++ b/packages/react-magma-dom/src/components/Toast/Toast.stories.tsx
@@ -122,3 +122,49 @@ export const TwoLine = {
     ...Default.args,
   },
 };
+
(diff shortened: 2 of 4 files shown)


nikitaorliak-cengage (approved):

silvalaura:
Can we add this Strict example to storybook for easy testing?

moathabuhamad-cengage (approved):

PR #2475, opened 2026-05-04
fix(TreeView): Stop All subitems are selected from being announced on…

### LINKED ISSUE #2444: Selectable TreeView: subitems are selected is announced multiple times
**Describe the bug**
a11y Justin reported that selectable treeviews have a repeated announcement:
https://github.com/user-attachments/assets/cd97da71-5635-4d7e-8e3d-ed6d2011a138

I can reproduce this in our docs as well with VO.

**To Reproduce**
1. Go to a selectable tree view and expand a parent item
2. Use a screenreader (can reproduce with NVDA + VO) and listen to "All subitems are selected" getting announced the number of parent children inside the selected item

**Expected behavior**
Unsure. Probably should not announce repeatedly, but we need to investigate other Trees and determine how it should behave

moathabuhamad-cengage:
Closes: #2444

## Checklist 

## How to test

- Go to a selectable tree view and expand a parent item
- Use a screenreader (can reproduce with NVDA + VO) and listen to "All subitems are selected" getting announced only once

### DIFF
--- a/.changeset/fix-treeview-repeated-subitem-announcement.md
+++ b/.changeset/fix-treeview-repeated-subitem-announcement.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+fix(TreeView): Stop All subitems are selected from being announced once per descendant when a parent is checked.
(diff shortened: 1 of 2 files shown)


artemtrusov-cengage:
This solution does not work in macOS with VoiceOver - the message is announced twice

PR #2490, opened 2026-05-08
feat(ButtonGroup): add isList prop for semantic list rendering

### LINKED ISSUE #2472: ButtonGroup: new `isList` prop (Investigation)
https://jira.cengage.com/browse/A11Y-7345

**Description:**
In several dialogs (e.g., Add folder, Add group, Add to class, Questions, Create question), action buttons such as Cancel, Add, Save, or Place are wrapped inside a list structure (`<ul>/<li>` or `role="list"`). This results in screen readers announcing a list (e.g., “list with 2 items”), which is not semantically relevant for standalone action buttons and may create confusion for users.

**Request:**
This ticket is to investigate whether this pattern is intentional and aligned with our accessibility and design standards. If not, consider introducing a configurable approach.

**Proposed Direction:**
- Evaluate if list semantics are necessary in these contexts.
- If flexibility is needed, introduce a prop (e.g., `isList`) for the component controlling the button group.
- When `isList = false`, render buttons without list semantics (no `<ul>/<ol>` or `role="list"`).
- Ensure buttons remain properly exposed with correct roles and accessible names.

**User Impact:**
Removing unnecessary list announcements would reduce cognitive load for screen reader users and improve clarity of available actions.

artemtrusov-cengage:
Closes: #2472

## What I did
- **New feature:** Added `isList` prop to `ButtonGroup` for semantic list rendering (`<ul>/<li>`) in addition to the default `<div>`.
- **Docs:** Updated documentation and stories to reflect new `isList` and role management features.

## Checklist 

## How to test
Open Storybook → ButtonGroup → open DevTools → check structure → enable isList → verify structure

### DIFF
--- a/.changeset/feat-ButtonGroup-add-isList-prop.md
+++ b/.changeset/feat-ButtonGroup-add-isList-prop.md
@@ -1,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+feat(ButtonGroup): add `isList` prop for semantic list rendering.
--- a/.changeset/feat-ButtonGroup-add-isList-prop.md
+++ b/.changeset/feat-ButtonGroup-add-isList-prop.md
@@ -1,5 +1,5 @@
 ---
-'react-magma-dom': patch
+'react-magma-dom': minor
 ---
 
 feat(ButtonGroup): add `isList` prop for semantic list rendering.
--- a/.changeset/fix-Treeview-VO-doesn't-announce-state.md
+++ b/.changeset/fix-Treeview-VO-doesn't-announce-state.md
@@ -1,5 +1,5 @@
 ---
-'react-magma-dom': minor
+'react-magma-dom': patch
 ---
 
 fix(Treeview): Add an announcement for the expansion state of the TreeItem.
--- a/.changeset/fix-Treeview-VO-doesn't-announce-state.md
+++ b/.changeset/fix-Treeview-VO-doesn't-announce-state.md
@@ -1,5 +1,5 @@
 ---
-'react-magma-dom': patch
+'react-magma-dom': minor
 ---
 
 fix(Treeview): Add an announcement for the expansion state of the TreeItem.
--- a/packages/react-magma-dom/src/components/ButtonGroup/index.tsx
+++ b/packages/react-magma-dom/src/components/ButtonGroup/index.tsx
@@ -224,7 +224,6 @@ const buildNoSpaceButtonStyles = (
 const buildDivChildStyles = (props: StyledButtonGroupProps) => css`
   margin: ${buildButtonMargin(props)};
   flex: ${buildFlex(props)};
-
   button {
     &:nth-child(2) {
       width: 40px;
--- a/packages/react-magma-dom/src/components/ButtonGroup/index.tsx
+++ b/packages/react-magma-dom/src/components/ButtonGroup/index.tsx
@@ -75,6 +75,16 @@ export interface ButtonGroupProps extends React.HTMLAttributes<HTMLDivElement> {
    * @internal
    */
   theme?: ThemeInterface;
+  /**
+   * ARIA role for the ButtonGroup container.
+   * @default "group" for div mode, "list" for list mode
+   */
+  role?: string;
+  /**
+   * Renders ButtonGroup as a `<ul>` with children wrapped in `<li>`.
+   * @default false
+   */
+  isList?: boolean;
 }
 
 export interface ButtonGroupContextInterface {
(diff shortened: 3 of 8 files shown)


packages/react-magma-dom/src/components/ButtonGroup/index.tsx
```diff
@@ -206,43 +221,85 @@ const buildNoSpaceButtonStyles = (
...
+const buildDivChildStyles = (props: StyledButtonGroupProps) => css`
+  margin: ${buildButtonMargin(props)};
+  flex: ${buildFlex(props)};
 
-  &:last-child:not(:only-child) {
-    > button {
-      margin-right: 0;
+  button {
+    &:nth-child(2) {
+      width: 40px;
```
moathabuhamad-cengage, line 240: Do we have a theme var for this 40px?

artemtrusov-cengage: ~~Thx, fixed it by adding `props.theme.spaceScale.spacing09`~~
I realized this isn't the right approach, and as far as I can see, we don't have a var for width.

.changeset/feat-ButtonGroup-add-isList-prop.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-magma-dom': patch
+---
+
+feat(ButtonGroup): add `isList` prop for semantic list rendering.
```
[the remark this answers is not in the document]
silvalaura: AI is right-- this is a minor change also because we are adding a prop

packages/react-magma-dom/src/components/ButtonGroup/index.tsx
```diff
@@ -75,6 +75,16 @@ export interface ButtonGroupProps extends React.HTMLAttributes<HTMLDivElement> {
...
+   * ARIA role for the ButtonGroup container.
+   * @default "group" for div mode, "list" for list mode
+   */
+  role?: string;
+  /**
+   * Renders ButtonGroup as a `<ul>` with children wrapped in `<li>`.
+   * @default false
+   */
+  isList?: boolean;
 }
```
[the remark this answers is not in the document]
silvalaura: This seems valid

packages/react-magma-dom/src/components/ButtonGroup/index.tsx
silvalaura: Yeah, I'm not sure we want to hard code this as our buttons can have different sizes

silvalaura:
When isList is on with split button, there's a tiny gap to the left that is not present in div
image
image

silvalaura:
In the small breakpoint example, turning isList on breaks the styles
image
image

packages/react-magma-dom/src/components/ButtonGroup/index.tsx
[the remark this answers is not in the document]
artemtrusov-cengage: We use `Emotion`, which returns a `div`, together with `forwardRef`. Our component can render as either `div` or `ul`, which creates typing complexity. I can make it polymorphic, but because `Emotion` internally types the result as `HTMLDivElement`, I'll have to use type assertion `ref as React.Ref<HTMLDivElement>`. If a user types the `ref` as `HTMLUListElement`, TypeScript won't throw an error, even though the actual type might not match.

packages/react-magma-dom/src/components/ButtonGroup/index.tsx
artemtrusov-cengage: This part is for the split button. I didn’t change the logic, just reused the existing one. I noticed an issue with `alignment: fill` - it causes incorrect wrapping of the button. I’ll fix it now.

As for 40px, I think it’s fine since it’s used for the split button icon, and this size has already been in our codebase for a long time. Please correct me if I’m wrong.

.changeset/fix-Treeview-VO-doesn't-announce-state.md
```diff
@@ -1,5 +1,5 @@
 ---
-'react-magma-dom': patch
+'react-magma-dom': minor
```
silvalaura: I think the wrong changeset was changed :)

artemtrusov-cengage: oops... 🫣
Thank you for your careful attention to detail

packages/react-magma-dom/src/components/ButtonGroup/index.tsx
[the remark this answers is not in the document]
silvalaura: Great. Do you think this is worth documenting in docs, or do you think it'll be obvious to adopters?

packages/react-magma-dom/src/components/ButtonGroup/index.tsx
silvalaura: Okay, let's leave it!

silvalaura (approved):

moathabuhamad-cengage (approved):

PR #2555, opened 2026-06-22
fix(Tag): Prevent spreading ARIA attributes onto the inner label `span` to resolve `aria-allowed-attr` violation.

### LINKED ISSUE #2554: Tag: Component incorrectly passes ARIA attributes to <span>, causing accessibility violation
**Describe the bug**
The Tag component spreads `{...rest}` props to both `StyledTag` and `LabelWrap`. When `StyledTag` renders as a `<button>`, ARIA attributes are valid because the element has an implicit button role. However, the same ARIA attributes are also applied to `LabelWrap`, which renders as a `<span>`. Since `<span>` doesn't support these ARIA attributes, this results in an accessibility violation (aria-allowed-attr).

**Expected behavior**
ARIA attributes should only be applied to elements that support them. The Tag component should not pass ARIA-related props from `{...rest}` to `LabelWrap` (`<span>`). Accessibility scans shouldn't report aria-allowed-attr violations for the component.

**Additional content**
Reference:
https://github.com/cengage/react-magma/blob/dev/packages/react-magma-dom/src/components/Tag/Tag.tsx#L393

Related accessibility rule:
https://dequeuniversity.com/rules/axe/4.11/aria-allowed-attr

nikitaorliak-cengage:
Closes: #2554

## What I did

Prevent spreading ARIA attributes onto the inner label `span` to resolve `aria-allowed-attr` violation.

## Checklist 

## How to test

Pull the branch -> Open Storybook locally -> Tag -> Pass some props (e.g., color/background-color/aria-label) -> Check passed props in devtools

### DIFF
--- a/packages/react-magma-dom/src/components/Tag/Tag.tsx
+++ b/packages/react-magma-dom/src/components/Tag/Tag.tsx
@@ -228,14 +228,6 @@ function buildSvgOpacity(props) {
 }
 
 function buildTagPadding(props) {
-  if (props.icon) {
-    switch (props.size) {
-      case 'small':
-        return `0 ${props.theme.spaceScale.spacing02}`;
-      default:
-        return `${props.theme.spaceScale.spacing02} 6px`;
-    }
-  }
   switch (props.size) {
     case 'small':
       return `0 ${props.theme.spaceScale.spacing02}`;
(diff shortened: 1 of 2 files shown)


packages/react-magma-dom/src/components/Tag/Tag.tsx
```diff
@@ -390,7 +373,7 @@ export const Tag = React.forwardRef<HTMLButtonElement, TagProps>(
         {...rest}
       >
         {icon}
-        <LabelWrap size={size} theme={theme} {...rest}>
```
silvalaura, line 393: Could removing this be a breaking change for our adopters? Do we need to pass rest but not the role?

packages/react-magma-dom/src/components/Tag/Tag.tsx
```diff
@@ -328,7 +312,6 @@ const StyledSpan = styled.span<{
 
 const LabelWrap = styled.span<{
   size: string;
-  icon?: any;
```
silvalaura, line 331: Why are we removing this again?

silvalaura (approved):

artemtrusov-cengage (approved):

PR #2635, opened 2026-07-27
feat(Dropdown): add mainButtonRef prop to split button

### LINKED ISSUE #1849: DropdownSplitButton: Ability to pass ref for (left) Button
**Describe the bug**
When using a DropdownSplitButton, the ref I pass is only for the dropdown button and not the left Button. I want to be able to pass a ref to either so that I can set my focus to either button.

**Expected behavior**
User can pass a ref for either button in DropdownSplitButton

artemtrusov-cengage:
Closes: #1849

## What I did
Added a new primaryButtonRef prop to pass a ref to the primary (left) button of the Split Button dropdown.

## Checklist 

## How to test

1. Open the Dropdown Storybook
2. Open the Custom Ref Template example (second Split Button example)
3. Verify that focus is returned to the primary (left) button of the Split Button

### DIFF
--- a/packages/react-magma-dom/src/components/Dropdown/DropdownSplitButton.tsx
+++ b/packages/react-magma-dom/src/components/Dropdown/DropdownSplitButton.tsx
@@ -46,9 +46,9 @@ export interface DropdownSplitButtonProps extends ButtonStyles {
    */
   onClick?: () => void;
   /**
-   * Ref for the primary action button (left side).
+   * Ref for the main action button (left side).
    */
-  primaryButtonRef?: React.Ref<HTMLButtonElement>;
+  mainButtonRef?: React.Ref<HTMLButtonElement>;
 }
 
 export const DropdownSplitButton = React.forwardRef<
@@ -66,7 +66,7 @@ export const DropdownSplitButton = React.forwardRef<
     'aria-label': ariaLabel,
     children,
     id,
-    primaryButtonRef,
+    mainButtonRef,
     variant = ButtonVariant.solid,
     onClick,
     leadingIcon,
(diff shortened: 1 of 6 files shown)


serhiiyanhol-cengage (approved):

packages/react-magma-dom/src/components/Dropdown/DropdownSplitButton.tsx
```diff
@@ -45,6 +45,10 @@ export interface DropdownSplitButtonProps extends ButtonStyles {
    * Function that fires when the button is clicked
    */
   onClick?: () => void;
+  /**
+   * Ref for the primary action button (left side).
+   */
+  primaryButtonRef?: React.Ref<HTMLButtonElement>;
```
silvalaura: We use "primary" to refer to a color. In this case, we are referring to the "big" button instead of the little one, it has nothing to do with the color. What else could we name this to make sense without the word primary?

silvalaura (changes requested):
Code looks fine, but want to reconsider the prop name

packages/react-magma-dom/src/components/Dropdown/DropdownSplitButton.tsx
artemtrusov-cengage: Renamed the prop to `mainButtonRef`. I couldn't come up with a better name 🙂

If this name works, we can keep it and merge the PR. Otherwise, we can choose a different name.

Some alternatives suggested by AI:
- `actionButtonRef`
- `toggleButtonRef`
- `leftActionRef`
- `leftButtonRef`
- `contentButtonRef`

silvalaura: I think main is fine here. Thanks Artem!

silvalaura (approved):

nikitaorliak-cengage:
When I press `Esc` -- the focus comes back to the right (icon) button

nikitaorliak-cengage (approved):

PR #2648, opened 2026-07-29
feat(Dropdown): add apiRef prop to control the dropdown state

### LINKED ISSUE #2541: Dropdown: add useRef support
**Is your feature request related to a problem? Please describe.**

Currently, the Dropdown component cannot be controlled via a ref. This makes it harder to implement custom behaviors and advanced interaction patterns.

**Describe the solution you'd like**

Add ref support to the Dropdown component, similar to the [Popover component](https://react-magma.cengage.com/version/4.14.0/api/popover/#controlled_popover_with_api_ref ). 

This would allow consumers to control the dropdown state through a ref when needed.

**Additional context**

[Slack conversation](https://cengage.slack.com/archives/C8Y4FR63U/p1780575866437849)

artemtrusov-cengage:
Closes: #2541

## What I did
Following the same approach as Popover, I added an apiRef to control the Dropdown state.

## Checklist 

## How to test
Open the Storybook -> API Ref example -> Verify that everything works

### DIFF
--- a/packages/react-magma-dom/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react-magma-dom/src/components/Dropdown/Dropdown.tsx
@@ -34,8 +34,8 @@ export enum DropdownAlignment {
 }
 
 export interface DropdownApi {
-  closeDropdownManually(event?: React.SyntheticEvent): void;
-  openDropdownManually(): void;
+  closeDropdownManually(event: React.SyntheticEvent): void;
+  openDropdownManually(event: React.SyntheticEvent): void;
 }
 
 export interface DropdownProps extends React.HTMLAttributes<HTMLDivElement> {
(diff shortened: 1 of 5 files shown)


packages/react-magma-dom/src/components/Dropdown/Dropdown.tsx
```diff
@@ -33,6 +33,11 @@ export enum DropdownAlignment {
   end = 'end',
 }
 
+export interface DropdownApi {
+  closeDropdownManually(event?: React.SyntheticEvent): void;
+  openDropdownManually(): void;
```
serhiiyanhol-cengage: PopoverApi/DatePickerApi declare event on both methods (untyped, open* keeps it unused for a symmetric signature). This PR drops it on openDropdownManually() and types close's event — nicer, but now the three apiRef APIs don't match. Can/should we align?

packages/react-magma-dom/src/components/Dropdown/Dropdown.tsx
```diff
@@ -33,6 +33,11 @@ export enum DropdownAlignment {
   end = 'end',
 }
 
+export interface DropdownApi {
```
serhiiyanhol-cengage, line 36: Should the `DropdownApi` be exported from the top-level index.ts so consumers can type their ref, e.g. useRef<DropdownApi>()? Right now only DropdownProps is exported (index.ts), so the type isn't importable from the package. Popover does this — index.ts has export type { PopoverProps, PopoverApi } from './components/Popover'

packages/react-magma-dom/src/components/Dropdown/Dropdown.tsx
artemtrusov-cengage: Aligned **Dropdown API** with the other components so both methods take `event` for consistency. I also typed event as `React.SyntheticEvent` to match current component usage, and I did not change the others to avoid potentially breaking existing consumer behavior.

serhiiyanhol-cengage (approved):

nikitaorliak-cengage (approved):

PR #2656, opened 2026-08-05
feat(charts): add `titlePrefix` and `titleSuffix` props to the chart toolbar

### LINKED ISSUE #2653: Charts: Add beforeTitle and afterTitle props to ChartToolbar
Add `beforeTitle` and `afterTitle` props of type `React.ReactNode` to `ChartToolbarConfig`.

These props are available only in `chartToolbar` mode, since without it the title is rendered by Carbon and there is no React extension point.

The title remains plain text and keeps its existing `titleLevel` semantics.

## Requirements

- Slot content renders outside the heading element, so the heading's accessible name remains options.title.
- Slot content appears before the toolbar actions in DOM order, so keyboard focus reaches it before the fullscreen and export buttons.
- The title and slots share a flex container with min-width: 0, allowing long titles to shrink instead of pushing the toolbar actions off.
- Wrap the slot area in InverseContext.Provider; otherwise, Magma components rendered in slots will use the light theme even on an inverse chart.

`afterTitle` is the chart equivalent of the `additionalContent` prop already used by **Input** and **Select**. The same accessibility guidance should apply: the **IconButton** should have an `aria-label` that matches the **Tooltip** text.

## Acceptance Criteria

- Add tests covering DOM order, the heading's accessible name, and inverse mode support for slot content.
- Add a Storybook example demonstrating the recommended Tooltip + IconButton pattern.
- Update the documentation to describe the new props.
- Verify the implementation in both fullscreen and inverse modes.

artemtrusov-cengage:
Closes: #2653

## What I did
Two new props have been added to allow rendering custom content **before** and **after** the chart title.

### This also includes:

- tests covering the new functionality,
- Storybook examples,
- documentation updates.

## Screenshots
Screenshot 2026-08-05 at 12 09 03

## Checklist 

## How to test
- Open Storybook → Chart Info Tooltip example -> Check how the feature works overall.
- Open the Docs → Charts demos → Toolbar customization → Custom content around the title -> Verify the example on the page and review the documentation.


serhiiyanhol-cengage (approved):

website/react-magma-docs/src/pages/data-visualization/chart-demos.mdx
```diff
@@ -2437,6 +2437,64 @@ export function Example() {
 }
 ```
 
+### Custom content around the title
+
+`beforeTitle` and `afterTitle` let you render custom content inline with the built-in chart title. They are the chart equivalent of the `additionalContent` prop on [Input](../../api/input#help_link), and the same guidance applies: they are best used with a `Tooltip` wrapping an `IconButton`, with the button's `aria-label` matching the tooltip text.
```
silvalaura: I wonder if we should call these `additionalContentBeforeTitle`??? it's kind of long, but would be very explicit (same with after title)

silvalaura:
Can we add an example (or just a badge or something that will make it obvious) so that the docs site has an example of using the before slot?

website/react-magma-docs/src/pages/data-visualization/chart-demos.mdx
artemtrusov-cengage: `additionalContentBeforeTitle` feels a bit too long to me. Maybe we can go with something simpler like `titlePrefix / titleSuffix` or `beforeTitleContent / afterTitleContent`

nikitaorliak-cengage (approved):

website/react-magma-docs/src/pages/data-visualization/chart-demos.mdx
silvalaura: Oooh I kind of like titlePrefix / titleSuffix

